PR #109, opened 2016-08-30
Use property accessors for values that stay the same in branch copying methods

apeteri:
This is done to prevent accidentally using a field carrying the previous value instead of the method parameter in `withT(T newValue)` methods of Branch subclasses.


apeteri:
Merged with commit e292215281f9ccbb864195e303c09e63a7f6a475.

PR #110, opened 2016-09-21
SO-2022 Change complex map field types

nagyo:
Complex map's map group and map priority fields were stored as byte which made it impossible to import maps with group/priority numbers higher than Byte.MAX_VALUE (127), so the fields were changed to integer.


snomed/com.b2international.snowowl.snomed.importer.rf2/src/com/b2international/snowowl/snomed/importer/rf2/validation/SnomedExtendedMapTypeRefSetValidator.java
```diff
@@ -67,7 +69,7 @@ private void validateMapCategory(final List<String> row) {
 		final String uuid = row.get(0);
 		final String effectiveTime = row.get(1);
 		final String mapCategory = row.get(12);
-		if (!isComponentExists(mapCategory, ReleaseComponentType.CONCEPT)) {
+		if (getComponentCategory(mapCategory) == ComponentCategory.CONCEPT && !isComponentExists(mapCategory, ReleaseComponentType.CONCEPT)) {
```
apeteri: I think we should reject reference set rows where the `mapCategoryId` value is not a valid SCTID. But that also holds true for all reference set validators, not just this one (report an error if one of the component SCTIDs in a column is nonsensical or does not match the expected type).

apeteri:
Changing the base of this PR to be included in #131.

PR #129, opened 2017-02-17
SO-2179 file attachment api

cmark:
A very simple rpc based low-level Java API to upload and download file (zip files only) attachments.
The server will store the attachments in a directory called `attachments` relative to the currently configured resources directory.


cmark:
👍

PR #132, opened 2017-03-03
SO-2178-job-api

cmark:
Introduce request based job API.
The implementation allows **any Request** to be wrapped into a job to monitor its execution.

Major changes:
* Added a single EventBus instance to the global ApplicationContext to handle all incoming requests (Remote Jobs are still managed by the Eclipse Job API under the hood)
* Removed repository specific IEventBus handlers
* Support java.util.Date types in document mapping
* Introduced RxJava based Notification observable for all system level notifications
* Request interface is now a FunctionalInterface (automatically detects the return type based on the generic type parameters)
* Flatten RequestBuilder hierarchy (classes/methods that supported asynchronous request wrapping are now extracted to interfaces with default methods)

Minor changes/fixes:
* DelegatingServiceProvider now disposes all locally registered services during server shutdown
* ReviewManager is now an IDisposableService


cmark:
👍

PR #133, opened 2017-03-05
Bumped up the Xtext version to 2.11.0 and the MWE2 to 2.9.0.

kittaakos:
Please note, once the changes are merged, then one has to update the Xtext/Xtend tooling version to 2.11 in Eclipse. There were [breaking Xtend changes] which affect the active annotations. More details about the issue are available [here].

Also, I had to provide a workaround for an issue with the `contipref` library. It does not work with JUnit 4.12 due to this [bug]. I have tried to get the latest version of this library from Maven, but I had no luck. I was looking for version `2.4.3`. I decided to check out the source and provide a jar from a local build. But this project is quite fragile; when I run the Maven build from command line, I get errors, when running the tests from Eclipse I get completely different errors. What is the purpose of this library? Are you storing your benchmark results somewhere for historical comparison or it is used only for assertions regarding the execution times? Although I have no experience with [JMH], that seems to be the recommended way to have a proper benchmark. BTW, the owner suggests the [same] and explicitly declares that to project is not continued anymore.

I have not yet created a dedicated project for the ECL testing, but that is my plan. Do you have any objections?

I look forward to your feedback on this PR.

[breaking Xtend changes]: https://github.com/eclipse/xtext-lib/commit/da32a7d4bfc91678208f8e04690d22278e219792
[here]: https://github.com/eclipse/xtext-xtend/issues/122
[bug]: https://github.com/lucaspouzac/contiperf/issues/9
[JMH]: http://openjdk.java.net/projects/code-tools/jmh/
[same]: https://github.com/lucaspouzac/contiperf/issues/8#issuecomment-161315536

Signed-off-by: Akos Kitta <kittaakos@gmail.com>


cmark:
Thank you for the PR.
We plan to replace `contiperf` with `JMH`, but the workaround is okay for now.

What kind of ECL testing project would you like to create/add? The one that Xtext generates (`c.b.s.snomed.ecl.tests`)?
Our plan with the number of projects right now is to reduce it to a reasonable amount.
For example, it is okay to add the Xtext generated test project to the project set, but that project should contain only grammar, validation, formatting, etc. tests. The evaluation logic should be part of the core SNOMED CT test bundle (`c.b.s.snomed.datastore.tests`), as it is right now. 
Also we plan to merge all bundles that correspond to a single terminology into a single one. For example: merge every `*.snomed.*` bundle into a single `c.b.s.snomed.core` bundle (except SCG, ESCG, ECL grammar projects).

I hope this helps, feel free to ping me on Skype if you have further questions.

kittaakos:
> The evaluation logic should be part of the core SNOMED CT test bundle (c.b.s.snomed.datastore.tests), as it is right now.

Correct. I would like to do exactly what you've described. Probably I do not even need to run the tests with `tycho-surefire` but `maven-surefire`. I need neither the OSGi classloader nor the Eclipse platform. Do you have pure maven tests? I did not find them. Maybe I have just missed them. Or is the  Tycho way the preferred one? I am asking because, for instance, the LS does not need Eclipse at all.

cmark:
> Do you have pure maven tests? Or is the Tycho way the preferred one?

We have basic JUnit tests (like the `commons.tests˙` bundle), but we let Tycho to run them all. This way, we only configure the `tycho-surefire-plugin` in the root pom and use `eclipse-test-plugin` packaging in the actual module. I think you can do the same with the ECL test project, but if it's easier for you to use `maven-surefire`, just use `eclipse-plugin` instead of `eclipse-test-plugin` in the pom, and you're good to go.

kittaakos:
Sounds great, once you merge this PR, I start working on the ECL grammar and the tests. Just a short remark; I am glad that the headless Maven build runs so smoothly and fast. Great job!

kittaakos:
One more quick question on this PR; how long will it roughly take to get this PR merged? I guess you need to update the development environments, TPs maybe other things and so on before applying the changes onto the `develop` branch.

cmark:
> how long will it roughly take to get this PR merged?

Our plan is to merge it after the 5.7.0 release, when we start the dev.iteration of 5.8.0-SNAPSHOT. Hopefully tomorrow or Wednesday.

kittaakos:
Excellent. Thank you!

cmark:
5.7.0 is out. Merging develop. Thank you for the contribution, @kittaakos!

PR #135, opened 2017-03-06
SO-2185 RF2 Export API refactor

cmark:
This PR simplifies the RF2 export API by introducing the Request based API on top of the existing one (see `SnomedRf2ExportRequestBuilder`). Moved all of existing API classes to an internal package (see `snomed.datastore.internal.rf2`).
Additionally added support for `getFile()` in `FileRegistry` API.


cmark:
Thanks @apeteri!

PR #136, opened 2017-03-09
ECL grammar changes

kittaakos:
This PR contains the following grammar changes:
 - Optional `WS` before and after the `PIPE` terminals.
 - Case insensitive keyword support. For instance, one can write `aNd` not just `AND`.
 - I have added basic parser tests from the language specification. (Please note, the tests are executed as Tycho Surefire tests, although it is not necessary.)
 - I have removed a couple of terminals from the grammar, it should not cause any side-effects but simplifies the generated code.

This PR does not contain the following change, but I am planning to apply it if you agree; currently if one creates a new resource, it raises parser error because the main entry has `[1..1]` [cardinality]. I would like to wrap the `ExpressionConstraint` rule into a `Script` and have `[0..1]` cardinality so that the user does not see an error upfront when creating a new ECL file. Also, I need to adapt the code that transforms the DSL into a query to handle missing constraint.

Please ping me if you want me to change anything regarding this PR.

[cardinality]: https://github.com/kittaakos/snow-owl/blob/ecl_grammar_changes/snomed/com.b2international.snowowl.snomed.ecl/src/com/b2international/snowowl/snomed/ecl/Ecl.xtext#L22


kittaakos:
> This PR does not contain the following change, but I am planning to apply it if you agree; currently if one creates a new resource, it raises parser error because the main entry has [1..1] cardinality.

I decided to add that as well so that we can have the grammar changes in one PR. I am closing this one.

PR #138, opened 2017-03-16
Feature: repository health status and repository info API

ekovacs:
https://snowowl.atlassian.net/browse/SO-2216
https://snowowl.atlassian.net/browse/SO-1969


core/com.b2international.snowowl.datastore.server/src/com/b2international/snowowl/datastore/server/internal/CDOBasedRepository.java
```diff
@@ -348,5 +363,81 @@ public void handleCommitInfo(CDOCommitInfo commitInfo) {
...
+		if (emptyDb && emptyIndex) {
+			return true; // empty dataset
+		}
+		
+		if (emptyDb ^ emptyIndex) {
+			LOG.error("{} is in inconsistent state. CDO {} but INDEX {}", getCdoRepository().getRepositoryName(), contentMessage(emptyDb), contentMessage(emptyIndex));
+			return false; // either CDO or index was deleted but not the other.
+		}
+		
+		return !hasInvalidCDOTimeStamps(cdoCommitInfos, getLast(indexCommitInfos).getTimeStamp(), branch);
```
apeteri: To be absolutely, positively sure that we have the same content, commit infos coming from CDO and the index should be compared pairwise. The two lists should have an equal length, and the timestamps should match when iterating over the two lists in lockstep (the lists should be sorted by commit timestamp first if they are not in this form already). 

Such a comparison would also cover the checks above (non-zero CDO commit infos against zero indexed commit infos, or the other way round).

core/com.b2international.snowowl.datastore.server/src/com/b2international/snowowl/datastore/server/internal/CDOBasedRepository.java
```diff
@@ -348,5 +363,81 @@ public void handleCommitInfo(CDOCommitInfo commitInfo) {
...
+		while (cdoCommitInfosIterator.hasNext()) {
+			CDOCommitInfo cdoCommitInfo = cdoCommitInfosIterator.next();
+			
+			if (cdoCommitInfo.getTimeStamp() <= lastIndexCommitTimestamp) {
+				//removing valid commits from the problematic list
+				cdoCommitInfosIterator.remove();
+			} 
+		}
+		
+		if (!problematicCommitInfos.isEmpty()) {
```
apeteri: If an older restored DB backup is matched up with a newer index, all the CDO commit infos from the problem list would be consumed; it should still be reported as a problem (ie., that now the index is ahead of the DB).

core/com.b2international.snowowl.datastore.server/src/com/b2international/snowowl/datastore/server/DatastoreServerBootstrap.java
```diff
@@ -147,9 +150,26 @@ public void run(SnowOwlConfiguration configuration, Environment env, IProgressMo
...
 			initializeRepositories(configuration, env);
 			initializeRequestSupport(env, configuration.getModuleConfig(RepositoryConfiguration.class).getNumberOfWorkers());
+			verifyRepositories(env);
 		}
 	}
 	
+	@SuppressWarnings("deprecation")
+	private void verifyRepositories(Environment env) {
+		final DefaultRepositoryManager repositories = (DefaultRepositoryManager) env.service(RepositoryManager.class);
+		Optional<InternalRepository> inConsistentRepository = repositories.repositories()
```
apeteri: Latin prefixes should not be considered camelCase boundaries; please rename this variable to `inconsistentRepository`.

core/com.b2international.snowowl.datastore.server/src/com/b2international/snowowl/datastore/server/DatastoreServerBootstrap.java
```diff
@@ -147,9 +150,26 @@ public void run(SnowOwlConfiguration configuration, Environment env, IProgressMo
...
+			.stream()
+			.filter(repository -> repository instanceof InternalRepository).map(InternalRepository.class::cast)
+			.filter(repository -> !repository.isConsistent(BranchPathUtils.createMainPath())) 
+			/* XXX: or should we check all the branches, not just the main??*/
```
apeteri: Good question! 😄 The problem is that in a production environment, there are _lots_ of branches. There is no easy way of retrieving commit infos from CDO across branches either, and even if it were, it might take up a lot of memory so you'd have to do it one at a time, in which case it would be slow instead of memory-consuming. 

We can keep the check on `MAIN` only for now.

core/com.b2international.snowowl.datastore.server/src/com/b2international/snowowl/datastore/server/DatastoreServerBootstrap.java
```diff
@@ -147,9 +150,26 @@ public void run(SnowOwlConfiguration configuration, Environment env, IProgressMo
...
+			.findAny();
+		
+		if (inConsistentRepository.isPresent()) {
+				LOG.error("Found inconsistent repository: {}, halting boot.", inConsistentRepository.get().getCdoRepository().getRepositoryName());
+				Thread.currentThread().suspend();
```
apeteri: Suspending the thread is dangerous; while it certainly blocks further operations here, we don't know what's running on the side, and it should still be possible to shut down cleanly as well.

Existing examples that break the startup mid-process are throwing `com.b2international.snowowl.core.SnowOwlApplication.InitializationException` if a problem is encountered. Can you check if it would work here as well?

ekovacs: https://jira.ihtsdotools.org/browse/APDS-124
> If the commits do not match Log a message at FATAL level and make Snow Owl pause. Do not exit the process as this will cause supervisor to restart it.


we may want to do here differently then the original request, that explicitly requested for suspending, rather than halting / deactivating.

core/com.b2international.snowowl.datastore.server/src/com/b2international/snowowl/datastore/server/DatastoreServerBootstrap.java
ekovacs: thank you! 🙇

core/com.b2international.snowowl.datastore.server/src/com/b2international/snowowl/datastore/server/internal/CDOBasedRepository.java
```diff
@@ -235,7 +250,7 @@ public RevisionBranch getParentBranch(String branchPath) {
 		bind(RevisionIndex.class, revisionIndex);
 		// initialize the index
 		index.admin().create();
-		
+		getDelegate().services().registerService(Searcher.class, indexClient.searcher());
```
ekovacs: you are right, this is bad to register this to app context.
however i have problem getting the index requests through, as at this point it seems that the searcher is not available at CommitInfoSearchRequest.doExecute(CommitInfoSearchRequest.java:59)
throwing an NPE at me...

core/com.b2international.snowowl.datastore.server/src/com/b2international/snowowl/datastore/server/internal/CDOBasedRepository.java
ekovacs: during import, for versions: there is only one commit for the index, but there are multiple commits for CDO, thus the size never matches. For this reason @cmark suggested to check for only after index CDO commits

core/com.b2international.snowowl.datastore.server/src/com/b2international/snowowl/datastore/server/DatastoreServerBootstrap.java
```diff
@@ -147,9 +151,26 @@ public void run(SnowOwlConfiguration configuration, Environment env, IProgressMo
...
+		Optional<InternalRepository> inconsistentRepository = repositories.repositories()
+			.stream()
+			.filter(repository -> repository instanceof InternalRepository).map(InternalRepository.class::cast)
+			.filter(repository -> !repository.isConsistent(BranchPathUtils.createMainPath()))
+			/* XXX: we are checking MAIN only */
+			.findAny();
+		
+		if (inconsistentRepository.isPresent()) {
+			String message = String.format("Inconsistent repository: %s.", inconsistentRepository.get().getCdoRepository().getRepositoryName());
+			throw new SnowOwlApplication.InitializationException(message);
```
cmark: Unfortunately, the current implementation prevents users/admins to perform `reindex` or other diagnostic/maintenance operations, since they depend on the services initialized by the boot process. Let's chat about this issue when you have time, @ekovacs, @apeteri!

core/com.b2international.snowowl.datastore.server/src/com/b2international/snowowl/datastore/server/internal/CDOBasedRepository.java
```diff
@@ -97,11 +101,13 @@
  */
 public final class CDOBasedRepository extends DelegatingServiceProvider implements InternalRepository, CDOCommitInfoHandler {
 
-	private static final String REINDEX_KEY = "snowowl.reindex";
+	private static final Logger LOG = LoggerFactory.getLogger(CDOBasedRepository.class);
```
cmark: Use `LoggerFactory.getLogger("repository")` instead of the implementation class.

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/Repository.java
```diff
@@ -56,4 +56,15 @@
...
+	/**
+	 * @return Returns the {@link RepositoryState state} for this repository. The repository is considerred healthy if it is in {@link RepositoryState#CONSISTENT} 
+	 */
+	RepositoryState getRepositoryState();
+	
+	/**
+	 * @since 5.8 
+	 */
+	enum RepositoryState {
+		UNKNOWN, CONSISTENT, INCONSISTENT, REINDEXING;
```
cmark: Use GREEN, YELLOW, RED states instead of the ones above.

apeteri: 🚥

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/Repository.java
```diff
@@ -56,4 +56,15 @@
 	 */
 	void sendNotification(RepositoryEvent event);
 
+	/**
+	 * @return Returns the {@link Health health} state for this repository. The repository is considered healthy if it is in {@link Health#GREEN} state.
```
cmark: Please remove the `Returns` word from `@return` part of the javadoc (this way it reads like `Returns: Returns...`).

core/com.b2international.snowowl.datastore.server/src/com/b2international/snowowl/datastore/server/internal/CDOBasedRepository.java
```diff
@@ -348,5 +321,102 @@ public void handleCommitInfo(CDOCommitInfo commitInfo) {
...
+		boolean validContent = !hasInvalidCDOTimeStamps(cdoCommitInfos, getLast(indexCommitInfos).getTimeStamp(), branchPath);
+		
+		return validContent == true ? Health.GREEN : Health.RED;
+	}
+
+	private String contentMessage(boolean empty) {
+		return empty ? "IS EMPTY" : "HAS CONTENT";
+	}
+
+	private boolean hasInvalidCDOTimeStamps(List<CDOCommitInfo> cdoCommitInfos, long lastIndexCommitTimestamp, IBranchPath branch) {
```
cmark: If you use only the `path` part of an `IBranchPath`, then it is recommended to use a String to represent your path.

core/com.b2international.snowowl.datastore.server/src/com/b2international/snowowl/datastore/server/internal/CDOBasedRepository.java
```diff
@@ -348,5 +321,102 @@ public void handleCommitInfo(CDOCommitInfo commitInfo) {
...
+		boolean emptyDb = cdoCommitInfos.isEmpty();
+		boolean emptyIndex = indexCommitInfos.getItems().isEmpty();
+
+		if (emptyDb && emptyIndex) {
+			return Health.GREEN; // empty dataset
+		}
+		
+		if (emptyDb ^ emptyIndex) {
+			LOG.error("{} is in inconsistent state. CDO {} but INDEX {}", getCdoRepository().getRepositoryName(), contentMessage(emptyDb), contentMessage(emptyIndex));
+			return Health.RED; // either CDO or index was deleted but not the other.
```
cmark: If the underlying SQL database is empty, then the state should be `Health.RED`, but if there are database commits, but no index commits, then allow developers to reindex the dataset. Use `Health.YELLOW` in this case.

core/com.b2international.snowowl.datastore.server/src/com/b2international/snowowl/datastore/server/reindex/ReindexRequestBuilder.java
```diff
@@ -41,4 +42,9 @@ public ReindexRequestBuilder setFailedCommitTimestamp(long failedCommitTimestamp
 		return req;
 	}
 
+	@Override
+	public Health[] allowedHealthstates() {
+		// allowing reindexing requests to execute in all health states.
+		return Health.values();
```
cmark: `Health.RED` should indicate major problems with the repository. Reject all requests sent to repositories in this state.

core/com.b2international.snowowl.datastore/src/com/b2international/snowowl/datastore/request/HealthCheckingRequest.java
```diff
@@ -0,0 +1,73 @@
+/*
+ * Copyright 2011-2017 B2i Healthcare Pte Ltd, http://b2i.sg
```
cmark: New files should indicate only the inception year in the license header.

core/com.b2international.snowowl.datastore/src/com/b2international/snowowl/datastore/request/HealthCheckingRequest.java
```diff
@@ -0,0 +1,73 @@
...
+
+/**
+ * A delegating request that provides repository state assertion for a given allowed {@link Repository.Health health} states, before the request proceeds to executing the delegate.
+ * @param <C>
+ *            - the required context type for this {@link Request}
+ * @param <B>
+ *            - the type of the result
+ * @since 5.8
+ */
+public class HealthCheckingRequest<C extends ServiceProvider, B> extends DelegatingRequest<C, ServiceProvider, B> {
```
cmark: `Request` implementations should be either `final` or `abstract`.

core/com.b2international.snowowl.datastore/src/com/b2international/snowowl/datastore/request/HealthCheckingRequest.java
```diff
@@ -0,0 +1,73 @@
...
+
+	private static final long serialVersionUID = 1L;
+	
+	private List<Health> allowedHealthStates;
+	private String repositoryId;
+
+	protected HealthCheckingRequest(String repositoryId, Request<ServiceProvider, B> next, Repository.Health... healths) {
```
cmark: Instead of creating two `RepositoryContext` instances, it would be better to nest this request into an existing `RepositoryRequest` like this:

```
new RepositoryRequest(id, new HealthCheckRequest(next));
```

core/com.b2international.snowowl.datastore/src/com/b2international/snowowl/datastore/request/HealthCheckingRequest.java
```diff
@@ -0,0 +1,73 @@
...
+		if (!getAllowedHealthStates().contains(repositoryHealth)) {
+			throw new NotAvailableException("Requests for this repository [{}] are not allowed to execute with health state: {}.", repository.id(), repositoryHealth);
+		}
+	}
+
+	private Repository getRepository(C context) {
+		return context.service(RepositoryManager.class).get(repositoryId);
+	}
+	
+	private List<Health> getAllowedHealthStates() {
```
cmark: Remove private getters.

core/com.b2international.snowowl.server.console/src/com/b2international/snowowl/server/console/MaintenanceCommandProvider.java
```diff
@@ -58,6 +65,7 @@
 	private static final String LISTBRANCHES_COMMAND = "listbranches";
 	private static final String LISTREPOSITORIES_COMMAND = "listrepositories";
 	private static final String DBCREATEINDEX_COMMAND = "dbcreateindex";
+	private static final String REPOSITORY_HEALTH_COMMAND = "repositoryhealth";
```
cmark: Listing repositories should include the current health status. I suggest removing this new `repositoryHealth` command and instead including the status in the `listrepositories` command.

core/com.b2international.snowowl.api.rest/src/com/b2international/snowowl/api/rest/admin/RepositoryRestService.java
```diff
@@ -69,7 +67,7 @@
 	public @ResponseBody DeferredResult<Repositories> getRepositories(
 			@ApiParam
 			@RequestParam(value="id", required=false)
-			List<String> idFilter) {
+			String[] idFilter) {
```
apeteri: It seems to work in `SnomedReferenceSetMemberRestService#search`, can't really tell why; information on this topic is contradictory.

cmark:
👍

PR #145, opened 2017-04-10
Branch Compare Java API

cmark:
This PR adds the following changes to support proper branch compare mechanism:
* SHA-1 `_hash` has been added to all documents. It's a unique key to the actual semantic content of a document. It will be used by revision compare to filter out false positive changed revisions. The time required to compute the value for this field is `~1ms` depending on the size of the document. 
* `^` character has been added to `RevisionIndex.read(String, RevisionIndexRead)` method to support reading revisions and content from the base of a branch (basically this is the state before the branch fork point). For example to read the original version of a SNOMED CT concept from a branch, you can execute
`GET /snowowl/snomed-ct/v2/MAIN/MyAwesomeBranch^/concepts/123456789`.
* `RepositoryRequests.branching().prepareCompare()` has been added as the new primary branch compare API (replacing the old review API, which is still available, but will be deprecated in favor of the new one). The difference between the two APIs, that the new compare API returns the identifiers of the new/changed/deleted components, and it does not try limit them to some sort of container ID. This will work properly in case of non-SNOMED CT terminologies as well. Also it fetches only the identifiers (using the latest partial doc. loading improvements), so it's faster than fetching all fields.


commons/com.b2international.index.api/src/com/b2international/index/revision/DefaultRevisionIndex.java
```diff
@@ -109,65 +125,64 @@ private RevisionCompare compare(final RevisionBranch base, final RevisionBranch
...
-				final Map<Class<? extends Revision>, LongSet> newAndChangedComponents = newHashMap();
-				final Map<Class<? extends Revision>, LongSet> deletedAndChangedComponents = newHashMap();
+				final Multimap<Class<? extends Revision>, Revision.Views.StorageKeyAndHash> newAndChangedComponents = ArrayListMultimap.create();
+				final Multimap<Class<? extends Revision>, Revision.Views.StorageKeyAndHash> deletedAndChangedComponents = ArrayListMultimap.create();
 				
 				// query all registered revision types for new, changed and deleted components
 				for (Class<? extends Revision> typeToCompare : typesToCompare) {
-					final Query<Revision.Views.StorageKeyOnly> newAndChangedQuery = Query
-							.selectPartial(Revision.Views.StorageKeyOnly.class, typeToCompare)
+					final Query<Revision.Views.StorageKeyAndHash> newAndChangedQuery = Query
```
apeteri: Please collapse the boolean connectivity words 😄 Both make sense to me, but let's use either eg. `newAndChanged` or `newOrChanged`.

commons/com.b2international.index.api/src/com/b2international/index/revision/Revision.java
```diff
@@ -76,6 +90,11 @@ public long getStorageKey() {
 	public static final String COMMIT_TIMESTAMP = "commitTimestamp";
 	public static final String SEGMENT_ID = "segmentId";
 	public static final String REPLACED_INS = "replacedIns";
+	
+	/**
+	 * Revision fields that should not be part of any hash value.
+	 */
+	public static final Set<String> REV_FIELDS = ImmutableSet.of(BRANCH_PATH, COMMIT_TIMESTAMP, SEGMENT_ID, REPLACED_INS);
```
apeteri: Are `_id` and `_hash` not in play when the hash is computed?

commons/com.b2international.index.lucene/src/com/b2international/index/json/JsonDocumentSearcher.java
```diff
@@ -236,25 +237,43 @@ public void close() throws Exception {
...
+		
 		// if select is a different type, then use that as JsonView on from, otherwise select all props
 		final ObjectReader reader = select != from 
 				? mapper.reader(select).without(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES) 
 				: mapper.reader(select);
 		
 		final String[] ids = new String[scoreDocs.length - offset];
 		final byte[][] sources = new byte[scoreDocs.length - offset][];
+		final String[] hashes;
+		if (isWithHash) {
```
apeteri: `ids[]` could get the same treatment above.

core/com.b2international.snowowl.datastore/src/com/b2international/snowowl/datastore/request/compare/BranchCompareRequest.java
```diff
@@ -0,0 +1,113 @@
...
+import com.b2international.snowowl.core.branch.BranchManager;
+import com.b2international.snowowl.core.domain.RepositoryContext;
+import com.b2international.snowowl.core.events.Request;
+import com.b2international.snowowl.datastore.index.RevisionDocument;
+import com.google.common.collect.ImmutableSet;
+
+/**
+ * @since 5.9
+ */
+final class BranchCompareRequest implements Request<com.b2international.snowowl.core.domain.RepositoryContext, CompareResult> {
```
apeteri: Is the type parameter fully qualified for some special reason?

core/com.b2international.snowowl.datastore/src/com/b2international/snowowl/datastore/request/compare/BranchCompareRequest.java
```diff
@@ -0,0 +1,113 @@
...
+		this.baseBranch = baseBranch;
+	}
+	
+	void setCompareBranch(String compareBranch) {
+		this.compareBranch = compareBranch;
+	}
+	
+	@Override
+	public CompareResult execute(RepositoryContext context) {
+		final RevisionIndex index = context.service(RevisionIndex.class);
```
apeteri: Is it possible to wrap the request in a remote job externally?

commons/com.b2international.index.lucene/src/com/b2international/index/json/Index.java
```diff
@@ -199,7 +209,7 @@ private void addToDoc(final Document doc, final String name, final JsonNode node
 			final Field field = mapping.getField(name);
 			final Class<?> fieldType = NumericClassUtils.unwrapCollectionType(field);
 			
-			if (docValues && NumericClassUtils.isCollection(field)) {
+			if (docValues && NumericClassUtils.isCollection(fieldType)) {
```
apeteri: `fieldType` is now an eg. `int` if the field's type was `IntCollection` because of the unwrapping above, so this check will not throw for the cases that we expect it to.

commons/com.b2international.index.api/src/com/b2international/index/mapping/DocumentMapping.java
```diff
@@ -154,6 +155,13 @@ public Field getField(String name) {
 		return fieldMap.get(name);
 	}
 	
+	public Class<?> getFieldType(String key) {
```
apeteri: [LuceneQueryBuilder#visit(ScoreFunction)](https://github.com/b2ihealthcare/snow-owl/blob/develop/commons/com.b2international.index.lucene/src/com/b2international/index/query/LuceneQueryBuilder.java#L172) would benefit from this shortened method as well.

commons/com.b2international.index.api/src/com/b2international/index/revision/Revision.java
cmark: They are not part of the `_source` value, and the _hash is computed from that by removing the metafields (except the `storageKey`) from the revision document.

core/com.b2international.snowowl.datastore/src/com/b2international/snowowl/datastore/request/compare/BranchCompareRequest.java
cmark: Eclipse autocomplete magic!

commons/com.b2international.index.lucene/src/com/b2international/index/json/Index.java
cmark: Ohh, I see, then I revert this change back. Thanks!

commons/com.b2international.index.api/src/com/b2international/index/revision/DefaultRevisionIndex.java
cmark: Makes sense. I'll use `newOrChanged` then, because a revision can not be `new` and `changed` at the same time.

cmark:
Thanks @apeteri for the review! Merging to develop.

PR #147, opened 2017-04-23
Elasticsearch support

cmark:
This PR adds support for **experimental** elasticsearch support.

Dependency changes:
* `Elasticsearch` `2.3.3` and all its required dependencies have been added to the Lucene wrapper bundle
* Added `com.fasterxml.jackson.dataformat.jackson-dataformat-smile`
* Added `org.jboss.netty`
* Added `com.ning.compress-lzf`
* Bumped `Guava` to `18.0`
* Bumped `Jackson` to `2.6.6`
* Bumped `snakeyaml` to `1.15`

New module:
* `com.b2international.index.es` - Elasticsearch implementation of the index API

Index API changes:
* Script based bulk update operations (Script annotation is available for doc types to define the script they will use for bulk updates)
* Elasticsearch cannot handle BigDecimals properly, but fortunately we already do it in the Lucene implementation via HBase's OrderedBytes. To allow ES to use BigDecimal sorting, the byte value needed to be converted to an ordered Base64 encoded String (all tests pass).
  * The index API now enforces this conversion via a Jackson module for all BigDecimal fields (affects the Lucene implementation as well)
* The implementation does not support the SHA-1 `_hash` fields on revision documents thus it cannot skip them from the compare result and will return them, clients should query the base and head of the branch and remove the false positive hits manually.
* The Lucene implementation now adds doc value fields  for the `_id` field 
* Improved performance of revision purge operation by loading only the `_id` fields from the index
* Custom scoring now uses the Script annotations as well

Index layout changes:
* The SNOMED CT Identifier index had to be moved to its own index due to mapping conflicts with Remote Job and other document mappings

Doc Mapping changes:
* `SnomedConceptDocument.referencedComponentType` now uses `short`
* `SnomedConstraintDocument.type` changed to `predicateType`

RESTful API changes:
* JSON representations now include non-null values instead of non-empty

Configuration changes:
* It is possible to configure the underlying Elasticsearch instance via a `elasticsearch.yml` file. (put the configuration file into the `configuration` directory). This the underlying elasticsearch instance can be configured in a way to communicate with a cluster of nodes and use them for the terminology servers index.
* Added `numberOfShared` configuration option to repository.index node in snowowl_config.yml
* Added `commitConcurrencyLevel` configuration option to repository.index node in snowowl_config.yml


commons/com.b2international.index.es/src/com/b2international/index/EsDocumentWriter.java
```diff
@@ -0,0 +1,225 @@
...
+					admin.log().debug("Successfully processed bulk request");
+					if (response.hasFailures()) {
+						for (BulkItemResponse itemResponse : response.getItems()) {
+							checkState(!itemResponse.isFailed(), "Failed to commit bulk request in index '%s', %s", admin.name(), itemResponse.getFailureMessage());
+						}
+					}
+				}
+			})
+			.setConcurrentRequests(Math.max(1, Runtime.getRuntime().availableProcessors() / 4))
+			.setBulkActions(5000)
```
nagyo: magic numbers 🎱 -> constant?

commons/com.b2international.index.lucene/src/com/b2international/index/json/BulkUpdateOperation.java
```diff
@@ -48,10 +57,30 @@ public BulkUpdateOperation(BulkUpdate<T> update, ObjectMapper mapper, Mappings m
 	@Override
 	public void execute(IndexWriter writer, Searcher searcher) throws IOException {
 		final DocumentMapping mapping = mappings.getMapping(update.getType());
+		final String scriptName = update.getScript();
+		final String script = mapping.getScript(scriptName).script();
+		
+		final GroovyShell shell = new GroovyShell();
```
apeteri: We should limit access from scripts to bound variables, simple arithmetic operations, a limited set of utility methods (eg. `java.lang.Math`) when the feature comes out of experimental status.

commons/com.b2international.index.lucene/src/com/b2international/index/query/ScoreValueSource.java
```diff
@@ -0,0 +1,80 @@
...
+				} catch (IOException exception) {
+					throw new RuntimeException(exception);
+				}
+			}
+		};
+	}
+
+	@Override
+	public boolean equals(Object o) {
+		return o == this;
```
apeteri: Instead of reference equality, we should change the implementation to use the `query` field in `equals` and `hashCode` computations.

commons/com.b2international.index.es/src/com/b2international/index/admin/EsIndexAdmin.java
```diff
@@ -0,0 +1,271 @@
...
+				ImmutableMap.builder()
+					.put("_all", ImmutableMap.of("enabled", false))
+					.putAll(toProperties(mapping))
+					.build()
+			);
+			esMappings.put(type, typeMapping);
+ 		}
+ 		
+		for (String type : esMappings.keySet()) {
+			final Map<String, Object> typeMapping = esMappings.get(type).stream().sorted((m1, m2) -> -1 * Ints.compare(m1.toString().length(), m2.toString().length())).findFirst().get();
```
cmark: Longer ones win over shorter ones. It's a compatibility hack to allow mappings to still specify both the interface and the concrete class for the index mappings.

commons/com.b2international.index.es/src/com/b2international/index/admin/EsIndexAdmin.java
```diff
@@ -0,0 +1,271 @@
...
+	public void delete() {
+		if (exists()) {
+			DeleteIndexResponse response = client().admin().indices().prepareDelete(name).get();
+			checkState(response.isAcknowledged(), "Failed to delete index %s", name);
+		}
+	}
+
+	@Override
+	public <T> void clear(Class<T> type) {
+		// TODO remove all documents matching the given type, based on mappings
```
cmark: Nope.

commons/com.b2international.index.es/src/com/b2international/index/query/EsQueryBuilder.java
```diff
@@ -0,0 +1,305 @@
...
+		needsScoring = true;
+		deque.push(QueryBuilders
+				.functionScoreQuery(innerQuery, ScoreFunctionBuilders.scriptFunction(script))
+				.boostMode(CombineFunction.REPLACE));
+	}
+	
+	private void visit(BoolExpression bool) {
+		final BoolQueryBuilder query = QueryBuilders.boolQuery();
+		query.disableCoord(true);
+		// first add the mustClauses, then the mustNotClauses, if there are no mustClauses but mustNot ones then add a match all before
```
cmark: Copy-paste bug!!! 😄

commons/com.b2international.index.es/src/com/b2international/index/query/EsQueryBuilder.java
```diff
@@ -0,0 +1,305 @@
...
+	}
+
+	private void visit(TextPredicate predicate) {
+		final String field = toFieldPath(predicate);
+		final String term = predicate.term();
+		final MatchType type = predicate.type();
+		QueryBuilder query;
+		switch (type) {
+		case PHRASE:
+			{
```
cmark: Will fix it later on develop.

commons/com.b2international.index.es/src/com/b2international/index/query/EsQueryBuilder.java
```diff
@@ -0,0 +1,305 @@
...
+		case FUZZY:
+			query = QueryBuilders.fuzzyQuery(field, term).fuzziness(Fuzziness.ONE).prefixLength(1);
+			break;
+		case PARSED:
+			query = QueryBuilders.queryStringQuery(TextConstants.escape(term)).allowLeadingWildcard(true).defaultOperator(Operator.AND);
+			break;
+		default: throw new UnsupportedOperationException("Unexpected text match type: " + type);
+		}
+		if (query == null) {
+			query = MATCH_NONE;
```
cmark: Yup, copied from Lucene implementation. 😄

cmark:
Thanks for the reviews and the help @nagyo and @apeteri. 🏆

PR #150, opened 2017-06-07
SO-2485: Generate inferred relationship IDs in bulk

apeteri:
https://snowowl.atlassian.net/browse/SO-2485


snomed/com.b2international.snowowl.snomed.reasoner.server/src/com/b2international/snowowl/snomed/reasoner/server/diff/SourceConceptNamespaceAndModuleAssigner.java
```diff
@@ -15,92 +15,77 @@
...
-	public void allocateRelationshipNamespacesAndModules(Set<String> conceptIds, final SnomedEditingContext editingContext) {
-
-		//default module is the fall back
-		relationshipDefaultModuleConcept = editingContext.getDefaultModuleConcept();
-
+	public void allocateRelationshipIdsAndModules(Multiset<String> conceptIds, final SnomedEditingContext editingContext) {
+		Multiset<String> reservedIdsByNamespace = HashMultiset.create();
+		for (Multiset.Entry<String> conceptIdWithCount : conceptIds.entrySet()) {
+			String namespace = SnomedIdentifiers.create(conceptIdWithCount.getElement()).getNamespace();
+			reservedIdsByNamespace.add(namespace, conceptIdWithCount.getCount());
```
apeteri: `INT` namespaces are registered in the `Multiset` with a `null` key. It doesn't seem to hurt (the underlying `HashMap` allows using `null` as a key), but in other parts of the code we use the String literal `"INT"` for increased readability. Should I change it here as well?

cmark: I would change the `SnomedIdentifier.getNamespace()` to not return `null`, but the `INT` literal in the case of INT namespace. But that would require additional changes in other places as well, you can leave it like this for now.

PR #154, opened 2017-08-03
SDD-770: Fix NPE thrown while saving classification results

AAAlinaaa:
Fix NPE thrown while saving classification results of equivalent singapore medicinal product descendants:
 1. Include stated relationships in checking for equivalent concepts
that need to be fixed
 2. Filter out inboun redundant relationships while fixing equivalent
concepts
 3. Avoid attempt to delete already deleted relationships in the
deletion plan
 4. Allow for deletion plan completion even if delete items are empty


nagyo (approved):
OK.

PR #157, opened 2017-08-17
Index Lucene fragment groovy memory leak fix

cmark:
This PR fixes a possible memory leak found in the Groovy script evaluation inside the Lucene based index implementation. 
The implementation now caches the compiled script class for later use and instantiates a new instance every time the code would like to execute the script.

### DIFF
--- a/commons/com.b2international.index.api.tests/src/com/b2international/index/GroovyMemoryLeakTest.java
+++ b/commons/com.b2international.index.api.tests/src/com/b2international/index/GroovyMemoryLeakTest.java
@@ -27,6 +27,7 @@ import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
 
 import org.apache.commons.lang.RandomStringUtils;
+import org.junit.Ignore;
 import org.junit.Test;
 
 import com.b2international.index.Fixtures.Data;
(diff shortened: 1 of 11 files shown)


commons/com.b2international.index.api.tests/src/com/b2international/index/GroovyMemoryLeakTest.java
```diff
@@ -0,0 +1,92 @@
...
+
+/**
+ * @since 5.10.11
+ */
+public class GroovyMemoryLeakTest extends BaseIndexTest {
+
+	private static final int NUM_DOCS = 10_000;
+	
+	@Test
+	public void tryToGenerateMemoryLeak() throws Exception {
```
apeteri: It might be worth noting here in a comment that the test may just "fizzle out" because of memory exhaustion instead of actually failing. Or does it now fail reliably with the time limit set?

commons/com.b2international.index.lucene/src/com/b2international/index/admin/BaseLuceneIndexAdmin.java
```diff
@@ -210,6 +218,23 @@ public IndexSearcher newSearcher(IndexReader reader, IndexReader previousReader)
...
+		final Class<? extends Script> scriptClass = getScriptClass(script);
+		try {
+			return scriptClass.newInstance();
+		} catch (InstantiationException | IllegalAccessException e) {
+			throw new IndexException("Couldn't instantiate groovy script", e);
+		}
+	}
+	
+	private Class<? extends Script> getScriptClass(String script) {
+		if (!scriptCache.containsKey(script)) {
```
apeteri: I'd still love to see `LoadingCache` here, otherwise two threads can find that the key is not present in the map.

apeteri (approved):

commons/com.b2international.index.api.tests/src/com/b2international/index/GroovyMemoryLeakTest.java
cmark: Unfortunately, it does not fail reliably and consistently. 😿 
The timeout is there to ensure that the VM will shut down eventually.
We can play with the numbers (number of docs, number of threads and the timeout) to find the ones that reliably fail in case of memleak and pass when the fix prevents the memleak. 
Or I just write a nice guide how to detect the memory leak and how to verify the fix.

apeteri: It's OK to document it only.

nagyo (approved):
👍

cmark:
Thanks @apeteri and @nagyo.

PR #160, opened 2017-08-18
SO-1830: Make equivalent concepts siblings in the inferred taxonomy

apeteri:
https://snowowl.atlassian.net/browse/SO-1830


cmark (approved):
LGTM!

PR #161, opened 2017-08-29
SDD-788: Improve relationship CD member handling in distribution normal form

apeteri:


apeteri:
`DataHasValue` expressions are still not used because ELK emits a warning in the log each time it is encountered, see https://github.com/liveontologies/elk-reasoner/blob/a3531b73651e4306c2b4e7330dc6092522f41617/elk-reasoner/src/main/java/org/semanticweb/elk/reasoner/stages/InputLoadingStage.java#L336

apeteri:
I tried to change the base branch, but it added a lot of commits to the list, so it is still on `4.x` at the moment.

apeteri:
Obsoleted by changes on #288.

PR #162, opened 2017-08-29
SO-2613: Branch merging changes

apeteri:
- Reverse change set comparison order in DefaultCDOMerger
- "Flip" handling of conflicts which now happen from the opposing
branch's point of view
- Move CDO-to-domain level conflict object mapping from utility classes
into conflict processors
- Update test cases


apeteri:
We have decided to go with a different approach in the meantime.

PR #163, opened 2017-08-31
SO-2600 permission enhancements

cmark:
This PR removes all previous permission constants and adds a new set of constants (six to be precise) to use throughout the server.
Also removed the client side service implementation of the `IAuthorizationService` class.


core/com.b2international.snowowl.datastore.server/src/com/b2international/snowowl/datastore/server/session/ApplicationSessionManager.java
```diff
@@ -235,6 +227,10 @@ public void loginWithResponse(final byte[] response) throws SecurityException {
 
 			final String password = new String(decryptedResponse, RANDOM_BYTES_LENGTH, decryptedResponse.length - RANDOM_BYTES_LENGTH, Charsets.UTF_8);
 			final Collection<Role> roles = authenticate(userId, password);
+			
+			if (!loginEnabled && !roles.contains(SpecialRole.ADMINISTRATOR)) {
```
apeteri, line 231: Wondering if we should ever refer to roles directly instead of permissions... 🤔 This use case would require an "administrative login" permission, though.

apeteri (approved):
LGTM

nagyo (approved):

cmark:
Thanks, @apeteri and @nagyo. 😄

PR #167, opened 2017-10-04
SDD-775: Convert escg refset query (^) from concept to member search...

AAAlinaaa:
...and ensure multiple memberships in the same refset can be handled by
the SnomedRf2IndexInitializer
Note: Active members with inactive referenced components will now show
up among the search results


nagyo (approved):
LGTM!

PR #173, opened 2017-10-12
SO-2662: single refset export fix

cmark:
Allow exported content filtering by component type.
Single reference set only export will happen when you select a single
refset + refset member type.
Also remove the restriction to allow a single export at any given point
in time.

### DIFF
--- a/snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/request/rf2/SnomedRf2ExportRequestBuilder.java
+++ b/snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/request/rf2/SnomedRf2ExportRequestBuilder.java
@@ -31,6 +31,7 @@ import com.b2international.snowowl.snomed.core.domain.Rf2ReleaseType;
  */
 public final class SnomedRf2ExportRequestBuilder extends BaseRequestBuilder<SnomedRf2ExportRequestBuilder, BranchContext, UUID> implements RevisionIndexRequestBuilder<UUID> {
 
+	private String userId;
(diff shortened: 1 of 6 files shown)


snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/request/rf2/SnomedRf2ExportRequestBuilder.java
```diff
@@ -41,9 +42,15 @@
 	private boolean extensionOnly;
 	private String namespace;
 	private Collection<String> refSets = Collections.emptySet();
+	private Collection<String> componentTypes = Collections.emptySet();
```
apeteri, line 45: Does the default empty set mean that everything will be exported?

apeteri: I guess it doesn't matter right now 😃

apeteri (approved):
LGTM otherwise!

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/request/rf2/SnomedRf2ExportRequestBuilder.java
cmark: Yes, right now we use it only to detect single refset export. It can be used later to filter the content by component type, unless we rewrite the whole export. 😄

nagyo (approved):
[LGTM](http://lgtm.in/i/ajRa04V4q)

PR #179, opened 2017-11-27
SO-2726 DSV export fix: add description of ExportItem to exportSetting..

mucst:
.. when includeDescriptionId is indicated as true in order to eliminate column shift bug.


cmark:
According to the DSV export specification, the changes of this PR are incorrect. The `includeDescriptionId` boolean flag controls whether the selected Description-like columns should also have an id subcolumn along with the default term column. Closing.

PR #180, opened 2017-11-27
SO-2191: Improve sequential item identifier generation

apeteri:
Commits for a similar issue in the 4.x release are adapted to the current development branch in this PR.


cmark (approved):
LGTM!

PR #184, opened 2017-12-12
[sync] Fix wrong HTTP calls in CisSnomedIdentifierService

nagyo:


apeteri (approved):
Cross-checked HTTP request methods against https://github.com/IHTSDO/component-identifier-service/blob/master/api/swagger-ids.json , looks good to me!

PR #186, opened 2017-12-12
[sync] Delete description by removing the container reference

nagyo:
- this will allow removing descriptions that are not in the concept's
description list but were resurrected by certain stale branch operations

https://jira.ihtsdotools.org/browse/INFRA-1901


cmark (approved):
LGTM!

PR #193, opened 2018-01-24
[core.domain] Make IComponent an IdProvider

AAAlinaaa:


apeteri (changes requested):
"core.domain" and "core.api" views of components would become intermixed with the addition of this interface, which we'd like to keep strictly separated.

apeteri:
Following discussion, the change could be avoided by fixing a call site in a different project. Closing PR.

PR #199, opened 2018-02-21
Term based concept sorting support

cmark:
* This PR adds new term based sorting functionality to the SNOMED CT Concept Search API. To accomplish this kind of sorting properly the index format needed to be changed. 
From now on a concept document will contain description fragments (id, storageKey, term, typeId, languageRefSetIds) to allow scripts to access these data for sorting.
* This PR also changes how `pt()` and `fsn()` expansion works. Expanding `pt()` or `fsn()` now won't try to look for the description in the index but use the data indexed in the description fragments array on the concept. This will improve `pt()` and `fsn()` expansion performance significantly.
* This PR also changes `referringMembers` and `referringMappingMembers` index fields. These used to track mapping and non-mapping active membership of all SNOMED CT core components to easily search for whether a component is an active member of a reference set or not. Unfortunately, these fields can't be used to search inactive membership efficiently. To support this, SNOMED CT component documents now has `memberOf` and `activeMemberOf` fields instead of `referringMembers` and `referringMappingMembers`. `memberOf` field tracks all memberships including active and inactive while `activeMemberOf` track only active memberships. The new fields track all reference set memberships in these fields, the distinction of mapping vs. non-mapping has been removed.
* This PR also adds a new CDORevisionCache implementation to reduce memory overhead when the server is using a revision cache. The new cache can be configured to evict cached revision based on time and size (will be added soon). This PR changes the default cache to this implementation `EvictingRevisionCache`.


commons/com.b2international.index.api/src/com/b2international/index/Script.java
```diff
@@ -30,20 +30,13 @@
...
-	 * @return
+	 * @return the script itself
 	 */
 	String script();
 	
-	/**
-	 * Optionally specify the fields to be loaded for the script
-	 * @return
-	 */
-	String[] fields() default {};
```
apeteri, line 47: Is this capability (to only use certain fields in a script) no longer useful to have?

commons/com.b2international.index.api/src/com/b2international/index/query/SortBy.java
```diff
@@ -127,27 +173,25 @@ public String toString() {
...
+		private final List<SortBy> sorts = newArrayList();
 		
-		public Builder add(String field, Order order) {
-			sortOrderMap.put(field, order);
+		public Builder sortByField(String field, Order order) {
+			sorts.add(new SortByField(field, order));
+			return this;
+		}
+		
+		public Builder byScript(String script, Map<String, Object> arguments, Order order) {
```
apeteri: The two builder methods should have consistent naming. Either this needs the `sort-` prefix or the other needs to be cut.

commons/com.b2international.index.api/src/com/b2international/index/query/SortBy.java
```diff
@@ -127,27 +173,25 @@ public String toString() {
 	}
 	
 	public static final class Builder {
-		private final Map<String, Order> sortOrderMap = Maps.newLinkedHashMap();
+		private final List<SortBy> sorts = newArrayList();
 		
-		public Builder add(String field, Order order) {
-			sortOrderMap.put(field, order);
+		public Builder sortByField(String field, Order order) {
+			sorts.add(new SortByField(field, order));
```
apeteri: The factory method for the field-based `SortBy` should be used here.

commons/com.b2international.index.es/src/com/b2international/index/EsDocumentSearcher.java
```diff
@@ -254,25 +256,47 @@ public void cancelScroll(String scrollId) {
...
+				default:
+					req.addSort(SortBuilders.fieldSort(field).order(sortOrder));
+				}
+			} else if (item instanceof SortByScript) {
+				SortByScript sortByScript = (SortByScript) item;
+				SortBy.Order order = sortByScript.getOrder();
+				String script = sortByScript.getName();
+				SortOrder sortOrder = order == SortBy.Order.ASC ? SortOrder.ASC : SortOrder.DESC;
+				
+				// if this is a named script the get it from the current mapping
```
apeteri: ⚠️ the &rarr; then

core/com.b2international.snowowl.datastore.server/src/com/b2international/snowowl/datastore/server/EvictingRevisionCache.java
```diff
@@ -0,0 +1,368 @@
...
+
+		synchronized (revisionLists) {
+			RevisionList list = revisionLists.getIfPresent(key);
+			if (list == null) {
+				list = new RevisionList();
+				revisionLists.put(key, list);
+			}
+
+			if (list.addRevision((InternalCDORevision) revision)) {
+				typeRefIncrease(id, revision.getEClass());
```
apeteri, line 189: There seems to be an imbalance in the typeRef counter, here it is registering all revisions that refer to a CDOID-EClass pair...

core/com.b2international.snowowl.datastore.server/src/com/b2international/snowowl/datastore/server/EvictingRevisionCache.java
```diff
@@ -0,0 +1,368 @@
...
+
+	public InternalCDORevision removeRevision(CDOID id, CDOBranchVersion branchVersion) {
+		Object key = createKey(id, branchVersion.getBranch());
+		synchronized (revisionLists) {
+			RevisionList list = revisionLists.getIfPresent(key);
+			if (list != null) {
+				list.removeRevision(branchVersion.getVersion());
+				if (list.isEmpty()) {
+					revisionLists.invalidate(key);
+					typeRefDecrease(id);
```
apeteri: ...but here the counter is only decremented on the removal of the last revision, once.

core/com.b2international.snowowl.datastore.server/src/com/b2international/snowowl/datastore/server/EvictingRevisionCache.java
```diff
@@ -0,0 +1,368 @@
...
+				} else if (v < version) {
+					break;
+				}
+			}
+
+			return null;
+		}
+
+		public synchronized boolean addRevision(InternalCDORevision revision) {
+			if (revision.getBranch().getID() == 34 && revision.getVersion() == 2 && revision.getEClass().getName().equals("Concept")) {
```
apeteri: ⚠️ Diagnostic code snippet

core/com.b2international.snowowl.datastore/src/com/b2international/snowowl/datastore/CDOCommitChangeSet.java
```diff
@@ -130,17 +139,88 @@ public long getTimestamp() {
...
+				.filter(input -> eClass.isSuperTypeOf(input.getValue()))
+				.transform(input -> input.getKey())
+				.toSet();
+	}
+	
+	@Override
+	public <T extends CDOObject> Iterable<T> getDirtyComponents(Class<T> type, Set<EStructuralFeature> allowedFeatures) {
+		return FluentIterable.from(getDirtyComponents(type))
+			.filter(input -> {
+				final DirtyFeatureDeltaVisitor visitor = new DirtyFeatureDeltaVisitor(allowedFeatures);
```
apeteri, line 152: Not sure how much it impacts performance, but if the visitor's state could be reset to "no changes" after each check, it could be created once and moved to a local variable.

core/com.b2international.snowowl.datastore/src/com/b2international/snowowl/datastore/CDOCommitChangeSet.java
```diff
@@ -130,17 +139,88 @@ public long getTimestamp() {
...
+			visitDelta(delta);
+		}
+		
+		@Override
+		public void visit(CDOContainerFeatureDelta delta) {
+			visitDelta(delta);
+		}
+		
+		private void visitDelta(CDOFeatureDelta delta) {
+			hasAllowedChanges |= allowedFeatures.contains(delta.getFeature());
```
apeteri, line 217: `CDOFeatureDeltaVisitorImpl` uses an exception to jump quickly out of list feature delta visiting if not all list deltas need to be examined. While this is generally considered an anti-pattern, it is accepted as way of unwinding the call stack if other options are not available: https://blog.jooq.org/2013/04/28/rare-uses-of-a-controlflowexception/

We might want to do the same here, and then `DirtyFeatureDeltaVisitor` would be stateless. If it runs all the way through, it's a negative answer, if it throws an exception at the first feature delta which refers to an interesting feature, it's a positive answer.

snomed/com.b2international.snowowl.snomed.datastore.tests/src/com/b2international/snowowl/snomed/datastore/index/change/BaseChangeProcessorTest.java
```diff
@@ -127,6 +131,16 @@ protected final void registerSetRevisionDelta(CDOObject object, EStructuralFeatu
 		getRevisionDelta(object).addFeatureDelta(featureDelta);
 	}
 	
+	protected final void registerAddRevisionDelta(CDOObject object, EStructuralFeature feature, int index, Object value) {
+		final CDOAddFeatureDeltaImpl featureDelta = new CDOAddFeatureDeltaImpl(feature, 0, value);
```
apeteri: `index` is passed in as an argument, but 0 is used in the add feature delta.

snomed/com.b2international.snowowl.snomed.datastore.tests/src/com/b2international/snowowl/snomed/datastore/index/change/BaseChangeProcessorTest.java
```diff
@@ -127,6 +131,16 @@ protected final void registerSetRevisionDelta(CDOObject object, EStructuralFeatu
...
+		getRevisionDelta(object).addFeatureDelta(featureDelta);
+	}
+	
+	protected final void registerRemoveRevisionDelta(CDOObject object, EStructuralFeature feature, int index) {
+		final CDORemoveFeatureDeltaImpl featureDelta = new CDORemoveFeatureDeltaImpl(feature, 0);
```
apeteri: `index` is passed in as an argument, but 0 is used in the remove delta.

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/converter/SnomedConceptConverter.java
```diff
@@ -143,6 +146,19 @@ protected SnomedConcept toResource(final SnomedConceptDocument input) {
 		if (input.getStatedAncestors() != null) {
 			result.setStatedAncestorIds(input.getStatedAncestors().toArray());
 		}
+		
+		if (expand().containsKey("preferredDescriptions") || expand().containsKey(SnomedConcept.Expand.PREFERRED_TERM) || expand().containsKey("fsn()")) {
```
apeteri: `SnomedConcept.Expand.FULLY_SPECIFIED_NAME` is already defined; `preferredDescriptions` could be added there as well.

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/converter/SnomedConceptConverter.java
```diff
@@ -206,8 +231,17 @@ private void expandPreferredTerm(List<SnomedConcept> results, final Set<String>
 		if (!expand().containsKey(SnomedConcept.Expand.PREFERRED_TERM)) {
 			return;
 		}
-			
-		final Map<String, SnomedDescription> terms = getDescriptionHelper().getPreferredTerms(conceptIds, locales());
+		
+		final List<SnomedDescription> synonyms = newArrayList();
+		for (SnomedConcept result : results) {
+			for (SnomedDescription description : result.getPreferredDescriptions()) {
+				if (!Concepts.FULLY_SPECIFIED_NAME.equals(description.getTypeId())) {
```
apeteri, line 223: Skip `Concepts.TEXT_DEFINITION`s here as well.

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/index/change/ConceptChangeProcessor.java
```diff
@@ -53,38 +45,64 @@
...
+			.add(SnomedPackage.Literals.DESCRIPTION__TERM)
+			.add(SnomedPackage.Literals.DESCRIPTION__TYPE)
+			.add(SnomedPackage.Literals.DESCRIPTION__LANGUAGE_REF_SET_MEMBERS)
+			.build();
+	private static final Set<EStructuralFeature> ALLOWED_LANG_MEMBER_CHANGE_FEATURES = ImmutableSet.<EStructuralFeature>builder()
+			.add(SnomedRefSetPackage.Literals.SNOMED_REF_SET_MEMBER__ACTIVE)
+			.add(SnomedRefSetPackage.Literals.SNOMED_LANGUAGE_REF_SET_MEMBER__ACCEPTABILITY_ID)
+			.build();
+	
+	private static final Ordering<SnomedDescriptionFragment> DESCRIPTION_FRAGMENT_ORDER = Ordering.natural()
```
apeteri: If I'm not reading it the wrong direction, this `Ordering` seems to be sorting by storage key first. But this means that the rest of the ordering steps are wasted, because the storage key is unique, so there will be no ties.

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/index/change/ReferringMemberChangeProcessor.java
```diff
@@ -77,10 +74,8 @@ public ReferringMemberChangeProcessor(final short referencedComponentType) {
 							throw new RuntimeException("Unknown old value type: " + delta.getOldValue());
 						}
 						final Boolean newValue = (Boolean) delta.getValue();
-						if (Boolean.TRUE == oldValue && Boolean.FALSE == newValue) {
-							addChange(memberChanges, dirtyMember, MemberChangeKind.REMOVED);
-						} else if (Boolean.FALSE == oldValue && Boolean.TRUE == newValue) {
-							addChange(memberChanges, dirtyMember, MemberChangeKind.ADDED);
+						if ((Boolean.TRUE == oldValue && Boolean.FALSE == newValue) || (Boolean.FALSE == oldValue && Boolean.TRUE == newValue)) {
```
apeteri: Maybe unboxing to primitive `boolean`s would be better here, just to see if any of the values where `null` for some reason.

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/index/entry/SnomedConceptDocument.java
```diff
@@ -520,6 +576,10 @@ public boolean isStructural() {
 		return structural;
 	}
 	
+	public List<SnomedDescriptionFragment> getDescriptions() {
```
apeteri: 🤔 Is there still time to change the name to `getPreferredDescriptions` or similar, to reflect it better what this field stores?

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/index/change/ConceptChangeProcessor.java
```diff
@@ -171,6 +193,25 @@ public String apply(SnomedRefSet input) {
...
+					Collection<Description> dirtyDescriptions = dirtyDescriptionsByConcept.get(id);
+					if (!dirtyDescriptions.isEmpty()) {
+						Multimap<String, SnomedDescriptionFragment> newDescriptions = HashMultimap.create(Multimaps.index(currentDoc.getDescriptions(), SnomedDescriptionFragment::getId));
+						for (Description dirtyDescription : dirtyDescriptions) {
+							newDescriptions.removeAll(dirtyDescription.getId());
+							if (dirtyDescription.isActive() && !getPreferredLanguageMembers(dirtyDescription).isEmpty()) {
+								newDescriptions.put(dirtyDescription.getId(), toDescriptionFragment(dirtyDescription));
+							}
+						}
+						doc.descriptions(newDescriptions.values().stream().sorted(DESCRIPTION_FRAGMENT_ORDER).collect(Collectors.toList()));
```
apeteri: I will need some help understanding this part. 😕

apeteri (changes requested):

commons/com.b2international.index.api/src/com/b2international/index/Script.java
cmark: It was used by the Lucene index API implementation to load the listed fields for the script's execution. 
In 6.0, we've removed that module, thus this is no longer in use. 
Elasticsearch loads fields/docvalues/source on demand depending on the actual script code and it does not require fields to be explicitly named before the execution, so this fields() argument is unnecessary.

commons/com.b2international.index.api/src/com/b2international/index/query/SortBy.java
cmark: Yes, thanks!

core/com.b2international.snowowl.datastore.server/src/com/b2international/snowowl/datastore/server/EvictingRevisionCache.java
cmark: Fixed it!

core/com.b2international.snowowl.datastore/src/com/b2international/snowowl/datastore/CDOCommitChangeSet.java
cmark: We can save memory that's for sure, especially during import! 😄

core/com.b2international.snowowl.datastore/src/com/b2international/snowowl/datastore/CDOCommitChangeSet.java
cmark: Sounds great! Combining it with the extract local variable refactoring (or to even a field).

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/index/change/ConceptChangeProcessor.java
cmark: You're right! Removed unnecessary type and term ordering.

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/index/entry/SnomedConceptDocument.java
cmark: Renamed it, we will import a new dataset from scratch anyways.

apeteri (approved):

cmark:
Thanks @apeteri!

PR #200, opened 2018-02-21
[reasoner] Sort equivalent concepts by storage keys, and choose the ...

AAAlinaaa:
... one with the smallest storage key to keep.


apeteri (approved):
🌠

PR #204, opened 2018-02-27
[exporter] Keep track of grouped and ungrouped export items in the ...

AAAlinaaa:
... same map, filter for only inferred relationships to export, don't
double count relationships in both grouped and ungrouped sets


apeteri (approved):
LGTM!

PR #211, opened 2018-03-05
Explicitly set value on a CDORemoveFeatureDeltaImpl so it can be read...

AAAlinaaa:
... later by CDOSingleValueFeatureDeltaImpl's writeValue


cmark (approved):
LGTM!

PR #214, opened 2018-03-08
Improve module dependency collector algorithm

cmark:
* Load only the necessary data instead of all concepts.
* Remove ModuleCollectorConfiguration and ModuleCollectorConfigurationThreadLocal.
* Filter out same source-target module dependencies.
* Remove unnecessary module dependency collection based on new modules.
* Modules are concepts and should be handled by the collectConceptModuleDependencies method already.


snomed/com.b2international.snowowl.snomed.datastore.server/src/com/b2international/snowowl/datastore/server/snomed/version/SnomedPublishManager.java
```diff
@@ -80,9 +77,9 @@
 public class SnomedPublishManager extends PublishManager {
 
 	private Set<String> componentIdsToPublish = newHashSet();
-	
-	// moduleId to referencedComponentId
-	private Collection<Pair<String, String>> moduleToReferenceComponentIds;
+
+	// sourceModuleId to targetModuleId map
+	private Map<String, String> moduleDependencies;
```
apeteri: This should be a `Multimap` so that multiple targets can be added for a single source module.

apeteri (dismissed):

nagyo (approved):

zoliMolnar (approved):

cmark:
Thank you! 😄

PR #216, opened 2018-03-19
SO-2919: Prefetch required components when creating reference set members

apeteri:


cmark (approved):
LGTM!

A possible future improvement would be to make the prefetch logic an implicit part of any SNOMED CT specific transactional request (just like we do it with the IdRequest) so that the clients using the API do not need to know there is an alternative way of sending bulk member create/update/etc changes. Also, that would mean we could extend this logic to support descriptions/relationships and any other kind of change request (not just create).

PR #222, opened 2018-04-24
SO-3007: Support simple map reference sets with map target description

apeteri:


cmark:
@nagyo This PR is going to be merged into your [branch](https://github.com/b2ihealthcare/snow-owl/tree/improvement/new_refset_types) (and new PR) that contains all the recently added refset types, right?

PR #225, opened 2018-05-02
SO-2649 Fix preferred descriptions field name in SnomedConceptDocument

zoliMolnar:
Fixed field name of preferred descriptions and initialized it in the builder method.


cmark (approved):
LGTM!

PR #227, opened 2018-05-03
SO-3019: Fix deletion of referring members

zoliMolnar:
This pull request fixes the wrong behaviour of the deletion of referring members.

Removed transient filter from SnomedEditingContext added class Issue3019FixDeletionOfReferringMembersTest to test correct
deletion of referring members


cmark (approved):
LGTM!

PR #236, opened 2018-05-18
SO-3045: Update indexed representation of MRCM

apeteri:
- Add support for source-only object mappings
- Add concept set definitions, predicates and attribute constraints from the MRCM Ecore model as document snippets
- Add domain-level representation for all parts as well
- Support the interpretation of the extended domain models in clients


cmark (approved):
LGTM!

apeteri:
Developer note: when using an existing dataset, run the following command while Snow Owl is running to remove the previous index and mapping:

```
$ curl -XDELETE http://localhost:9200/snomedstore-snomedconstraintdocument
```

...then restart Snow Owl and import the MRCM file again, in "clear" mode.

PR #242, opened 2018-06-21
SO-3119 Fix map type dsv export when exporting referenced component label

mucst:


snomed/com.b2international.snowowl.snomed.exporter.server/src/com/b2international/snowowl/snomed/exporter/server/dsv/MapTypeRefSetDSVExporter.java
```diff
@@ -100,9 +102,11 @@ public File executeDSVExport(final OMMonitor monitor) throws SnowowlServiceExcep
 		final IEventBus bus = applicationContext.getService(IEventBus.class);
 		
 		final SnomedReferenceSet refSet = SnomedRequests.prepareGetReferenceSet(exportSetting.getRefSetId())
-				.setExpand("members(limit:" + Integer.MAX_VALUE + ", expand(referencedComponent(expand(pt()))))")
+				.setExpand("members(limit:" + Integer.MAX_VALUE + ", expand(referencedComponent(expand(preferredDescriptions()))))")
 				.setLocales(languageSetting.getLanguagePreference())
-				.build(SnomedDatastoreActivator.REPOSITORY_UUID, branchPath.getPath()).execute(bus).getSync();
+				.build(SnomedDatastoreActivator.REPOSITORY_UUID, branchPath.getPath())
+				.execute(bus)
+				.getSync();
```
norban: Revert all changes, and instead of using the predefined language preferences by the server, use the correct locales

norban (changes requested):

snomed/com.b2international.snowowl.snomed.exporter.server/src/com/b2international/snowowl/snomed/exporter/server/dsv/MapTypeRefSetDSVExporter.java
mucst: since the setLocales does not effect the setExpand anyway (as discussed with @cmark ), I removed it.

cmark: According to our conversation, I said that the current expansion parameter `preferredDescriptions()` does not use locales because you are expanding all preferred descriptions of the concepts and not a single specific description. I also said that you should use either `pt()` or `fsn()` expansion with the client's locales to reuse logic and not reinvent it (custom description lookup code can be removed). 
@norban suggested this in his review comment.

tamasmucs:
package declarations added, and script return vals conform to the logic of the new validation framework

mucst:
done

norban (approved):

cmark (approved):
LGTM!

PR #251, opened 2018-08-01
Use high-level Elasticsearch REST client for indexing

apeteri:


commons/com.b2international.index.api/src/com/b2international/index/IndexClientFactory.java
```diff
@@ -63,6 +63,11 @@
 	 */
 	String COMMIT_CONCURRENCY_LEVEL = "concurrencyLevel";
 
+	/**
+	 * Configuration key to specify the URL of the Elasticsearch instance to connect to
+	 */
+	String CLIENT_URL = "clientUrl";
```
cmark: I suggest renaming this configuration key and also the key present in the Snow Owl YAML config file to `clusterUrl`.

cmark (approved):
LGTM!

PR #252, opened 2018-08-01
Rest API dependency version bump

zoliMolnar:


core/com.b2international.snowowl.api.rest/src/com/b2international/snowowl/api/rest/SnowowlApiConfig.java
```diff
@@ -48,108 +63,93 @@
...
 @EnableWebMvc
-public class ServicesConfiguration extends WebMvcConfigurerAdapter {
-
-	private SpringSwaggerConfig springSwaggerConfig;
-	private ServletContext servletContext;
+@EnableSwagger2
+@Configuration
+@ComponentScan
+@Import({ SnowowlSecurityConfig.class })
+@PropertySource(value="classpath:com/b2international/snowowl/api/rest/service_configuration.properties")
```
apeteri, line 82: Do the &commat;`Value` annotations work in the code below (and in other classes which use it), are you seeing the expected values from the property file in Swagger's API info section? If not, this snippet is also needed:

```
@Bean
public static PropertySourcesPlaceholderConfigurer ppc() throws IOException {
    return new PropertySourcesPlaceholderConfigurer();
}
```

snomed/com.b2international.snowowl.snomed.api.rest/src/com/b2international/snowowl/snomed/api/rest/config/SnomedSecurityConfig.java
```diff
@@ -15,27 +15,35 @@
...
-
-import com.b2international.snowowl.snomed.api.rest.SnowOwlAuthenticationProvider;
+import org.springframework.security.web.firewall.HttpFirewall;
+import org.springframework.security.web.firewall.StrictHttpFirewall;
 
 /**
  * @since 7.0
  */
 @Configuration
-@EnableGlobalMethodSecurity(securedEnabled=true, proxyTargetClass=true)
```
apeteri, line 35: Note: This annotation corresponds to this bit in the .xml version:  `<global-method-security secured-annotations="enabled" proxy-target-class="true" />`, the corresponding feature is documented here: http://www.baeldung.com/spring-security-method-security

It got removed there, and it got removed here; I don't remember using &commat;`Secured` annotations on any method at the moment, so this _might_  be OK. Otherwise put it back in! 😄

snomed/com.b2international.snowowl.snomed.api.rest/src/com/b2international/snowowl/snomed/api/rest/config/SnomedSecurityConfig.java
```diff
@@ -57,17 +65,7 @@ protected void configure(final HttpSecurity http) throws Exception {
 	
 	@Override
 	protected void configure(final AuthenticationManagerBuilder auth) throws Exception {
-		auth.authenticationProvider(snowOwlAuthenticationProvider());
+		auth.authenticationProvider(authenticationProvider);
 	}
 
-	@Bean
-	@Override
-	public AuthenticationManager authenticationManagerBean() throws Exception {
```
apeteri: This method should be kept! As I understand it, the flow of execution goes something like this:

- This configuration class inherits from `WebSecurityConfigurerAdapter`, which allows it to mess around with the `AuthenticationManagerBuilder` in the method above, before the manager is built;
- We set the provider on the builder to `authenticationProvider`;
- We should also expose the built `AuthenticationManager`, but if this dummy override is not present, it will not be available.

The reason is that the superclass does not have a &commat;`Bean` annotation on this method. See `WebSecurityConfigurerAdapter#authenticationManagerBean()`'s javadoc.

apeteri (changes requested):

core/com.b2international.snowowl.api.rest/src/com/b2international/snowowl/api/rest/SnowowlApiConfig.java
zoliMolnar: The @Value annotations work as intended

snomed/com.b2international.snowowl.snomed.api.rest/src/com/b2international/snowowl/snomed/api/rest/config/SnomedSecurityConfig.java
zoliMolnar: There are no methods which use @secured annotations so I've removed them, should I still add them back in any case?

core/com.b2international.snowowl.api.rest/src/com/b2international/snowowl/api/rest/SnowowlApiConfig.java
apeteri: (A random GitHub user just got mentioned; talking about Java annotations in comments is hard. Use `&commat;` to escape the &commat; symbol.)

apeteri (approved):

zoliMolnar:
Thank you!

core/com.b2international.snowowl.api.rest/src/com/b2international/snowowl/api/rest/SnowowlApiConfig.java
zoliMolnar: I'll keep that in mind for future references, thank you.

PR #255, opened 2018-08-06
Use index queries to retrieve branch data

apeteri:
Use raw queries with field selection and sorting to retrieve:
- The largest segment ID in use
- The largest internal branch ID in use


cmark (approved):
LGTM!

nagyo (approved):
👍

PR #258, opened 2018-08-13
SO-3160 : Introduce dataTypes in ConcreteDomainElements, fix bug ...

cmark:
...resulting in multiple identical concrete domain generation during
classification, ensure single unique id of classification jobs per
branch


apeteri (approved):
🆗

PR #260, opened 2018-08-24
SO-3189: Support effective time filter in validation issues API

norban:


snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/validation/detail/SnomedValidationIssueDetailExtension.java
```diff
@@ -59,6 +61,7 @@
 		
 		public static final String COMPONENT_STATUS = "componentStatus";
 		public static final String COMPONENT_MODULE_ID = "componentModuleId";
+		public static final String COMPONENT_EFFECTIVE_TIME = "effectiveTime";
```
cmark: Use two keys with `START` and `END` suffixes instead of a single key. This will make this API consistent with the other APIs so clients can use the same logic on their side when using this API.

cmark (changes requested):

cmark (approved):
LGTM!

PR #261, opened 2018-08-28
[index] Add searchAfter token support for 6.x

apeteri:


apeteri:
`TODO`: If score-based sorting will be used (see [SortBy](https://github.com/b2ihealthcare/snow-owl/blob/6.x/commons/com.b2international.index.api/src/com/b2international/index/query/SortBy.java#L53)) with a `searchAfter` token, the internal document ID might need to be added as a tie-breaker.

cmark (approved):
LGTM!

PR #262, opened 2018-09-03
Issue/so 3187 validation view filtering

AAAlinaaa:


cmark:
We've decided to index labels on the issue documents to support filtering them by component label instead of fetching the entire universe of SNOMED CT on each `ValidationIssueSearchRequest` with a valid label filter. Closing this PR.

PR #270, opened 2018-09-21
SO-3241: Avoid evaluating large description results when they can be...

AAAlinaaa:
... restricted by snomed constraints on the description carrying
concepts in snomed concept search requests


cmark (changes requested):
How much performance gain we get after applying this change? It would be great to see some numbers. :)

cmark (approved):

cmark:
Performance numbers and load test case will be added in another PR. Merging.

PR #272, opened 2018-10-09
Add support to read environment variables from the snowowl.yml config file

zoliMolnar:
This pull request adds logic to extract the values from environment variables during the parsing of module configs.


commons/com.b2international.commons/src/com/b2international/commons/config/ConfigurationFactory.java
```diff
@@ -184,14 +189,46 @@ private void resolveModuleConfigs(Configuration config) throws IOException {
...
+	private void replaceEnvironmentVariables(JsonNode node) {
+		final Iterator<Entry<String, JsonNode>> fields = node.fields();
+		while (fields.hasNext()) {
+			final Entry<String, JsonNode> field = fields.next();
+			
+			final String fieldName = field.getKey();
+			final String textValue = field.getValue().textValue();
+			if (!Strings.isNullOrEmpty(textValue)) {
+				// match ${ENV_VAR_NAME}
+				final Pattern pattern = Pattern.compile("\\$\\{(\\w+)\\}");
```
cmark: Move to static const field.

commons/com.b2international.commons/src/com/b2international/commons/config/ConfigurationFactory.java
```diff
@@ -184,14 +189,46 @@ private void resolveModuleConfigs(Configuration config) throws IOException {
...
+				final Matcher matcher = pattern.matcher(textValue); 
+				if (matcher.find()) {
```
cmark: Use `java.util.regex.Matcher.matches()` instead.

commons/com.b2international.commons/src/com/b2international/commons/config/ConfigurationFactory.java
```diff
@@ -184,14 +189,46 @@ private void resolveModuleConfigs(Configuration config) throws IOException {
...
+					final String envVariableName = matcher.group(1) == null ? matcher.group(2) : matcher.group(1);
+					final String envVariableValue = System.getenv(envVariableName);
+					if (!Strings.isNullOrEmpty(envVariableValue)) {
```
cmark: Use `System.getenv().containsKey()` to check whether an `ENV_VAR` is specified with the given name and override even if the value is an empty `String`.

cmark (changes requested):

tests/com.b2international.snowowl.test.dependencies/.classpath
```diff
@@ -2,6 +2,8 @@
 <classpath>
 	<classpathentry kind="src" path="src"/>
 	<classpathentry kind="src" path="xtend-gen"/>
+	<classpathentry exported="true" kind="lib" path="system-rules-1.18.0-sources.jar"/>
```
cmark: This should be added as `sourcepath` to the existing `system-rules` classpath entry.

commons/com.b2international.commons/src/com/b2international/commons/config/ConfigurationFactory.java
```diff
@@ -184,14 +193,44 @@ private void resolveModuleConfigs(Configuration config) throws IOException {
...
+			
+			final String fieldName = field.getKey();
+			final String textValue = field.getValue().textValue();
+			if (!Strings.isNullOrEmpty(textValue)) {
+				final Matcher matcher = ENV_VAR_PATTERN.matcher(textValue); 
+				if (matcher.matches()) {
+					final String envVariableName = matcher.group(1);
+					if (System.getenv().containsKey(envVariableName)) {
+						final String envVariableValue = System.getenv(envVariableName);
+						((ObjectNode) node).put(fieldName, envVariableValue);
```
cmark, line 223: ⚠️ Potential type error!

Please write test cases for the following types to verify that this will indeed work correctly:
* Boolean
* Short
* Integer
* Float
* Long
* Float
* Double
* String

cmark (changes requested):

cmark (approved):
LGTM!

PR #274, opened 2018-10-13
SO-3262: Fix various refset import related issues

AAAlinaaa:
1. Ensure changes are correctly reported at the end of import
2. Remove trailing white spaces from the last column in csv import to
ensure ids are read correctly
3. Fix index out of bounds exception in SnomedCompositeImporter
4. Ensure Synonym service is available in SnomedConceptCreateRequest
5. Make sure new reference set appears in the refset view by creating
both inferred and stated is a relationships to appropriate refset parent
concept on import

### DIFF
--- a/snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/request/SnomedConceptCreateRequest.java
+++ b/snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/request/SnomedConceptCreateRequest.java
@@ -140,9 +140,10 @@ public final class SnomedConceptCreateRequest extends BaseSnomedComponentCreateR
 	}
 
 	private void convertDescriptions(TransactionContext context, final String conceptId) {
+		TransactionContext newContext = context.inject().bind(Synonyms.class, new Synonyms(context)).build();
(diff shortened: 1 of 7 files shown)


snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/request/SnomedConceptCreateRequest.java
```diff
@@ -140,9 +140,10 @@ private Concept convertConcept(final TransactionContext context) {
 	}
 
 	private void convertDescriptions(TransactionContext context, final String conceptId) {
+		TransactionContext newContext = context.inject().bind(Synonyms.class, new Synonyms(context)).build();
```
cmark: Domain-specific requests should not alter the incoming context, they should only use the services provided by them. `SnomedConceptCreateRequest` instances are usually being executed inside a `com.b2international.snowowl.snomed.datastore.request.SnomedBulkRequest<R>`, which already binds the `Synonyms` service to the context.
Executing `SnomedConceptCreateRequest` instances outside of `SnomedBulkRequest` is allowed, but in those cases, the context should be able to provide the `Synonyms` cache/service, meaning that the creator of the context should inject the necessary services before passing them to the requests.

cmark (changes requested):

cmark (approved):
LGTM!

PR #281, opened 2018-11-13
Elasticsearch TCP client support

cmark:
This PR changes the index layer to support TCP based connections to remote Elasticsearch clusters.
* `clusterUrl` configuration property now accepts `tcp://` scheme
* Bump Elasticsearch version to 6.5.0
* Use local Java client to communicate with Elasticsearch in embedded mode
* Add `-Dso.index.es.useHttp` system property to force HTTP connection in embedded Elasticsearch scenarios


commons/com.b2international.index/pom.xml
```diff
@@ -8,4 +8,129 @@
 	</parent>
 	<artifactId>com.b2international.index</artifactId>
 	<packaging>eclipse-plugin</packaging>
+	
+	<properties>
+		<elasticsearch.version>6.4.2</elasticsearch.version>
+		<log4j2.version>2.11.1</log4j2.version>
+    </properties>
+    
+    <!-- NOTE: in case of dependency changes, please update the includeArtifactIds as well in the maven-dependency-plugin configuration as well, listing all (exact and transitive) dependency artifactIds -->
```
apeteri:
```suggestion
    <!-- NOTE: in case of dependency changes, please update "includeArtifactIds" in maven-dependency-plugin's configuration as well, listing all (exact and transitive) dependency artifacts -->
```

commons/com.b2international.index/src/com/b2international/index/Indexes.java
```diff
@@ -60,7 +60,7 @@ public static Index createIndex(String name, ObjectMapper mapper, Mappings mappi
 	}
 	
 	private static ObjectMapper configure(ObjectMapper mapper) {
-		return mapper.copy()
```
apeteri: `mapper` is usually set to `env.service(ObjectMapper.class)`, so without copying it, the statements below will change the configuration of our shared `ObjectMapper` instance.

commons/com.b2international.index/src/com/b2international/index/es/EsDocumentSearcher.java
```diff
@@ -488,8 +488,10 @@ private void addSort(DocumentMapping mapping, SearchSourceBuilder reqSource, Sor
 			} else {
 				topHitsAgg
 					.storedFields(STORED_FIELDS_NONE)
-					.fetchSource(false)
-					.fieldDataFields(aggregation.getFields());
+					.fetchSource(false);
+				
+				aggregation.getFields().forEach(field -> topHitsAgg.docValueField(field, "use_field_mapping"));
```
apeteri, line 493: 👍

commons/com.b2international.index/src/com/b2international/index/es/EsClientConfiguration.java
```diff
@@ -71,6 +95,22 @@ public boolean equals(final Object obj) {
 		final EsClientConfiguration other = (EsClientConfiguration) obj;
 
 		// First client configuration for a host wins
-		return Objects.equals(host, other.host);
+		return Objects.equals(clusterUrl, other.clusterUrl);
```
apeteri: Cluster name should be also included in `equals(Object)` and `hashCode()` checks. The client pool is using this configuration object as a map key.

commons/com.b2international.index/src/com/b2international/index/es/client/EsClient.java
```diff
@@ -0,0 +1,148 @@
...
+			LOG.info("Connecting to Elasticsearch cluster at '{}'{}, connect timeout: {} ms, socket timeout: {} ms.", 
+					configuration.getClusterUrl(),
+					configuration.isProtected() ? " using basic authentication" : "",
+					configuration.getConnectTimeout(),
+					configuration.getSocketTimeout());
+			
+			if (configuration.isHttp()) {
+				return new EsHttpClient(configuration);
+			} else {
+				checkState(configuration.isTcp(), "Only TCP and HTTP clients are supported");
```
apeteri, line 121: We could also throw an exception if the transport client is configured with a user name and password, as that is only supported in the x-pack version.

commons/com.b2international.index/src/com/b2international/index/es/client/http/EsHttpClient.java
```diff
@@ -0,0 +1,179 @@
...
+	
+	@Override
+	public final ClearScrollResponse clearScroll(ClearScrollRequest req) throws IOException {
+		return clientExt.clearScroll(req, RequestOptions.DEFAULT);
+	}
+	
+	@Override
+	public BulkByScrollResponse updateByQuery(String index, String type, int batchSize, Script script, int numberOfSlices, QueryBuilder query)
+			throws IOException {
+		UpdateByQueryRequest updateByQueryRequest = new UpdateByQueryRequest(new SearchRequest());
```
apeteri: The zero-arg constructor also starts out with an unpopulated `SearchRequest`: https://github.com/elastic/elasticsearch/blob/bc5b1afb55c3b9a75c8d3f69c7832765918b439c/server/src/main/java/org/elasticsearch/index/reindex/UpdateByQueryRequest.java#L47-L49

apeteri (changes requested):

commons/com.b2international.index/src/com/b2international/index/es/client/EsClient.java
apeteri: See [Java Client and Security](https://www.elastic.co/guide/en/elastic-stack-overview/6.5/java-clients.html) in the Elastic Stack documentation.

commons/com.b2international.index/src/com/b2international/index/es/client/http/EsHttpClient.java
cmark: Bumped to ES 6.5.0 in the meantime, which already got rid of the mentioned constructor.

commons/com.b2international.index/src/com/b2international/index/Indexes.java
cmark: This change is only required on the `7.x` branch. Reverting.

apeteri (approved):

nagyo (approved):

PR #282, opened 2018-11-14
Support multiple code systems versioning

norban:
Ticket reference: https://snowowl.atlassian.net/browse/SDD-901

### DIFF
--- a/core/com.b2international.snowowl.datastore/src/com/b2international/snowowl/datastore/request/version/CodeSystemVersionCreateRequest.java
+++ b/core/com.b2international.snowowl.datastore/src/com/b2international/snowowl/datastore/request/version/CodeSystemVersionCreateRequest.java
@@ -98,10 +98,7 @@ final class CodeSystemVersionCreateRequest implements Request<ServiceProvider, B
 		Set<String> repositoryIds = context.service(RepositoryManager.class).repositories().stream().map(Repository::id).collect(Collectors.toSet());
 		for (String repositoryId : repositoryIds) {
 			try {
-				codeSystem = CodeSystemRequests.prepareGetCodeSystem(codeSystemShortName)
-						.build(repositoryId)
-						.execute(context.service(IEventBus.class))
-						.getSync();
+				codeSystem = fetchCodeSystem(context, repositoryId, codeSystemShortName);
 			} catch (NotFoundException e) {
 				//ignore
 			}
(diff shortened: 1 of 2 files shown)


core/com.b2international.snowowl.datastore/src/com/b2international/snowowl/datastore/request/version/CodeSystemVersionCreateRequest.java
```diff
@@ -124,18 +124,58 @@ public Boolean execute(ServiceProvider context) {
...
+		return Boolean.TRUE;
+	}
+	
+	private Map<CodeSystemEntry, IVersioningManager> createVersioningManagers(final ServiceProvider context, final CodeSystemEntry codeSystem) {
+		final Map<CodeSystemEntry, IVersioningManager> versioningManagersByCodeSystem  = Maps.newHashMap();
+		
+		final IVersioningManager versioningManager = VersioningManagerBroker.INSTANCE.createVersioningManager(codeSystem.getTerminologyComponentId());
+		versioningManagersByCodeSystem.put(codeSystem, versioningManager);
+
+		if (versioningManager.needsOtherCodeSystemsToVersion()) {
```
cmark, line 142: Not sure if this flag is really required, I would only use the `codeSystemDependenciesByRepositoryId()` method with a `forEach` if additional code systems need to be versioned along with the current one.

core/com.b2international.snowowl.datastore/src/com/b2international/snowowl/datastore/request/version/CodeSystemVersionCreateRequest.java
```diff
@@ -124,18 +124,58 @@ public Boolean execute(ServiceProvider context) {
...
+			final Map<String, String> codeSystemsByRepositoryId = versioningManager.codeSystemDependenciesByRepositoryId();
+			for (Entry<String, String> entry : codeSystemsByRepositoryId.entrySet()) {
```
cmark, line 144: Also this entire logic could go to the `versionManager`, we simply ask the "root" `versionManager` to gather and return all dependent version managers along with itself. Other code that depends on this logic can reuse that functionality via the well-named method.

core/com.b2international.snowowl.datastore/src/com/b2international/snowowl/datastore/request/version/CodeSystemVersionCreateRequest.java
```diff
@@ -124,18 +124,58 @@ public Boolean execute(ServiceProvider context) {
...
+	
+	
+	private CodeSystemEntry fetchCodeSystem(ServiceProvider context, final String repositoryId, final String codeSystemShortName) {
+		return CodeSystemRequests.prepareGetCodeSystem(codeSystemShortName)
+			.build(repositoryId)
+			.execute(context.service(IEventBus.class))
+			.getSync();
+	}
+
+	private void createVersion(ServiceProvider context, IVersioningManager versioningManager, CodeSystemEntry codeSystem, String user) {
```
cmark, line 169: Reusing this method in a `forEach` to version all dependent code systems as well long with the current one certainly does its job, but we can improve it to limit the number of unwanted content changes while the versioning runs for a single code system.

Refactor the versioning process to `acquire` and `release` locks for all target code systems at once at the beginning and at the end of the process.

cmark (changes requested):

PR #289, opened 2018-12-11
Fix script arguments unrecognized by Elasticsearch

cmark:
Convert Set and ImmutableSet to List instance when connecting to remote Elasticsearch via TCP.


apeteri (approved):
👍

cmark:
Thanks @apeteri!

PR #290, opened 2018-12-11
Allow non-SNOMED CT mapCategoryId value

cmark:
This is to support UK map block type refset by reusing the existing
extended map type.


apeteri (approved):
👍

PR #291, opened 2018-12-12
Consider CD members in NormalFormGenerator if a group only contains relationships with a non-zero union group number

cmark:


nagyo (approved):
LGTM!

PR #303, opened 2019-02-14
No lock timeout in RF2 import

cmark:
This will throw LockException to notify clients about the repository locked state.


apeteri (approved):
Does removing the `catch` clause result in the exception propagating to the handler in the other change? LGTM if so.

cmark:
> Does removing the `catch` clause result in the exception propagating to the handler in the other change? LGTM if so.

Yes, it will propagate to the handler. Merging.

PR #310, opened 2019-03-04
[api] fix snowowl.api.rest MANIFEST.MF

cmark:
Replace Require-Bundle entries with Import-Package to fix startup issue
in Virgo packages.


apeteri (approved):
🍎

PR #317, opened 2019-03-08
[reasoner] Ignore property chain hierarchies collection if no type IDs..

cmark:
...make use of this functionality, to speed up normal form computation


apeteri (approved):
💯

PR #318, opened 2019-03-11
SDD-905: Classification changes related to equivalent concept merging

apeteri:
- Make equivalent concept merging implementations extensible
- Skip property chain hierarchy collection when no property chains/transitive properties are defined
- Make `Promise` a subclass of `Observable`


cmark (approved):
LGTM!

PR #322, opened 2019-03-20
Release v6.13.1 bugfixes

cmark:


apeteri (approved):
These are two small fixes to the history view that I've found, self-👍

PR #346, opened 2019-05-06
OWLAP-49: Fix calculation of definition status using OWL axioms

nagyo:
https://jira.ihtsdotools.org/browse/OWLAP-49


cmark (approved):
LGTM!

PR #349, opened 2019-05-07
SO-3595: Wrap sync BulkRequests with IdRequest upon execution

nagyo:
https://snowowl.atlassian.net/browse/SO-3595


cmark (approved):
LGTM!

PR #362, opened 2019-06-17
Disable stated MRCM rule

AAAlinaaa:


snomed/com.b2international.snowowl.validation.snomed/src/main/resources/validation-rules-common.json
```diff
@@ -43,23 +43,5 @@
     "checkType": "EXPENSIVE",
     "type": "script-groovy",
     "implementation": "scripts/rule_mrcm_constraint_type.groovy"
-  },
-  {
-  	"id": "rule_stated_mrcm_constraint",
```
cmark, line 48: Is it possible to just comment out these rules and mark corresponding test cases with `@Ignore` annotation? If yes, please do that instead of blindly removing them completely.

cmark (changes requested):

snomed/com.b2international.snowowl.validation.snomed/src/main/resources/validation-rules-common.json
cmark: Let's just remove it for now.

cmark (approved):

PR #366, opened 2019-06-24
SO-3639: Remove logic that disallowed reactivating a concept while adding a pending move indicator, or association targets

zoliMolnar:


cmark (approved):
LGTM!

PR #379, opened 2019-07-23
SO-3661 Component Identifier Service API

cmark:
This PR adds partial support for [Component Identifier Service API](https://github.com/IHTSDO/component-identifier-service). Snow Owl with EMBEDDED ID service can be used as a CIS in a cluster of Snow Owl instances. Also, the PR adds a separate CIS only product based the Snow Owl platform.

* New modules
  * com.b2international.snowowl.snomed.cis (core CIS and SNOMED CT ID services)
  * com.b2international.snowowl.snomed.cis.rest (Support CIS API)
  * com.b2international.snowowl.snomed.cis.product (CIS Product)
  * com.b2international.snowowl.snomed.cis.feature (CIS Feature)


apeteri:
lgtm (and you, as you have mentioned it earlier) has a small-ish point: the loop shown below could be reworked into eg. a do-while loop, so that the operation runs at least once, regardless of the value set in `numberOfReauthTries`.

https://github.com/b2ihealthcare/snow-owl/blob/97028c76da26bf882fae39a6d25e67f9014232ab/cis/com.b2international.snowowl.snomed.cis/src/com/b2international/snowowl/snomed/cis/client/CisSnomedIdentifierService.java#L531-L561

cis/com.b2international.snowowl.snomed.cis.product/assembly/common/bin/cis.sh
```diff
@@ -0,0 +1,77 @@
...
+                -Djetty.home.bundle=org.eclipse.jetty.osgi.boot \
+                -Dorg.eclipse.jetty.util.log.class=org.eclipse.jetty.util.log.Slf4jLog \
+                -XX:+UseConcMarkSweepGC \
+                -XX:CMSInitiatingOccupancyFraction=75 \
+                -XX:+UseCMSInitiatingOccupancyOnly \
+                -XX:+HeapDumpOnOutOfMemoryError \
+                -Djdk.security.defaultKeySize=DSA:1024 \
+                $SO_JAVA_OPTS"
+
+cd "$KERNEL_HOME";
```
apeteri: `pushd` and `popd` should also work under Bash, if you want to return to the working directory the script was originally started from.

cis/com.b2international.snowowl.snomed.cis.product/assembly/common/configuration/jetty-http.xml
```diff
@@ -0,0 +1,41 @@
...
+          <Array type="org.eclipse.jetty.server.ConnectionFactory">
+            <Item>
+              <New class="org.eclipse.jetty.server.HttpConnectionFactory">
+                <Arg name="config"><Ref refid="httpConfig" /></Arg>
+              </New>
+            </Item>
+          </Array>
+        </Arg>
+        <Set name="host"><Property name="jetty.http.host" /></Set>
+        <Set name="port"><Property name="jetty.http.port" default="80" /></Set>
```
apeteri: It might be more convenient to set a non-default port number here right away (although it is overridden in the startup script as well).

apeteri (changes requested):

apeteri (approved):
👍

PR #383, opened 2019-08-02
[queryrefset] Ommit displaying inactive query refset members in the ...

AAAlinaaa:
... Update Query Refset Membre wizard


cmark (approved):
LGTM!

PR #400, opened 2019-09-20
SO-3763: Description character length validation

zoliMolnar:


snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/index/entry/SnomedDescriptionIndexEntry.java
```diff
@@ -68,6 +68,7 @@
 @Doc
 @JsonDeserialize(builder = SnomedDescriptionIndexEntry.Builder.class)
 @Script(name="normalizeWithOffset", script="(_score / (_score + 1.0f)) + params.offset")
+@Script(name="descriptionLength", script="doc['term.original'].value.length() > params.length")
```
cmark, line 71: In this case, I would not use a named script inject the whole script into the rule instead. Add the necessary support to the Expression language if required.

cmark (changes requested):

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/index/entry/SnomedDescriptionIndexEntry.java
apeteri: References to fields in the script are easier to update if they are all stored in the document class, though.

apeteri: It turns out that scripts can already be executed inline, it is not a requirement to add it as an annotation as we suspected. 👍 for moving it to the validation rule.

PR #406, opened 2019-09-30
SO-3769: Job auto clean up API support

zoliMolnar:


core/com.b2international.snowowl.datastore/src/com/b2international/snowowl/datastore/remotejobs/RemoteJob.java
```diff
@@ -44,21 +45,25 @@
 	private final String id;
 	private final ServiceProvider context;
 	private final Request<ServiceProvider, ?> request;
+	private final boolean autoClean;
 	
 	private String response;
 	private String user;
+	private IStatus status;
```
cmark: This should be either `transient` field or not a field at all. I suggest moving this as a variable into the `run(IProgressMonitor monitor)` method.

core/com.b2international.snowowl.datastore/src/com/b2international/snowowl/datastore/remotejobs/RemoteJob.java
```diff
@@ -44,21 +45,25 @@
 	private final String id;
 	private final ServiceProvider context;
 	private final Request<ServiceProvider, ?> request;
+	private final boolean autoClean;
```
cmark: Another option would be to remove the `final` modifier from the `autoClean` field and set it to `false` in case of an error, which would simplify the implementation of the `run` method significantly.

cmark (changes requested):

cmark (approved):
LGTM!

PR #420, opened 2019-11-07
SO-3575: Duplicate reference set member validation rule

zoliMolnar:


snomed/com.b2international.snowowl.validation.snomed/src/main/resources/scripts/rule_duplicate_members.groovy
```diff
@@ -0,0 +1,65 @@
...
+
+final RevisionSearcher searcher = ctx.service(RevisionSearcher.class)
+
+final List<ComponentIdentifier> issues = Lists.newArrayList()
+
+final String script = String.format("return %s + '_' + %s",
+					"doc.referenceSetId.value",
+					"doc.referencedComponentId.value")
+
+final ExpressionBuilder queryBuilder = Expressions.builder().filter(SnomedRefSetMemberIndexEntry.Expressions.refSetTypes(Collections.singleton(SnomedRefSetType.SIMPLE)))
```
apeteri: The candidates should be restricted to active members only, we don't leave any way of fixing the issue otherwise (except mass whitelisting) for released/historical components.

apeteri (changes requested):

apeteri (approved):

PR #429, opened 2019-11-14
Cancel classification docs for cancelled classification jobs

zoliMolnar:


cmark (approved):
I would consider using the Notification stream API instead of introducing another dependency to the Eclipse Job API. Apart from this the PR looks good.

PR #443, opened 2019-11-28
Release/7.2.0

nagyo:


core/com.b2international.snowowl.identity/src/com/b2international/snowowl/identity/JWTGenerator.java
```diff
@@ -87,8 +93,17 @@ public String generate(User user) {
...
-		final List<Permission> permissions = jwt.getClaim(PERMISSION).asList(String.class).stream().map(Permission::valueOf).collect(Collectors.toList());
-		return new User(subject, ImmutableList.of(new Role("oauth_scopes", permissions)));
+		final List<Role> roles = jwt.getClaims().entrySet().stream()
+				.filter(entry -> entry.getKey().startsWith(ROLE_PREFIX))
+				.map(entry -> {
+					final String roleName = entry.getKey();
+					final Claim claim = entry.getValue();
+					final List<Permission> permissions = claim.asList(String.class).stream().map(Permission::valueOf).collect(Collectors.toList());
+					
+					return new Role(roleName.substring(roleName.indexOf("_") + 1, roleName.length()), permissions);
```
cmark: You could just use `String.replace()` to replace the `ROLE_PREFIX` with an empty String and use the resulting value as the role name,

cmark (changes requested):

PR #449, opened 2019-12-03
Merge commit handling improvements

cmark:
This PR adds support for non-squash merge commits and improves the implementation of the branch merge logic.

### DIFF
--- a/commons/com.b2international.index.tests/src/com/b2international/index/revision/RevisionFixtures.java
+++ b/commons/com.b2international.index.tests/src/com/b2international/index/revision/RevisionFixtures.java
@@ -52,7 +52,7 @@ public class RevisionFixtures {
 			}
 
 			public Builder id(String id) {
-				this.id = field1;
+				this.id = id;
 				return getSelf();
 			}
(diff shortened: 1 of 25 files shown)


commons/com.b2international.index.tests/src/com/b2international/index/revision/RevisionFixtures.java
```diff
@@ -41,18 +41,18 @@ private RevisionFixtures() {
...
+			
 			public Builder(RevisionData revisionData) {
 				this.id = revisionData.getId();
 				this.field1 = revisionData.field1;
 				this.field2 = revisionData.field2;
 			}
 
 			public Builder id(String id) {
-				this.id = id;
+				this.id = field1;
```
apeteri: This fix should be kept; it was incorrect before the PR and fixed as part of a commit that has been reverted.

```suggestion
				this.id = id;
```

apeteri (approved):
LGTM. The only requested change is in test code, it can be fixed after merging the PR as well.

PR #460, opened 2019-12-13
SO-3845: Handle ECL expressions in attributes' concept set ...

AAAlinaaa:
... definitions in MRCM validation rules


cmark (approved):
LGTM!

PR #467, opened 2020-01-17
SO-3892 fix incorrectly displayed query refset changes

adeakgabi:
JIRA ticket: https://snowowl.atlassian.net/browse/SO-3892


cmark (approved):
LGTM!

PR #491, opened 2020-02-20
SO-3882: Template model and API

nagyo:


core/com.b2international.snowowl.datastore/src/com/b2international/snowowl/core/repository/CompositeComponentDeletionPolicy.java
```diff
@@ -0,0 +1,68 @@
...
+import java.util.function.Predicate;
+
+import com.b2international.snowowl.datastore.index.RevisionDocument;
+
+/**
+ * @since 7.4.0
+ */
+public class CompositeComponentDeletionPolicy implements ComponentDeletionPolicy {
+
+	Map<Class<?>, Predicate<RevisionDocument>> deletionPolicies;
```
apeteri: Probably doesn't make too much of a difference, but `Predicate<RevisionDocument>` occurrences could be replaced with `ComponentDeletionPolicy` to really drive the point home of this class being a _composite_ of other policies.

core/com.b2international.snowowl.datastore/src/com/b2international/snowowl/core/repository/CompositeComponentDeletionPolicy.java
```diff
@@ -0,0 +1,68 @@
...
+	}
+
+	CompositeComponentDeletionPolicy(final Map<Class<?>, Predicate<RevisionDocument>> deletionPolicies) {
+		this.deletionPolicies = deletionPolicies;
+	}
+
+	@Override
+	public boolean canDelete(final RevisionDocument revision) {
+		return deletionPolicies.entrySet().stream()
+				.filter(entry -> entry.getKey().isInstance(revision))
```
apeteri, line 60: If there is no "is the instance class same or subclass of this class"-kind of testing needed (where you are registering multiple applicable policies, eg. for `SnomedConceptDocument` → `SnomedComponentDocument` → `SnomedDocument`), lookup instead of iteration would be quicker:

```java
if (revision == null) { return false; }
return deletionPolicies.get(revision.getClass()).test(revision);
```

core/com.b2international.snowowl.datastore/src/com/b2international/snowowl/core/repository/RepositoryBuilder.java
```diff
@@ -145,6 +145,20 @@ public Repository build(Environment env) {
 				configurer.getAdditionalMappings().forEach(mappings::putMapping);
 			});
 		
+		if (deletionPolicy instanceof CompositeComponentDeletionPolicy) {
+			
+			CompositeComponentDeletionPolicy compositePolicy = (CompositeComponentDeletionPolicy) deletionPolicy;
+			
+			repositoryConfigurers.forEach(configurer -> {
+				configurer.getComponentDeletionPolicies().forEach( (clazz, predicate) -> {
+					if (!compositePolicy.getDeletionPolicies().containsKey(clazz)) {
```
apeteri: `compositePolicy.getDeletionPolicies().putIfAbsent(clazz, predicate);`

It's also a bit unexpected (for me) to see a builder on `CompositeComponentDeletionPolicy`, suggesting that the result is immutable, then proceed to pull a writable `Map` out via the get method and inject extra entries. A factory method could be added to the builder that starts with an existing `CompositeComponentDeletionPolicy`.

core/com.b2international.snowowl.datastore/src/com/b2international/snowowl/core/repository/TerminologyRepositoryConfigurer.java
```diff
@@ -32,4 +36,9 @@
 	 */
 	Collection<Class<?>> getAdditionalMappings();
 
+	/**
+	 * @return additional component deletion policies for subtypes of RevisionDocument
+	 */
+	Map<Class<?>, Predicate<RevisionDocument>> getComponentDeletionPolicies();
```
apeteri: This can also be hidden entirely behind the `ComponentDeletionPolicy` interface:

```suggestion
	ComponentDeletionPolicy getComponentDeletionPolicy();
```

apeteri (changes requested):

core/com.b2international.snowowl.datastore/src/com/b2international/snowowl/core/repository/CompositeComponentDeletionPolicy.java
nagyo: There is a need for those checks though, the main SNOMED deletion policy is applicable for all subtypes of SnomedDocument. I will add a fast evaluation part to the beginning of that method, that could make processing a bit faster.

core/com.b2international.snowowl.datastore/src/com/b2international/snowowl/core/repository/TerminologyRepositoryConfigurer.java
nagyo: But we need that class information to be able to select the appropriate policy for a given type, don't we?

apeteri: My mistake; wanted to say `CompositeComponentDeletionPolicy`. This ties into my previous comment, this could be done if there would be a way to merge multiple composites into one, eg. via a `CompositeComponentDeletionPolicy#builder(CompositeComponentDeletionPolicy)` or `CompositeComponentDeletionPolicy.Builder#withPolicies(CompositeComponentDeletionPolicy)` method.

apeteri (approved):

PR #511, opened 2020-03-06
Feature: Generic Concept Request API

cmark:
This PR introduces a new Concept request API to access any concept from any CodeSystem using a few filters and returning a generic Concept model.


commons/com.b2international.commons/src/com/b2international/commons/options/Options.java
```diff
@@ -67,9 +67,24 @@
 	 *             if the specified key is <code>null</code>
 	 */
 	Object get(String key);
+	
+	/**
+	 * Returns the value to which the specified Enum key's name is mapped, or {@code null} if this map contains no mapping for the key.
+	 *
+	 * <p>
+	 * More formally, if this map contains a mapping from a key {@code k} to a value {@code v} such that {@code (key==null ? k==null :
+	 * key.equals(k))}, then this method returns {@code v}; otherwise it returns {@code null}. (There can be at most one such mapping.)
```
apeteri:
```suggestion
	 * key.name().equals(k))}, then this method returns {@code v}; otherwise it returns {@code null}. (There can be at most one such mapping.)
```

commons/com.b2international.commons/src/com/b2international/commons/options/OptionsBuilder.java
```diff
@@ -30,10 +30,20 @@
 	OptionsBuilder() {}
 	
 	public OptionsBuilder put(String key, Object value) {
-		options.put(key, value);
+		if (value != null) {
```
apeteri, line 33: I don't know if we ever call `put` twice for the same key, but `null` effectively removed the previous value in the method's "before" form, and now it doesn't.

core/com.b2international.snowowl.datastore/src/com/b2international/snowowl/core/request/ConceptSearchRequestEvaluator.java
```diff
@@ -0,0 +1,82 @@
...
+		 */
+		TERM_EXACT,
+
+		/**
+		 * A query expression defined in the target code system's query language
+		 */
+		QUERY,
+
+		/**
+		 *
```
apeteri: This option key is not explained.

apeteri (changes requested):

apeteri (approved):
LGTM!

PR #515, opened 2020-03-19
Template authoring support

apeteri:


commons/com.b2international.index/src/com/b2international/index/revision/StagingArea.java
```diff
@@ -789,7 +789,14 @@ public ArrayNode diff() {
 				while (elements.hasNext()) {
 					JsonNode node = elements.next();
 					final ObjectNode change = ClassUtils.checkAndCast(node, ObjectNode.class);
-					final String property = change.get("path").asText().substring(1);
+					
+					// Remove trailing segments in nested property paths (we are only interested in the top property)
+					String property = change.get("path").asText().substring(1);
```
cmark, line 794: I think you need to perform this String manipulation on line 767 as well. If yes, please extract it to a method so we can further improve changed property detection later on.

cmark (approved):
LGTM! Small change in the property detection, if that's done feel free to merge the PR @apeteri or @nagyo !

commons/com.b2international.index/src/com/b2international/index/revision/StagingArea.java
apeteri: That is an unfiltered differencer which should only ignore `created` and `revised` columns.

PR #539, opened 2020-04-22
Add issue templates

cmark:


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,27 @@
...
+---
+
+**Describe the bug**
+A clear and concise description of what the bug is.
+
+**To Reproduce**
+
+Useful things to include in the report:
+* Example Request/Response pairs
+* Example `snowowl.yml` configuration
```
apeteri, line 20: I don't have exact words for it, but we could mention two things:

* The submitter should make sure not to expose any credentials accidentally when including files or snippets
* The reproduction steps, requests, and responses should form a _minimal_ reproducible example, if possible - it is harder to try and spot an error in megabytes of logs than looking at a smallish configuration object.

apeteri (approved):

.github/ISSUE_TEMPLATE/bug_report.md
cmark: Yeah good ideas, will include them soon. 👍

apeteri (approved):
:+1:

PR #542, opened 2020-04-23
INFRA-42: Shrink size of Snow Owl OSS docker images

nagyo:
- use prebuilt CentOS 7 image as a base. The base image comes with daily package updates and AdoptOpenJDK11 pre-installed
- use multistage build to get rid of install layers
- use simple tar.gz artifact to extract and install Snow Owl
- set explicit user and group ID
- removed enforced ownership mutation in case of bind-mounts. It is now the responsibility of the user to setup necessary permissions for log, data and configuration dirs.


cmark (approved):
LGTM!

apeteri (approved):

PR #545, opened 2020-04-26
INFRA-30: Extract docker image build to separate job

nagyo:


docker/Dockerfile
```diff
@@ -40,22 +42,22 @@ EXPOSE 2036 8080
 
 LABEL org.label-schema.build-date="${BUILD_TIMESTAMP}" \
   org.label-schema.vcs-ref="${GIT_REVISION}" \
-  org.label-schema.version="${VERSION}" \
+  org.label-schema.version="${TAG}" \
   org.label-schema.license="Apache-2.0" \
-  org.label-schema.name="Snow Owl OSS" \
+  org.label-schema.name="${PRODUCT_NAME}" \
   org.label-schema.schema-version="1.0" \
-  org.label-schema.url="https://github.com/b2ihealthcare/snow-owl" \
```
cmark, line 47: Set the old default values to the arguments, so anyone who comes to this repo can build the image with just `docker build .`.
I know Nexus is something that's still required, we will remove that dependency in the next iteration of the docker build process.

cmark (changes requested):

docker/Dockerfile
nagyo: Default values are set for arguments so the image can be built as before

Jenkinsfile
```diff
@@ -39,52 +31,30 @@ node('docker') {
...
+	if (currentBuild.resultIsBetterOrEqualTo('SUCCESS')) {
+		
+		build job: 'snow-owl-docker-build', parameters: [
+			string(name: 'groupId', value: 'com.b2international.snowowl'),
+			string(name: 'artifactId', value: 'com.b2international.snowowl.server.update'),
+			string(name: 'artifactVersion', value: currentVersion),
+			string(name: 'classifier', value: 'oss'),
+			string(name: 'extension', value: 'tar.gz'),
+			string(name: 'imageClassifier', value: 'oss'),
+			string(name: 'gitRevision', value: revision),
```
apeteri, line 48: Is there a default value given for the registry in the docker build job?

apeteri (approved):
LGTM, with a single question.

Jenkinsfile
nagyo: yes, empty registry means hub.docker.com

cmark (approved):

docker/Dockerfile
cmark: Yes, but the default values are in the `Jenkinsfile`, right?

cmark: I see them now, okay.

PR #546, opened 2020-04-27
SO-3998: Code system API changes

apeteri:
* Make `branchPath` parameter optional in both Java and REST APIs and use the specified `shortName` as a source to create a child branch under `MAIN`. Automatically create and assign the child branch to the newly created CodeSystem.

* Use `CodeSystemURI` values in `extensionOf` property; require the property to be always set to an exact version dependency and not to the latest version of the Code System. For example, International Edition dependencies would require to specify the `SNOMEDCT` shortName along with a version path, like `SNOMEDCT/2020-01-31`.

* Add the `availableUpgrades` property to the Code System API. This shows the available `extensionOf` Code System URI values that can be used to upgrade the Code System.

* Make `language` property deprecated and add a `locales` `ExtendedLocale` type based property instead (`List<ExtendedLocale>`). The API should allow clients to create the Code System without any locales and update the locales later on, if required. Additionally, in SNOMED CT Tooling it would be awesome to upgrade the closest Code System with all imported languages reference sets with the locale value of `en-x-<languageRefSetId>`.

* To support a single namespace and a list of module values on each SNOMED CT specific Code System, we can add them directly to the model (other Code Systems will ignore the property completely, therefore they are optional) or add a generic `additionalProperties: Map<String, Object>` to attach arbitrary number of configuration key value pairs to the Code System model.


core/com.b2international.snowowl.core.rest/src/com/b2international/snowowl/core/rest/codesystem/CodeSystemRestService.java
```diff
@@ -58,14 +70,88 @@
...
 	@ApiResponses({
-		@ApiResponse(code = 200, message = "OK")
+		@ApiResponse(code = 200, message = "OK", response = CodeSystems.class),
+		@ApiResponse(code = 400, message = "Invalid search config", response = RestApiError.class),
+		@ApiResponse(code = 404, message = "Branch not found", response = RestApiError.class)
 	})
 	@GetMapping(produces = { AbstractRestService.JSON_MEDIA_TYPE })
-	public CollectionResource<CodeSystem> getCodeSystems() {
-		return CollectionResource.of(codeSystemService.getCodeSystems());
+	public @ResponseBody Promise<CodeSystems> searchByGet(final CodeSystemRestSearch params) {
```
cmark: The responsebody annotation can be removed, because the class is annotated with `RestController`.
```suggestion
	public Promise<CodeSystems> searchByGet(final CodeSystemRestSearch params) {
```

core/com.b2international.snowowl.core.rest/src/com/b2international/snowowl/core/rest/codesystem/CodeSystemRestService.java
```diff
@@ -58,14 +70,88 @@
 	private CodeSystemService codeSystemService;
 	
 	@ApiOperation(
-			value="Retrieve all code systems",
-			notes="Returns a list containing generic information about registered code systems.")
+		value="Retrieve Code Systems", 
+		notes="Returns a collection resource containing all/filtered registered Code Systems."
+			+ "<p>Results are always sorted by repositoryUuid first, sort keys only apply per repository."
```
cmark, line 76: Rename the `repositoryUuid` field to `repositoryId` in the domain class (we can keep it `repositoryUuid` in the index or replace it with a `repositoryId`, your call).

core/com.b2international.snowowl.core.rest/src/com/b2international/snowowl/core/rest/codesystem/CodeSystemRestService.java
```diff
@@ -58,14 +70,88 @@
...
+		if (sortBy.isEmpty()) {
+			sortBy = List.of(SortField.ascending(CodeSystemEntry.Fields.SHORT_NAME));
+		}
+
+		return CodeSystemRequests.prepareSearchCodeSystem()
+				.filterByIds(params.getId())
+				.setLimit(params.getLimit())
+				.setExpand(params.getExpand())
+				.sortBy(sortBy)
+				// .setSearchAfter(...) is not applied; we are searching in multiple repositories
```
cmark, line 137: I suggest throwing a `BadRequestException` if a client sends a value in the `searchAfter` property of the `CodeSystemRestSearch` params.

core/com.b2international.snowowl.core.rest/src/com/b2international/snowowl/core/rest/codesystem/CodeSystemRestService.java
cmark: The same applies to the other methods.

core/com.b2international.snowowl.core.rest/src/com/b2international/snowowl/core/rest/codesystem/CodeSystemService.java
```diff
@@ -45,32 +43,9 @@
 @Component
 public final class CodeSystemService {
```
cmark, line 44: Do we still need this class? Would it be possible to merge it into the core module?

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/codesystem/CodeSystem.java
```diff
@@ -60,8 +84,10 @@ public static Builder builder(final CodeSystemEntry input) {
 		private String iconPath;
 		private String terminologyId;
 		private String repositoryUuid;
```
cmark: Since we are modifying this model anyway, I suggest renaming this property to `repositoryId`.

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/codesystem/CodeSystemConverter.java
```diff
@@ -0,0 +1,113 @@
...
+
+import com.b2international.commons.http.ExtendedLocale;
+import com.b2international.commons.options.Options;
+import com.b2international.snowowl.core.domain.RepositoryContext;
+import com.b2international.snowowl.core.request.BaseResourceConverter;
+import com.b2international.snowowl.core.uri.CodeSystemURI;
+import com.google.common.collect.Multimaps;
+import com.google.common.collect.TreeMultimap;
+
+public final class CodeSystemConverter extends BaseResourceConverter<CodeSystemEntry, CodeSystem, CodeSystems> {
```
cmark, line 39: ⚠️ Missing since tag.

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/codesystem/CodeSystemEntry.java
```diff
@@ -203,153 +245,150 @@ private CodeSystemEntry(final String oid, final String name, final String shortN
...
+		 *   short name  | ABC       CDE       CDE
+		 *   OID         | 1.2.3.4   1.2.3.4   5.6.7.8
+		 * 
+		 *   In this alternative world, cs1 = cs2 and cs2 = cs3, and so cs1 should be 
+		 *   equal to cs3, but it isn't!
+		 * 
+		 * - While some requests take care to look up both short name and OID for a code system,
+		 *   other parts (client and UI) practically treat short names as unique
+		 * 
+		 * Consider removing "oid" from the equality check.
```
cmark, line 390: How about a few enforcing checks in the `CodeSystem` create API? Prevent creating new `CodeSystem` entries that use the same oid (or shortName) as another `CodeSystem`.
Or remove the possibility to get a `CodeSystem`s by its supposed to be unique oid. See the GET endpoint.

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/codesystem/CodeSystem.java
```diff
@@ -15,19 +15,41 @@
...
 /**
- * Captures metadata about a code system, which holds a set of real-world concepts of medical significance (optionally
- * along with different components forming a description of said concepts) and their corresponding unique code.
+ * Captures metadata about a code system, which holds a set of concepts of
+ * medical significance (optionally with other, supporting components that
+ * together make up the definition of concepts) and their corresponding unique
+ * code.
  */
 @JsonDeserialize(builder=CodeSystem.Builder.class)
 public class CodeSystem {
```
cmark: Since we are using this now in the Java API, this class should implement `Serializable`.

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/repository/TerminologyRepositoryInitializer.java
```diff
@@ -43,7 +43,7 @@
 	 */
 	@Override
 	public final Void execute(RepositoryContext context) {
-		CodeSystemEntry primaryCodeSystem = createPrimaryCodeSystem();
+		CodeSystem primaryCodeSystem = createPrimaryCodeSystem();
```
cmark, line 46: Just a sidenote: we were thinking about removing the repository initializer functionality completely to support full customization of the default Code Systems in a given tooling (for example to create `SNOMEDCT-UK-EDITION` instead of `SNOMEDCT` on the `MAIN` branch).

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/request/rf2/SnomedRf2ExportRequest.java
```diff
@@ -27,8 +27,18 @@
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.time.Instant;
-import java.util.*;
```
cmark: Revert this change, please.

cmark (approved):
LGTM!

core/com.b2international.snowowl.core.rest/src/com/b2international/snowowl/core/rest/codesystem/CodeSystemService.java
apeteri: `CodeSystemRestService` and `CodeSystemVersionRestService` are still sharing the remaining service method. I didn't want to touch this part to avoid expanding the PR's scope.

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/codesystem/CodeSystemEntry.java
apeteri: Code system creation will fail if either of the identifiers is non-unique: https://github.com/b2ihealthcare/snow-owl/blob/37080602ee0c753cf419a761e78ff54af90fd77e/core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/codesystem/CodeSystemCreateRequest.java#L158-L165

The search request looks for the given identifier(s) in an either-or fashion, however: https://github.com/b2ihealthcare/snow-owl/blob/37080602ee0c753cf419a761e78ff54af90fd77e/core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/codesystem/CodeSystemSearchRequest.java#L46-L49

PR #550, opened 2020-04-30
RF2 Export issue

cmark:
[snomed] make sure we return the version in descending effective time order when calculating exportable versions.

Also, make sure cut off base time calculation does not lead to
occasional branch not found exceptions.


nagyo (approved):
LGTM!

apeteri (approved):

PR #553, opened 2020-05-04
SO-3933: Export last released version of unpublished content

AAAlinaaa:


snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/request/rf2/exporter/Rf2Exporter.java
```diff
@@ -156,28 +217,121 @@ public final void exportBranch(
...
-								throw new SnowowlRuntimeException("Failed to write contents for file '" + exportFile.getFileName() + "'.");
-							}
-						});
-				}
+				exportBatch(branch, context, iterator, visitedComponentEffectiveTimes, fileChannel, exportFile.getFileName());
+			}
+		}
+	}
+	
+	private String getLatestVersionPath(RepositoryContext context, String branchPath) {
```
cmark, line 225: Hmm, it would be better to let the request compute the necessary paths to consider when exporting and let the Rf2Exporter just export the data. Now we have a mixed implementation due to one use case, which makes the class harder to understand. We need to come up with a better solution.

cmark (changes requested):

cmark:
Superseded by #1161

PR #554, opened 2020-05-04
[mrcm] Fix authorization issue with MRCM import/export

nagyo:


apeteri (commented):
🆗 
We could hide the impersonation parts in `AuthorizedEventBus` in the future; it is more like an `ExtraHeadersEventBus` at the moment. 😄

PR #560, opened 2020-05-18
Improve iconId calculation to handle new semantic types

apeteri:
1. An identifier is extracted from the semantic tag of the concept's first active FSN (language code can not be taken into account as it is not stored on preferred descriptions). 
1. If an icon with the concept's semantic tag identifier exists, it will be used as the concept's `iconId`
1. Otherwise, a BFS search visits all direct and indirect ancestor concepts, and checks whether an icon exists for the visited concept ID. Only the "closest" of such concepts are taken into account, unless an ancestor is reachable by different IS A routes.
1. If multiple IDs remain to pick from, the smallest SCTID wins, using `String` comparison semantics, otherwise the only result is used, or if no appropriate ID could be found, the identifier of `Snomed CT Concept`.


apeteri:
Current semantic tag identifiers that will be used as iconIDs:

* `clinical_drug`
* `disorder`
* `medicinal_product_form`
* `medicinal_product`
* `product`

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/index/change/SnomedRepositoryPreCommitHook.java
```diff
@@ -152,6 +153,32 @@ protected void preUpdateDocuments(StagingArea staging, RevisionSearcher index) t
...
+			}
+		});
+		
+		staging.getChangedRevisions(SnomedDescriptionIndexEntry.class, Set.of(
+				SnomedDescriptionIndexEntry.Fields.ACTIVE,
+				SnomedDescriptionIndexEntry.Fields.TERM)).forEach(diff -> {
+
+			SnomedDescriptionIndexEntry newRevision = (SnomedDescriptionIndexEntry) diff.newRevision;
+			
+			if (newRevision.isFsn()) {
```
cmark, line 169: ⚠️ Missing active check?

cmark (approved):
LGTM!

It would be great to have a few test cases in the e2e test suite that verifies the new iconId computation logic. Or something when we create concepts in the tests.

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/index/change/SnomedRepositoryPreCommitHook.java
apeteri: A status change in either direction may result in the concept's "official" semantic tag changing.

cmark: Right, then good to go.

PR #563, opened 2020-05-19
Export and Attachment API improvements

cmark:


core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/attachments/Attachment.java
```diff
@@ -113,4 +121,52 @@ public String toString() {
 		return MoreObjects.toStringHelper(getClass()).add("attachmentId", attachmentId).add("fileName", fileName).toString();
 	}
 
+	/**
+	 * Downloads the file to the specified exportPath. The exportPath argument accepts both directory and file paths. Passing a file path replaces the
+	 * default file name with the desired one. If the target path is a file path and there is an existing file, it will automatically overrides the
```
apeteri:
```suggestion
	 * default file name with the desired one. If the target path is a file path and there is an existing file, it will automatically overwrite the
```

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/attachments/Attachment.java
```diff
@@ -113,4 +121,52 @@ public String toString() {
...
+	 * @since 7.7
+	 * @see #download(ServiceProvider, Path)
+	 */
+	public Path download(ServiceProvider context, String exportPath) {
+		return download(context, Paths.get(exportPath));
+	}
+	
+	/**
+	 * Downloads the file to the specified exportPath. The exportPath argument accepts both directory and file paths. Passing a file path replaces the
+	 * default file name with the desired one. If the target path is a file path and there is an existing file, it will automatically overrides the
```
apeteri:
```suggestion
	 * default file name with the desired one. If the target path is a file path and there is an existing file, it will automatically overwrite the
```

apeteri (approved):
Minor nitpick in javadoc, otherwise LGTM!

PR #566, opened 2020-05-22
SO-4092: fix special search key pattern matching error

cmark:
Use simple String manipulation instead of regex when parsing a special
option key's value for field and value instances.

Example search string that fails with the old pattern matcher: `@ecl(123123 (finding))`


apeteri (approved):
👍

PR #568, opened 2020-05-23
SO-4081: add shortcuts to ECL evaluator

cmark:
ID1 OR ID2 OR IDN will result in flattenned ids(ID1, ID2, IDN)
expression (even when using parens).
ID1 AND ID2 AND IDN will result in match NONE expression (even when
using parens).
X MINUS * will resultl in match NONE expression (even when using
parens).
`<`,`<<`,`<!`,`>`,`>>`,`>!`,`^` operators now detect if an any
expression is wrapped in parenthesis (eg. `<(*)`).

### DIFF
--- a/snomed/com.b2international.snowowl.snomed.ecl/src/com/b2international/snowowl/snomed/ecl/Ecl.java
+++ b/snomed/com.b2international.snowowl.snomed.ecl/src/com/b2international/snowowl/snomed/ecl/Ecl.java
@@ -27,6 +27,7 @@ public final class Ecl {
 	public static final String ANY = "*";
 	public static final int MAX_CARDINALITY = -1;
 	public static final Joiner OR_JOINER = Joiner.on(" OR ");
+	public static final Joiner AND_JOINER = Joiner.on(",");
 	
 	private Ecl() {}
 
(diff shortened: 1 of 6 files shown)


snomed/com.b2international.snowowl.snomed.ecl/src/com/b2international/snowowl/snomed/ecl/Ecl.java
```diff
@@ -27,6 +27,7 @@
 	public static final String ANY = "*";
 	public static final int MAX_CARDINALITY = -1;
 	public static final Joiner OR_JOINER = Joiner.on(" OR ");
+	public static final Joiner AND_JOINER = Joiner.on(",");
```
nagyo: Could this introduce confusion when an expression is deserialized for representation on the UI? I mean for disjunction we use 'OR' where there is no alternative, but for conjunction we use ',' where there is a written alternative 'AND'? (this does not take the credit for the PR itself, just being :philosoraptor:)

nagyo (approved):
Other than the minor comment LGTM!

snomed/com.b2international.snowowl.snomed.ecl/src/com/b2international/snowowl/snomed/ecl/Ecl.java
cmark: Hmm, not sure, I assume if someone is writing ECL, he knows the actual ECL syntax, and commas are used in conjunctions. To simplify and make it consistent, I can change it to use the actual `AND` word as separator, instead of the `,`.

PR #570, opened 2020-05-27
Configure default Elasticsearch disk thresholds

cmark:
Enable 20/10/5g values for Elasticsearch disk usage monitor instead of the default 85%,90%,95%.


nagyo (approved):
LGTM!

PR #571, opened 2020-05-27
SO-4095: support active filter in generic ConceptSearchRequest

cmark:
Suggest only active content in `ConceptSuggestionRequest`.


apeteri (approved):
👍

PR #579, opened 2020-06-03
Add generic ComponentURI representation

AAAlinaaa:


cmark (changes requested):
As discussed over chat, this requires additional work, mainly:
* Change the TerminologyComponentURI class to ComponentURI and make it sure to properly represent the same data structure as it does now on client APIs.
* Remove the SnomedConceptDocument index schema changes, they are unnecessary, there is a method that can be used to convert a component shortCode to its String ID (see `TerminologyRegistry`)

PR #582, opened 2020-06-08
[icons] Include additional icon hierarchies based on semantic tags

apeteri:
With a freshly imported dataset, the icons mentioned in the original ticket look like the following:

image


cmark (approved):
LGTM!

PR #584, opened 2020-06-10
SO-4100: Extra concept non-current member fix

adeakgabi:
When inactivating-reactivating and inactivating the concept again, there will be only a single inactivation reason member on each component.
When reactivating the concept its descriptions with "Concept non-current" indicator will be deleted.


snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/index/change/ComponentInactivationChangeProcessor.java
```diff
@@ -83,6 +84,14 @@ public void process(StagingArea staging, RevisionSearcher searcher) throws IOExc
 				changedMembersByReferencedComponentId.put(((SnomedRefSetMemberIndexEntry) diff.newRevision).getReferencedComponentId(), diff);
 			});
 			
+			Map<String, SnomedRefSetMemberIndexEntry> refSetMembers = Maps.newHashMap();
+			for (Hits<SnomedRefSetMemberIndexEntry> hits : searcher.scroll(Query.select(SnomedRefSetMemberIndexEntry.class)
```
cmark: Do we actually need all indicators when there is an inactivated Concept? This can be really costly. 
I suggest limiting the number of loaded indicators to the actually used set (descriptions of inactivated concepts and only active inactivation members).
Also, any duplicates that are present in the dataset will be omitted by the code, since this HashMap can only reference a single member by the referenced component, while a referenced component can have more than one indicator (active, inactive, potential duplicates).

cmark (changes requested):

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/index/change/ComponentInactivationChangeProcessor.java
```diff
@@ -83,6 +85,21 @@ public void process(StagingArea staging, RevisionSearcher searcher) throws IOExc
 				changedMembersByReferencedComponentId.put(((SnomedRefSetMemberIndexEntry) diff.newRevision).getReferencedComponentId(), diff);
 			});
 			
+			final ImmutableSet<String> valueIds = ImmutableSet.of(Concepts.CONCEPT_NON_CURRENT);
+			final Expression expression = Expressions.builder()
+					.filter(SnomedRefSetMemberIndexEntry.Expressions.referenceSetId(Concepts.REFSET_DESCRIPTION_INACTIVITY_INDICATOR))
```
cmark: Can we restrict the number loaded members to just the necessary set of members?
How about restricting to the actually read values, descriptions in the transaction?

cmark (changes requested):

cmark (approved):
LGTM!

apeteri (approved):
👍

PR #588, opened 2020-06-16
SO-4091: Add semantic tag icon IDs

apeteri:
An addition to https://snowowl.atlassian.net/browse/SO-4091 that intends to resolve ambiguous cases where two direct parents have a different icon ID.


cmark (approved):
LGTM!

PR #596, opened 2020-06-25
[core] make TransactionContexts reusable for subsequent commits

cmark:
Support CodeSystemURI in all branch path based requests
Deprecate old repositoryId + branch request build methods.
Add throws declaration to LockRequest.doExecute(C).


apeteri (approved):
👍

PR #601, opened 2020-07-01
[core] Use single boolean property in TerminologyComponent annotation...

apeteri:
...to determine whether a component is allowed to participate in a
SNOMED CT map type reference member as the map target.


cmark (approved):
LGTM!

PR #602, opened 2020-07-01
[core] Use single boolean property in TerminologyComponent annotation...

cmark:
...to determine whether a component is allowed to participate in a
SNOMED CT map type reference member as the map target.


apeteri (approved):

apeteri:
I have already added the commit through a squash merge, sorry about the confusion.

PR #609, opened 2020-07-15
SO-4142: Support for setting target code system for Map type refsets

bbanfai:
Please review along with the changes in the ext repo.


nagyo (approved):
LGTM!

PR #610, opened 2020-07-16
SO-4025: Concept Map Compare

molnarlaura:
Add logic for comparing sets.

PR #614, opened 2020-07-21
SO-4131: Mapping Set and Reference Set Automapping features

kovacstomi:


cmark:
Will be merged along with PR #610 . Closing.

PR #617, opened 2020-07-24
Chr 2610 hardcode attempt

martinmart92:


martinmart92:
wrong place for this PR, sorry about this.

PR #618, opened 2020-07-27
SO-4168: Save concept without ungrouped owl axiom relationship

molnarlaura:


nagyo:
This issue will be resolved by validation, hence it should not be merged .

PR #621, opened 2020-07-29
Branch rebase/merge missing content issue

cmark:
This PR fixes the branch state calculation issue, which incorrectly returns UP_TO_DATE state of a branch that has actual changes, but due to a recent synchronization with the parent branch, it states that it does not have any changes.

There are multiple places that have been changed to fix the issue:
* Introduced a new `squashMerge` boolean flag on Commit documents to detect if a commit was a squash merge or not.
* Take actual commits into account before deciding whether a branch is in FORWARD or BEHIND state.
* Only merge commits can be empty from this time forward. After applying the changes in a StagingArea, double-check if there are any actual changes before adding the commit document and updating the branch document.
* Empty segments are no longer returned from `RevisionBranchRef.difference(RevisionBranchRef)` method.

Unrelated, clean up changes:
* Remove unused, obsolete branch doc REPLACE script (used in 6.x version only)


nagyo (approved):
LGTM!

PR #612, opened 2020-07-20
SO-4151: When finding unpublished duplicate FSNS, ensure they are ...

AAAlinaaa:
... compared to every other FSN and not only other unpublished FSNs

PR #625, opened 2020-08-03
Revert "SO-4151: When finding unpublished duplicate FSNS, ensure they are ..."

cmark:
Reverts b2ihealthcare/snow-owl#612

This causes an unexpected increase in validation execution time, so reverting change.


apeteri (approved):
LGTM

PR #627, opened 2020-08-05
[validation] fix missing limit setting when collecting attributes...

cmark:
...in rule110.
Also, report members and relationships individually instead of reporting
just the containing concepts.


apeteri (approved):
👍

PR #636, opened 2020-08-13
SO-4147: Expose parent lock context

AAAlinaaa:


cmark (commented):
This looks fine for now, but later it would be great to actually get this from the service context, since multiple locks can be wrapped into one an other, so we will get the inner most context in all cases.

cmark (approved):

PR #638, opened 2020-08-14
Fix for task synchronization conflicts

apeteri:


snomed/com.b2international.snowowl.snomed.core.rest.tests/src/com/b2international/snowowl/snomed/core/rest/branches/SnomedMergeApiTest.java
```diff
@@ -881,5 +881,42 @@ public void mergeThenRebaseOtherTask() throws Exception {
...
+
+		// Synchronization makes parentConcept visible on task 2
+		getComponent(task2, SnomedComponentType.CONCEPT, parentConcept).statusCode(404);
+		merge(branchPath, task2, "Synchronize task 2").body("status", equalTo(Merge.Status.COMPLETED.name()));
+		getComponent(task2, SnomedComponentType.CONCEPT, parentConcept).statusCode(200);
+
+		// Synchronization makes parentConcept visible on task 1
+		getComponent(task1, SnomedComponentType.CONCEPT, parentConcept).statusCode(404);
+		merge(branchPath, task1, "Synchronize task 1").body("status", equalTo(Merge.Status.COMPLETED.name()));
+		getComponent(task1, SnomedComponentType.CONCEPT, parentConcept).statusCode(200);
```
cmark, line 905: Does this synchronization still cause an extra revision (which is equivalent with the parent branch revision) on the child branch? If not, then it would worth adding an extra low-level check here to make sure there is no actual revision registered for `parentConcept` on the task2/task1 branches.

cmark (approved):
LGTM!

PR #645, opened 2020-08-19
SO-4186: Fix icon ID calculation in case there are multiple preferred FSNs for a concept

nagyo:
- order preferred descriptions by:
  - description type:
    - preferred US FSN
    - preferred GB FSN
    - SYNONYMs
    - all other description types
  - description id length
  - description id

https://snowowl.atlassian.net/browse/SO-4186


cmark (approved):
LGTM!

PR #646, opened 2020-08-24
SO-4229 support change vs remove use case in branch compare

cmark:
Also, eliminate property changes from compare if the component got removed at the end.


apeteri (approved):
👍

PR #648, opened 2020-08-25
SO-4225: Avoid creating duplicate description inactivation indicator...

AAAlinaaa:
... members during concept inactivation


nagyo (approved):
LGTM!

PR #649, opened 2020-08-26
SO-4198: LCS replace type import fails with NPE

molnarlaura:
The return type of commit() was changed to Long to be able to return null.


core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/repository/RepositoryTransactionContext.java
```diff
@@ -276,11 +276,11 @@ public long commit(String author, String commitComment, String parentLockContext
 		Commit commit = null;
 		try {
 			acquireLock(locks, lockContext, lockTarget);
-			final long timestamp = service(TimestampProvider.class).getTimestamp();
+			final Long timestamp = service(TimestampProvider.class).getTimestamp();
```
apeteri: ⚠️ See lgtm-bot's notice: the local variable `timestamp` does not need to be changed into a `Long` value, we just need to be able to return `null` at the end.

nagyo (approved):
LGTM!

PR #650, opened 2020-08-27
[snomed] Set referenced component type in query RF2 importer to concept

apeteri:


apeteri:
Some clarification: the new reference set wizard sets this component type for new query type reference sets, and additional logic in the UI ensures that the reference set editor is opened for each member, not the concept editor for the identifier concept.

cmark:
Closing PR. No longer needed on the 6.x branch.

PR #651, opened 2020-09-01
Deprecate SNOMEDCT code system shortname and repositoryId constants

cmark:
They will be removed in Snow Owl 8.0.
Clients should create their own SNOMED CT CodeSystem with their desired short name to designate the content they would like to create in Snow Owl.


apeteri (approved):
LGTM

PR #661, opened 2020-09-11
[commons] Add new string utils function

AAAlinaaa:


commons/com.b2international.commons.base/src/com/b2international/commons/StringUtils.java
```diff
@@ -268,7 +269,26 @@ public static boolean containsNumeric(final String s) {
...
+	
+	/**
+	 * Returns {@code true} if the specified string contains only numeric characters, {@code false} otherwise.
+	 * @param s the string to check. Can be {@code null}. If {@code null} this method returns {@code true}.
+	 * @return {@code true} if specified string contains only numeric characters.
+	 */
+	public static boolean isNumeric(final String s) {
+		
+		if (Strings.isNullOrEmpty(s)) {
+			return true;
```
cmark, line 282: Not sure what is the semantic of this return value. I understand you have stated this in the javadoc, but just looking at the method name I would expect it to return `true` only and only if the given input `String` represents a number, and return `false` otherwise. Since this is currently unused in this repository, I suggest moving this method closer to the current call site.

cmark (changes requested):

commons/com.b2international.commons.base/src/com/b2international/commons/StringUtils.java
apeteri: This was needed for a NICIP PR, but is now obsolete.

apeteri:
Closing PR.

PR #663, opened 2020-09-17
SO-4280: Enforce new rules to determine definition status of a concept

AAAlinaaa:
* If a concept only has equivalent class axioms make it defined
* If a concept only has sub class axioms make it primitive
* If a concept contains a mixture make it primitive as well

### DIFF
--- a/snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/request/SnomedOWLAxiomHelper.java
+++ b/snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/request/SnomedOWLAxiomHelper.java
@@ -26,7 +26,7 @@ import com.google.common.base.Strings;
  */
 public final class SnomedOWLAxiomHelper {
 
-	private static final String EQUIVALENTCLASSES = "equivalentclasses";
+	private static final String SUBCLASSOF = "subclassof";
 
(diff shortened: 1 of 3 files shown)


apeteri (changes requested):
Please re-formulate the expression so that the definition status only becomes `FULLY_DEFINED` if - ignoring any GCI axioms - all remaining axioms are of type "equivalent classes". 0 axioms of such kind should still result in `PRIMITIVE` status, which doesn't happen with the code in its current form.

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/request/SnomedOWLAxiomHelper.java
```diff
@@ -35,7 +35,7 @@ public static String getDefinitionStatusFromExpressions(Set<String> owlExpressio
 		
 		return owlExpressions.stream()
 				.filter(expression -> !Strings.isNullOrEmpty(expression))
```
apeteri, line 37: Filter GCI axioms here (in class `SnomedOWLAxiomHelper`), not in the create or update requests. They don't have a say in the concept's suggested definition status, but should be going through when creating or updating a SNOMED CT concept, just as before.

Or did they cause some other problem?

apeteri (changes requested):

cmark:
Putting this to on hold. Not sure about the proper logic here, need more information to decide.

PR #664, opened 2020-09-18
SO-4289: incorrect reuse of indicator member

cmark:
This PR addresses an issue with the automatic generation of concept non-current members during concept inactivation.
In cases when the concept and its descriptions were released earlier with active pending move indicator members, during inactivation, the descriptions got indicator members with incorrect effective dates due to a bug in the inactivation process.


apeteri (approved):
👍

PR #667, opened 2020-09-20
[lcs] Auxiliary support for configurable local code systems

apeteri:
- Search `CodeSystem`s by name (exact match, all prefixes, regex, original)
- Add utility method for splitting branch paths into segments
- Expose a package from Hibernate Validator so that custom validators can use the same constraint violation reporting class


cmark (approved):
LGTM!

The change looks backward compatible, but before merging, @apeteri, could you please confirm that it indeed works with pre-7.10 index schemas (CodeSystemEntry mapping has changed)? Thank you!

apeteri:
The use of field aliases should allow for earlier datasets, LCS code systems will have to be removed and re-imported.

image

cmark:
Thanks!

PR #674, opened 2020-09-22
[validation] Preserve locale order in computing issue labels

AAAlinaaa:


cmark (approved):
LGTM!

Rename `EXTENED_LOCALES` property to `LOCALES` and also make sure return no FSN or no PT for the concept in the new `indexBestPreferredByConceptId` method.

PR #679, opened 2020-09-24
ConceptMapMapping model improvements

kovacstomi:


snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/request/SnomedConceptMapSearchRequestEvaluator.java
```diff
@@ -61,43 +63,77 @@
 
 	@Override
 	public ConceptMapMappings evaluate(CodeSystemURI uri, BranchContext context, Options search) {
-		SnomedReferenceSetMembers referenceSetMembers = fetchRefsetMembers(uri, context, search);
-		return toCollectionResource(referenceSetMembers, uri, context, search);
+		final String preferredDisplay = search.get(OptionKey.DISPLAY, String.class);
+		SnomedDisplayTermType snomedDisplayTermType;
+		try {
```
cmark: This try-catch should be part of the SnomedDisplayTermType enum, unrecognized preferredDisplay strings should either:
* Fail with a clear error message
* Or fall back to the FSN as default, PT might not be available and might not be appropriate to use

cmark (changes requested):

nagyo (approved):
LGTM!

PR #684, opened 2020-10-05
SO-4243: Improve automapping lexical search

kovacstomi:


core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/repository/RevisionDocument.java
```diff
@@ -54,6 +56,14 @@ public static Expression idPrefix(String idPrefix) {
 		public static Expression idRegex(String idRegex) {
 			return regexp(Fields.ID, idRegex);
 		}
+		
+		public static Expression anyFieldValuePresent(final String field, final String value, final int minShouldMatch) {
```
cmark: These are just proxies over the `Expressions` class static method, they do not add actual value to the `RevisionDocument`, please remove these and use the `Expressions` methods directly instead.

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/request/SearchResourceRequest.java
```diff
@@ -212,10 +212,15 @@ public String toString() {
 	
 	@Min(0)
 	private int limit;
+	
+	@Min(0)
+	private int minTermMatch;
```
cmark: These new fields should be part of the options map instead. The others will be part of the options map later, but for now, please move these new fields inside the option map, you can use a new local OptionKey to access them and provide the same accessors as for the other fields.

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/request/SearchResourceRequestBuilder.java
```diff
@@ -42,6 +42,10 @@
 	
 	private int limit = 50;
 	
+	private int minTermMatch = 0;
```
cmark: What does the default value `0` mean? How can I ensure my search does not use this feature? What happens with existing searches that use this generic API?

cmark (changes requested):

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/request/SearchResourceRequestBuilder.java
kovacstomi: Fixed this. To enable the new feature this parameter must be configured.

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/request/SearchResourceRequestBuilder.java
```diff
@@ -151,7 +147,9 @@ public final B one() {
 	 * @return this builder instance
 	 */
 	public final B withMinTermMatch(int minTermMatch) {
-		this.minTermMatch = minTermMatch;
+		if (minTermMatch > 0) {
```
cmark: Throw a `BadRequestException` if minTermMatch is less than or equals to `0`.

cmark (changes requested):

cmark (approved):

PR #686, opened 2020-10-06
SO-4356: expose squash merge configuration property

cmark:
Expose squash property in merge API to make it possible to merge content
between branches without actually moving the content from the source
branch to the target branch.


nagyo (approved):
LGTM!

PR #688, opened 2020-10-07
SO-4306: include Value Set members in FSN/PT switch

kovacstomi:


core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/domain/Concept.java
```diff
@@ -45,28 +47,32 @@ public String getCodeSystem() {
 		return codeSystem;
 	}
 	
-	public String getTerm() {
-		return term;
+	public String getPreferredTerm() {
```
cmark: Not sure if it is a good idea to call this term a preferred term. If you would like to rename it, then let's use FHIR phrases, like designation or display here. `getDisplay` or `getDisplayTerm` would be alternatives I can think of.

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/core/domain/SnomedDescription.java
```diff
@@ -227,6 +229,29 @@ public String getCaseSignificanceId() {
 		return acceptabilities;
 	}
 	
+	/**
+	 * Return the common description equivalent of the description type identifier
+	 * 
+	 * @see Description
+	 * @return common description type
+	 */
+	public Description.Type getCommonDescriptionType() {
```
cmark: There are multiple issues with this method:
* Introduces a new JSON property in the serialized form of this representation class
* Returns UNKNOWN for synonym subtypes
* Returns UNKNOWN for partially loaded descriptions
Let's chat, why do you need this new enum type and where are you going to use it.

cmark (dismissed):

cmark:
@kovacstomi Could you please fix the conflict and also the spurious javadoc reported by LGTM? Thanks!

nagyo (approved):
LGTM!

PR #690, opened 2020-10-11
Configurable bulk commits

cmark:
This PR makes it possible to configure the previously hardcoded `10_000` bulk action items and `10MB` bulk action item size in MB used during commit operations. Configuring the two values are important when working with Elasticsearch clusters with restricted HTTP body sizes (like AWS Elasticsearch service).

### DIFF
--- a/commons/com.b2international.index/src/com/b2international/index/IndexClientFactory.java
+++ b/commons/com.b2international.index/src/com/b2international/index/IndexClientFactory.java
@@ -106,6 +106,16 @@ public interface IndexClientFactory {
 	 * Configuration key to specify the time to wait until yellow cluster health state is reached in milliseconds.
 	 */
 	String CLUSTER_HEALTH_TIMEOUT = "clusterHealthTimeout";
+	
+	/**
+	 * Configuration key to specify the number of bulk action items to send to the Elasticsearch cluster when performing bulk commit operations
+	 */
+	String BULK_ACTIONS_SIZE = "bulk_action_size";
+	
+	/**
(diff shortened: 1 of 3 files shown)


commons/com.b2international.index/src/com/b2international/index/IndexClientFactory.java
```diff
@@ -158,6 +168,10 @@
 	 * The default cluster.name value for embedded nodes and tcp based clients.
 	 */
 	String DEFAULT_CLUSTER_NAME = "elastic-snowowl";
+	
+	int DEFAULT_BULK_ACTIONS_SIZE = 10_000;
+			
+	int DEFAULT_BULK_ACTIONS_SIZE_IN_MB = 10;
```
apeteri: I'm wondering if we should go a under the posted limit, in case the final request size goes over 10 MB with some overhead not accounted for. 🤔 

For future reference: https://docs.aws.amazon.com/elasticsearch-service/latest/developerguide/aes-limits.html#network-limits

apeteri (approved):
LGTM, please see the comment below.

commons/com.b2international.index/src/com/b2international/index/IndexClientFactory.java
cmark: BulkProcessor should handle that well, I think.

Also, the smallest AWS ES instance limit is `10 MiB`, which is `10.048576 MB`.
If we limit ourselves to 10 MB by default, that should be less than the AWS 10 MiB limit.

cmark: After a quick discussion, we have decided to set the limit to 9 MB by default to allow a bit of headroom for other parts in the HTTP request.

PR #695, opened 2020-10-13
Support List values in search options

cmark:
This PR improves the API to store List search option values as Lists instead of converting them to distinct values in a Set.


nagyo (approved):
LGTM!

PR #696, opened 2020-10-17
Fix potential issue when merging revisions with no actual tracked field changes

cmark:
Either no changes or derived field only changes can cause duplicate revisions when merging a branch into another.
Commit documents now contain a change entry for all revision objects
regardless of whether they had a tracked property change or not.

### DIFF
--- a/commons/com.b2international.index.tests.tools/src/com/b2international/index/revision/BaseRevisionIndexTest.java
+++ b/commons/com.b2international.index.tests.tools/src/com/b2international/index/revision/BaseRevisionIndexTest.java
@@ -44,6 +44,7 @@ import com.fasterxml.jackson.databind.ObjectMapper;
  */
 public abstract class BaseRevisionIndexTest {
 	
+	protected static final String USER_ID = UUID.randomUUID().toString();
 	protected static final String MAIN = RevisionBranch.MAIN_PATH;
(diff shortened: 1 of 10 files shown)


apeteri (changes requested):
👍 , but now some of the commit detail assertions are breaking:

* historyOfChangedComponent(RevisionHistoryTest.java:67)
* compareBranchWithRevertedChanges(RevisionCompareTest.java:194)
* compareBranchWithChangedComponent(RevisionCompareTest.java:121)
* compareChangeOnMainSinceBranchBasePoint_Reverse(RevisionCompareTest.java:102)
* compareNestedRevisionData(NestedDocumentRevisionIndexTest.java:94)

Was the only issue with this approach earlier that it creates larger commit documents? IIRC we have considered including container roots and derived-only changes before, but have abandoned it at some point.

cmark:
> Was the only issue with this approach earlier that it creates larger commit documents? IIRC we have considered including container roots and derived-only changes before, but have abandoned it at some point.

Yes, but this is required to detect revision with derived-only field changes during compare and then merge/rebase them properly later. Old commit documents will continue to work, but in certain cases, manual change is needed if after a rebase a branch experiences duplicate revisions.

apeteri (approved):
👍

PR #698, opened 2020-10-21
SO-4236: Map compare enhancements

molnarlaura:


core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/compare/ConceptMapCompareProperties.java
```diff
@@ -0,0 +1,40 @@
+package com.b2international.snowowl.core.compare;
```
cmark: Add our Apache 2.0 license and since tag please.

cmark (changes requested):

cmark (approved):
LGTM!

PR #700, opened 2020-10-21
[classification] cache computed direct/indirect destination edges...

cmark:
...in InternalIdEdges to speed up normal form generation drastically.
This yields approximately an 80% speed improvement on the entire classification process without any initial data caching.


apeteri (approved):
👍

PR #705, opened 2020-10-29
Allow filtering stop words when executing automapping

kovacstomi:


cmark:
Too much API changes, this requires more thinking before blindly changing the underlying term search APIs. Rejecting.

PR #709, opened 2020-11-07
SO-4460: track array property changes

cmark:


core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/repository/RepositoryTransactionContext.java
```diff
@@ -280,7 +280,7 @@ public Long commit(String author, String commitComment, String parentLockContext
 			log().info("Persisting changes to {}@{}", path(), timestamp);
 			commit = staging.commit(null, timestamp, author, commitComment);
 			log().info("Changes have been successfully persisted to {}@{}.", path(), timestamp);
-			return commit == null ? null : commit.getTimestamp();
+			return commit == null ? Commit.NO_COMMIT_TIMESTAMP : commit.getTimestamp();
```
apeteri, line 283: It turns out there is exactly one caller of `TransactionContext#commit` that is interested in the timestamp, returned via a `CommitResult` instance: https://github.com/b2ihealthcare/snow-owl/blob/7f5e60d95cfba13301a842b43f5455660d2197e0/snomed/com.b2international.snowowl.snomed.reasoner/src/com/b2international/snowowl/snomed/reasoner/request/SaveJobRequest.java#L216

commons/com.b2international.index/src/com/b2international/index/revision/Revision.java
```diff
@@ -73,10 +76,6 @@ final void setCreated(RevisionBranchPoint created) {
 		this.created = created;
 	}
 	
-	final void setRevised(List<RevisionBranchPoint> revised) {
```
apeteri, line 76: Please add a comment somewhere that this field is only updated in Elasticsearch via the `updateRevised` script defined at the top, is never de-serialized (so if one looks at a `Revision` instance, it will not contain the revised branch+timestamp pairs) and is used in ES search expressions only.

I got a bit confused when I saw that the setter could be removed, Jackson is only using the field when converting to JSON, it is always initialized with an empty list and no one is interested in the contents in Java either 😃

commons/com.b2international.index/src/com/b2international/index/revision/Revision.java
```diff
@@ -122,15 +112,6 @@ protected ToStringHelper doToString() {
 				.add(Revision.Fields.REVISED, revised);
 	}
 	
-	/**
-	 * Converts an immutable object into a mutable builder with the usual prefixless setter methods.
-	 * 
-	 * @return
-	 */
-	protected Builder<?, ? extends Revision> toBuilder() {
```
apeteri, line 130: The inner `Builder` class could also be pushed down to `RevisionDocument`, like this method.

commons/com.b2international.index/src/com/b2international/index/revision/StagingArea.java
```diff
@@ -859,19 +849,39 @@ private RevisionDiff(Revision oldRevision, Revision newRevision) {
...
+						change = new RevisionPropertyDiff(property, serializeToCommitDetailValue(oldRevisionSource, property), serializeToCommitDetailValue(newRevisionSource, property))
+								.asPatch(mapper, oldRevisionSource, true /* includeFromValue */);
+						if (fieldsToSkip == null) {
+							fieldsToSkip = Sets.newHashSetWithExpectedSize(1); // expect a single array like property per type
+						}
+						fieldsToSkip.add(property);
+					}
 					
 					// Remove administrative revision fields from diff, but keep all other ones
 					if (!Revision.Fields.CREATED.equals(property) && !Revision.Fields.REVISED.equals(property)) {
```
apeteri: These two properties could be added to `fieldsToSkip` at initialization time. Otherwise `Revision.Fields.REVISED` will needlessly undergo the collection-like conversion above 🤔

apeteri (changes requested):
Please address the smaller issues reported below, and then it's good to go!

commons/com.b2international.index/src/com/b2international/index/revision/Revision.java
cmark: Yes, true, but there is one class that has this method implemented in the `index.tests` sources.

apeteri: It is used by a handful of rebase tests in `RevisionBranchMergeConflictTest` which could be replaced with `withXYZ` methods (making the builder on RevisionData also redundant, and allowing this refactor to take place). They take the revision and change a single field on it.

apeteri (approved):
👍

PR #710, opened 2020-11-09
SO-4422: Add filter for OID in code system search

apeteri:
Also make the OID property optional when creating new code systems. It should still be unique if specified.


apeteri (commented):
👍

cmark (approved):
LGTM!

PR #711, opened 2020-11-11
SO-4358: Add comments to Concept Map API

kovacstomi:


cmark:
Let's remove the comment pieces for now and keep only the `all()` helper method. Thanks!

cmark (changes requested):
Let's remove the comment pieces and keep only the new helper `all()` method.

cmark (approved):
LGTM!

PR #717, opened 2020-11-19
Speed up classpath scanning

cmark:
This PR improves the performance of the current ClassPathScanner by reducing the visibility to the packages that will certainly have at least one component annotated class.
This reduces the time it takes to scan the classpath for services to around 1 sec. (95% decrease from ~15-20 seconds).


apeteri (approved):

nagyo (approved):
So far LGTM!

PR #718, opened 2020-11-20
SO-4431: Exclude items from the "from" change set early in the...

apeteri:
...merging process in StagingArea, exempting them from conflict detection and resolution


cmark (approved):
LGTM! This looks great. Thank you!

PR #721, opened 2020-11-26
[index] log DEPRECATION level messages as WARN message in Slf4jAppender

cmark:
Also, log level name and message if trying to handle a message with an unrecognized level.


apeteri (approved):
🍃

PR #723, opened 2020-11-27
Fix default doc sorting issues

cmark:
This PR fixes the issues after migrated to `_doc` sorting in commit https://github.com/b2ihealthcare/snow-owl/commit/44d42182359a53732a83c22aeea1ac4c7af7a017
* Revert some of the logic to still use `_id` for searchAfter based paging if no `ID` field (the new ID annotation) is defined for the mapping.
* Use `_doc` as a tiebreaker for localScroll and for and in case of explicit scroll API use.

### DIFF
--- a/commons/com.b2international.index/src/com/b2international/index/es/EsDocumentSearcher.java
+++ b/commons/com.b2international.index/src/com/b2international/index/es/EsDocumentSearcher.java
@@ -396,7 +396,7 @@ public class EsDocumentSearcher implements Searcher {
 					// XXX: default order for scores is *descending*
 					reqSource.sort(SortBuilders.scoreSort().order(sortOrder)); 
 					break;
-				case "_default":
+				case "_default": //$FALL-THROUGH$
 					if (liveScroll) {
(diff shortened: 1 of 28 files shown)


commons/com.b2international.index/src/com/b2international/index/es/EsDocumentSearcher.java
```diff
@@ -397,8 +396,14 @@ private void addSort(DocumentMapping mapping, SearchSourceBuilder reqSource, Sor
 					// XXX: default order for scores is *descending*
 					reqSource.sort(SortBuilders.scoreSort().order(sortOrder)); 
 					break;
-				case DocumentMapping._DOC: //$FALL-THROUGH$
-					field = DocumentMapping._DOC;
+				case "_default":
```
apeteri: Please add the fall-through marker back; it is hard to spot the absence of a `break` statement, sometimes even with the comment present 😄👓

apeteri (approved):

PR #724, opened 2020-11-28
[core] Add property to ConceptMapMapping for indicating if a match...

apeteri:
...is approximate (the source or the target identifier matches, but code system information is inexact or unknown, so we can't tell if the component matches as well).


cmark (approved):
LGTM!

PR #730, opened 2021-01-04
Add support method to SCT description domain.

kovacstomi:
Check whether the description is preferred in a list of locale.

### DIFF
--- a/snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/core/domain/SnomedDescription.java
+++ b/snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/core/domain/SnomedDescription.java
@@ -17,6 +17,7 @@ package com.b2international.snowowl.snomed.core.domain;
 
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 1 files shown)


snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/core/domain/SnomedDescription.java
```diff
@@ -227,6 +229,20 @@ public String getCaseSignificanceId() {
 		return acceptabilities;
 	}
 	
+	@JsonIgnore
+	public boolean isPreferredInLocales(final List<ExtendedLocale> locales) {
+		for (ExtendedLocale locale: locales) {
+			for (AcceptabilityMembership membership: acceptabilities) {
```
cmark: `acceptabilites` field can be null if the client did not request it explicitly, however the `acceptabilityMap` is not null by default, I suggest falling back to the `acceptabilityMap` in case the `acceptabilities` field is `null`.

cmark (changes requested):

cmark (approved):
LGTM!

PR #731, opened 2021-01-04
SO-4136: Validate LDAP identity settings at server startup

apeteri:
- Test connection settings given in the LDAP identity provider configuration
- Make sure role objects can be read properly if the connection attempt succeeds
- Report error on failure; do not let the server continue

### DIFF
--- a/core/com.b2international.snowowl.identity.ldap/src/com/b2international/snowowl/identity/ldap/LdapIdentityPlugin.java
+++ b/core/com.b2international.snowowl.identity.ldap/src/com/b2international/snowowl/identity/ldap/LdapIdentityPlugin.java
@@ -29,9 +29,7 @@ public final class LdapIdentityPlugin extends Plugin implements IdentityProvider
 
 	@Override
 	public IdentityProvider create(Environment env, LdapIdentityProviderConfig configuration) throws Exception {
-		final LdapIdentityProvider identityProvider = new LdapIdentityProvider(configuration);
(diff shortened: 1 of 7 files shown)


core/com.b2international.snowowl.identity.ldap/src/com/b2international/snowowl/identity/ldap/LdapIdentityPlugin.java
```diff
@@ -29,12 +29,13 @@
 
 	@Override
 	public IdentityProvider create(Environment env, LdapIdentityProviderConfig configuration) throws Exception {
-		return new LdapIdentityProvider(configuration);
+		final LdapIdentityProvider identityProvider = new LdapIdentityProvider(configuration);
+		identityProvider.testLdapSettings();
```
cmark: It would be great to abstract and pull this up to the `IdentityProvider` with some awesome name later on and use it for all identity provider implementations.

cmark (approved):
LGTM!

core/com.b2international.snowowl.identity.ldap/src/com/b2international/snowowl/identity/ldap/LdapIdentityPlugin.java
apeteri: I did this; it's not getting much use at the moment, but can be useful for future identity providers that require connection to a remote server.

PR #733, opened 2021-01-06
SO-4590: support random Base64 UUID generation

cmark:
Add a single method to allow plug-in developers to use random Base64 UUIDs instead of just time-based Base64 UUIDs.


nagyo (approved):
LGTM!

PR #735, opened 2021-01-07
SO-3964: upgrade effective time restore

cmark:
This PR introduces a new Upgrade endpoint to upgrade Code System to a newer version and also ensures that effective time are restored from the correct branch when you have a deep extension scenario with an ongoing upgrade.


core/com.b2international.snowowl.core.rest/src/com/b2international/snowowl/core/rest/codesystem/CodeSystemRestService.java
```diff
@@ -243,7 +244,7 @@ public void update(
 		final String commitComment = String.format("Updated Code System %s", shortNameOrOId);
 		
 		if (codeSystem.getUpgradeOf() != null) {
-			throw new BadRequestException("'upgradeOf' property cannot be set through code system update API");
+			LoggerFactory.getLogger(getClass()).warn("'upgradeOf' property update support will be dropped in 8.0. It is only present in 7.x version for backward compatibility to let pre-8 datasets migrate their custom made upgrade CodeSystems work as upgrades.");
```
apeteri: The message at the end becomes a bit vague; I think I understand what the entire thing means 😅, but it could be shortened to something like this:

```suggestion
LoggerFactory.getLogger(getClass()).warn("'upgradeOf' property update support is only present in version 7.x and will be dropped with version 8.0.");
```

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/index/change/ComponentEffectiveTimeRestoreChangeProcessor.java
```diff
@@ -201,85 +190,29 @@ private boolean canRestoreEffectiveTime(SnomedDocument componentToRestore, Snome
...
-						break;
-					}
-				}
-
-			});
-
+		
+		// always check the direct extensionOf (aka parent) CodeSystem
+		if (relativeCodeSystem.getExtensionOf() != null) {
+			if (relativeCodeSystem.getExtensionOf().isHead()) {
```
apeteri, line 216: In `CodeSystemCreateRequest`, `extensionOf` is always mapped to a code system version entry of the parent code system, this should not happen:

https://github.com/b2ihealthcare/snow-owl/blob/c8067e26c1610c2c6ad86f6994b90cffe7d8c93e/core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/codesystem/CodeSystemCreateRequest.java#L172-L175

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/codesystem/CodeSystemRequests.java
```diff
@@ -68,6 +69,14 @@ public static CodeSystemVersionCreateRequestBuilder prepareNewCodeSystemVersion(
 		return new CodeSystemVersionCreateRequestBuilder();
 	}
 	
+	// Upgrade API
+	
+	public static CodeSystemUpgradeRequestBuilder prepareUpgrade(CodeSystemURI codeSystem, CodeSystemURI extensionOf) {
```
apeteri, line 74: Please add an example with any restriction placed on the two parameters. 

If I got it right, the request will upgrade `codeSystem` so that it becomes an extension of `extensionOf`? And so the first argument needs to be a "bare" URI containing only a short name, whereas the second needs to have a version ID appended after the short name.

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/codesystem/CodeSystemUpgradeRequest.java
```diff
@@ -0,0 +1,139 @@
...
+					.setCitation(currentCodeSystem.getCitation())
+					.setExtensionOf(extensionOf)
+					.setUpgradeOf(codeSystem)
+					.setIconPath(currentCodeSystem.getIconPath())
+					.setLocales(currentCodeSystem.getLocales())
+					.setLink(currentCodeSystem.getOrganizationLink())
+					.setName(String.format("Upgrade of '%s' to '%s'", currentCodeSystem.getName(), extensionOf))
+					.setRepositoryId(context.id())
+					.setTerminologyId(currentCodeSystem.getTerminologyId())
+					.build(context.id(), Branch.MAIN_PATH, context.service(User.class).getUsername(), String.format("Start upgrade of '%s' to '%s'", codeSystem, extensionOf))
```
apeteri, line 132: The commit message is a bit misleading, this is the final step 😄

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/codesystem/CodeSystemUpgradeRequest.java
```diff
@@ -0,0 +1,139 @@
...
+		
+		// or the selected extensionOf version is not present as a valid available upgrade
+		if (!availableUpgrades.contains(extensionOf)) {
+			throw new BadRequestException("Upgrades can only be performed to the next available version dependency.")
+				.withDeveloperMessage("Use '%s/<VERSION_ID>', where <VERSION_ID> is one of: '%s'", extensionOf.getCodeSystem(), availableUpgrades);
+		}
+		
+		// auto-generate the codeSystemId if not provided
+		// auto-generated upgrade IDs consist of the original CodeSystem's name and the new extensionOf dependency's URI
+		final String upgradeCodeSystemId;
```
apeteri, line 87: A check to see if the upgrade code system's short name is not already in use would be helpful, as a lot of work will be done before we find out if there was an issue with it.

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/codesystem/CodeSystemUpgradeRequest.java
```diff
@@ -0,0 +1,139 @@
...
+		if (codeSystemId == null) {
+			upgradeCodeSystemId = String.format("%s-%s-UPGRADE", codeSystem.getCodeSystem(), extensionOf.getPath() /*versionId*/);
```
apeteri, line 89: What happens if I upgrade a code system which has already been upgraded once?

apeteri (approved):

core/com.b2international.snowowl.core.rest/src/com/b2international/snowowl/core/rest/codesystem/CodeSystemRestService.java
cmark: Yeah, true, thanks for the suggestion.

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/index/change/ComponentEffectiveTimeRestoreChangeProcessor.java
cmark: Right, we can remove the if block then. 😄

cmark: Unfortunately, no, this might not be true in older systems, we need to keep this for now.

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/codesystem/CodeSystemRequests.java
cmark: Yes, the first URI is the `codeSystemId`, the second is the new dependency version URI.

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/codesystem/CodeSystemUpgradeRequest.java
cmark: Yes, the final step, but from the upgrade point of view, it is the start of the upgrade and the commit marks that point, that's why it has this commit message.

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/codesystem/CodeSystemUpgradeRequest.java
cmark: Branch creation will fail and will report a 409 issue in this case, which is enough for now, more complex checks will be implemented in the next 7.15.0 release.

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/codesystem/CodeSystemUpgradeRequest.java
cmark: An upgrade is essentially a branch, so if you start it again you will get a 409 Conflict status indicating that the branch already exists and you just cannot create another Branch/CodeSystem that represents the upgrade. Theoretically, it is possible to have multiple ongoing upgrades to different versions, but we will consider restricting those in the upcoming 7.15.0 release.

apeteri: I was thinking about whether this expression would create `SNOMEDCT-UK-SNOMEDCT/20200131-UPGRADE-SNOMEDCT/20200731-UPGRADE`-kind of short names if I upgrade to 01-31, then to 07-31.

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/codesystem/CodeSystemUpgradeRequest.java
apeteri: Isn't the original short name used for the new branch? That is different from the short name which will be created below.

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/codesystem/CodeSystemUpgradeRequest.java
```diff
@@ -0,0 +1,139 @@
...
+		String upgradeBranch = RepositoryRequests.branching().prepareCreate()
+			.setParent(parentBranch)
+			.setName(codeSystem.getCodeSystem())
+			.build()
+			.execute(context);
+		
+		// merge branch content from the current code system to the new upgradeBranch
+		Merge merge = RepositoryRequests.merging().prepareCreate()
+			.setSource(currentCodeSystem.getBranchPath())
+			.setTarget(parentBranch)
```
apeteri: Didn't this change to `upgradeBranch` in a commit? I could swear I saw it earlier 😕

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/codesystem/CodeSystemUpgradeRequest.java
apeteri: https://github.com/b2ihealthcare/snow-owl/blob/f2b19be624ddc9b9f26c45102085ea4b0e39a958/core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/codesystem/CodeSystemUpgradeRequest.java#L99-L103

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/codesystem/CodeSystemUpgradeRequest.java
cmark: It did change to that, it looks like that in my dev env.

PR #739, opened 2021-01-17
SO-4227: migrate 6.x donation logic

cmark:


commons/com.b2international.index/src/com/b2international/index/revision/StagingArea.java
```diff
@@ -786,6 +812,10 @@ void merge(RevisionBranchRef fromRef, RevisionBranchRef toRef, boolean squash, R
 			final Map<String, ? extends Revision> oldRevisionsById = FluentIterable.from(oldRevisions).uniqueIndex(Revision::getId);
 			
 			newRevisions.forEach(rev -> {
+				// skip new objects that are already marked as revised on merge source, content that is present on target should take place instead
+				if (externalRevisionsToReviseOnMergeSource.containsEntry(rev.getClass(), rev.getId())) {
```
apeteri: Should there be a similar check for `updatedRevisions` below?

apeteri (approved):

commons/com.b2international.index/src/com/b2international/index/revision/StagingArea.java
cmark: We will see, we can add it earlier, but for now, there is no test case for it.

apeteri (approved):

PR #740, opened 2021-01-18
[classification] Infer grouped additional properties without participating in redundancy tests

apeteri:
Also favor incoming groups over existing groups on a tie, as this makes the accompanying grouped additional properties more consistent, both for groups inferred "from above" as well as the concept's own (stated) groups.


apeteri:
Merged with squash commit so it can be reviewed later.

apeteri:
Reverted; in case of two equal groups that only differ in additional properties, the order of inheritance decides (randomly) whether we get these properties into the final inferred group or not.

It might be worth reinstating in the future, with an extra check for the number of additional properties, letting the group with the larger count win.

PR #743, opened 2021-01-22
SO-4622: Add force recreate version flag

AAAlinaaa:


cmark (changes requested):
This is not what the ticket actually describes. Please read it again and implement this new feature accordingly.

PR #745, opened 2021-01-22
Bump Tycho to 2.2.0

cmark:
Use the latest and greatest Tycho features via 2.2.0.


apeteri (approved):
LGTM! Thank you for looking at the upgrade.

PR #748, opened 2021-01-26
SO-4626: Fix NPE when converting correlation of refset members in SnomedConceptMapSearchRequestEvaluator

nagyo:
https://snowowl.atlassian.net/browse/SO-4626


cmark (approved):
LGTM!

PR #749, opened 2021-01-27
SO-4640: Change Date to String in create version API

cmark:
Change `effectiveTime` property to String type to avoid possible timezone issues on Java API level.
This is an improvement on the existing functionality, everything should behave the same as before.


core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/codesystem/CodeSystemVersionProperties.java
```diff
@@ -47,6 +45,6 @@
 	 * @return the effective date of this code system version (can be {@code null})
 	 */
 	@JsonFormat(shape=Shape.STRING, pattern="yyyyMMdd")
```
apeteri, line 47: Does this annotation do anything with a `String` typed property?

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/codesystem/version/CodeSystemVersionCreateRequest.java
```diff
@@ -92,11 +90,11 @@
 	// local execution variables
 	private transient Multimap<DatastoreLockContext, DatastoreLockTarget> lockTargetsByContext;
 	private transient Map<String, CodeSystem> codeSystemsByShortName;
+	private transient Date effectiveTimeDate;
 	
 	@Override
 	public Boolean execute(ServiceProvider context) {
-		final RemoteJob job = context.service(RemoteJob.class);
-		final String user = job.getUser();
+		final String user = context.service(User.class).getUsername();
```
apeteri, line 96: Do we have the `User` injected for requests that are running in a remote job?

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/codesystem/version/CodeSystemVersionCreateRequest.java
```diff
@@ -221,13 +221,6 @@ private void acquireLocks(ServiceProvider context, String user, Collection<CodeS
 
 				lockTargetsByContext.put(lockContext, lockTarget);
 			}
-			
-		} catch (final OperationLockException e) {
```
apeteri, line 225: Is this handled gracefully somewhere else?

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/codesystem/version/VersioningRequest.java
```diff
@@ -99,7 +99,7 @@ private final CodeSystemVersionEntry createVersion(final TransactionContext cont
 		return CodeSystemVersionEntry.builder()
```
apeteri, line 99: ⚠️ This file also needs a copyright notice update

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/date/EffectiveTimes.java
```diff
@@ -157,7 +156,12 @@ public static final long getEffectiveTime(@Nullable Date effectiveTime) {
 		if (null == effectiveTime) {
 			return UNSET_EFFECTIVE_TIME;
 		}
-		return effectiveTime.getTime();
+		return effectiveTime.toInstant()
+			      .atZone(ZoneId.of("GMT"))
```
apeteri: For consistency:

```suggestion
			      .atZone(ZoneOffset.UTC)
```

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/repository/RepositoryTransactionContext.java
```diff
@@ -277,8 +277,8 @@ public Long commit(String author, String commitComment, String parentLockContext
 		final DatastoreLockTarget lockTarget = createLockTarget(id(), path());
 		IOperationLockManager locks = service(IOperationLockManager.class);
 		Commit commit = null;
+		acquireLock(locks, lockContext, lockTarget);
```
apeteri: I don't remember how the locking goes internally, but if some locks can be granted when another fails, the granted ones should be released. In this case, `acquireLock` should remain inside the `try-finally` block.

apeteri (approved):
👍 I left some questions below.

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/codesystem/CodeSystemVersionProperties.java
cmark: Yes, I'm currently testing this, will let you know once I have something, I hope it does, if not, I'll remove it and add a default value to the API. Also, the `LocalDate` type could also work on the REST API.

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/codesystem/version/CodeSystemVersionCreateRequest.java
cmark: Yes, otherwise this would fail.

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/codesystem/version/CodeSystemVersionCreateRequest.java
cmark: Hmm, this needs to be converted to `LockedException`, I'll add something to make that easier. Or we could throw that exception instead of any low-level ones.

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/date/EffectiveTimes.java
cmark: Actually, this is how it looked like initially, but I intentionally changed it, since one requires a ZoneId and the other requires a ZoneOffset, anyway, the logic will remain the same, so happy to change it back. 😄

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/repository/RepositoryTransactionContext.java
cmark: Hmm, right, then we need to omit/skip locks that cannot be released. 😕

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/codesystem/CodeSystemVersionProperties.java
cmark: Cool, it renders the same example model in Swagger:

```
{
  "description": "string",
  "effectiveDate": "yyyyMMdd",
  "version": "string"
}
```

PR #750, opened 2021-01-28
SO-4293: Add "filter by parent" for generic concept search requests

apeteri:


apeteri:
This is not essential and can be added later.

cmark:
This has been added directly to the `8.x` branch along with the ancestor filter. Closing.

PR #751, opened 2021-01-29
SO-4641: Add moduleId to StatementFragment

cmark:
Add `moduleId` field to StatementFragment so that classifier extensions can include/exclude normal form generated content based on existing module's current moduleId value.


apeteri (approved):
👍

PR #753, opened 2021-02-02
SO-4633: Ensure description inactivator member inactivation is possible

AAAlinaaa:


apeteri:
I think this is a too heavy-handed approach that adds a lot of complexity for little gain. Take a look at how the delete/retire action in the "references" page handles language reference set members: the only thing that needs to be done is setting the inactivation indicator on the referenced component (description in this case) to `null` (this is also needed so that the "descriptions" page updates the status column before the editor is saved).

Edit: OK, there is another change which I should look at; I will add another note over in the other repository.

cmark:
Agree with @apeteri. Let's discuss the alternative solutions first, then decide on which would be the best.

PR #765, opened 2021-02-18
Prevent unnecessary branch delete attempt in case of regular version create requests

cmark:
Add missing else block to delete the branch only and only if the force
flag is enabled.


apeteri (approved):

nagyo (approved):
LGTM!

PR #769, opened 2021-02-22
[core] reduce large history memory consumption by dereferencing...

cmark:
...commit info objects as early as we can


cmark:
This will be resolved globally for all search requests in a new PR for the next minor release.

PR #774, opened 2021-03-18
SO-4670: Ensure language member updates are fully performed following...

AAAlinaaa:
... description retire

### DIFF
--- a/snomed/com.b2international.snowowl.snomed.core.rest.tests/src/com/b2international/snowowl/snomed/core/rest/components/SnomedDescriptionApiTest.java
+++ b/snomed/com.b2international.snowowl.snomed.core.rest.tests/src/com/b2international/snowowl/snomed/core/rest/components/SnomedDescriptionApiTest.java
@@ -193,9 +193,9 @@ public class SnomedDescriptionApiTest extends AbstractSnomedApiTest {
 		String descriptionId = createNewDescription(branchPath);
 		
 		Json requestBody = Json.object(
-				"active", false,
-				"commitComment", "Inactivated description"
-			);
+			"active", false,
+			"commitComment", "Inactivated description"
+		);
 
 		updateComponent(branchPath, SnomedComponentType.DESCRIPTION, descriptionId, requestBody)
 				.statusCode(204);
--- a/snomed/com.b2international.snowowl.snomed.core.rest.tests/src/com/b2international/snowowl/snomed/core/rest/components/SnomedDescriptionApiTest.java
+++ b/snomed/com.b2international.snowowl.snomed.core.rest.tests/src/com/b2international/snowowl/snomed/core/rest/components/SnomedDescriptionApiTest.java
@@ -62,10 +62,7 @@ import com.b2international.snowowl.snomed.cis.domain.IdentifierStatus;
 import com.b2international.snowowl.snomed.cis.domain.SctId;
 import com.b2international.snowowl.snomed.common.SnomedConstants.Concepts;
 import com.b2international.snowowl.snomed.common.SnomedRf2Headers;
-import com.b2international.snowowl.snomed.core.domain.Acceptability;
-import com.b2international.snowowl.snomed.core.domain.AssociationTarget;
-import com.b2international.snowowl.snomed.core.domain.InactivationProperties;
-import com.b2international.snowowl.snomed.core.domain.SnomedDescription;
+import com.b2international.snowowl.snomed.core.domain.*;
 import com.b2international.snowowl.snomed.core.domain.refset.SnomedRefSetType;
 import com.b2international.snowowl.snomed.core.domain.refset.SnomedReferenceSetMember;
 import com.b2international.snowowl.snomed.core.domain.refset.SnomedReferenceSetMembers;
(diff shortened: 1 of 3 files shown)


snomed/com.b2international.snowowl.snomed.core.rest.tests/src/com/b2international/snowowl/snomed/core/rest/components/SnomedDescriptionApiTest.java
```diff
@@ -190,6 +187,26 @@ public void deleteDescription() {
 		getComponent(branchPath, SnomedComponentType.DESCRIPTION, descriptionId)
 			.statusCode(404);
 	}
+	
+	@Test
+	public void retireDescription() {
+		String descriptionId = createNewDescription(branchPath);
+		
+		Json requestBody = Json.object(
+				"active", false,
```
cmark: ⚠️ Indentation!

cmark (approved):
LGTM!

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/index/update/ReferenceSetMembershipUpdater.java
```diff
@@ -79,9 +79,7 @@ private void processReferencingRefSetIds(final Multiset<String> memberOf, final
 			.stream()
 			.filter(c -> c.getChangeKind() == MemberChangeKind.REMOVED)
 			.forEach(change -> {
-				if (change.isActive()) {
-					activeMemberOf.remove(change.getRefSetId());
-				}
```
apeteri, line 84: This `if` block is still required. 

`memberOf` and `activeMemberOf` also keep track of how many members in the same reference set refer to this component, so if there are eg. 5 members in R1 that have this component set as the referenced component, and 3 are active, the two multisets will have entries `R1: 5` and `R1: 3`, respectively.

- When an **addition** is reported, we always increase the counter for the reference set in `memberOf`, but only increase `activeMemberOf` if the new member is active;
- When a **change** in status<sup>1</sup> is reported, we check if the member was inactivated or re-activated, in which case we modify the counter in `activeMemberOf` accordingly, but do not touch `memberOf` at all;
- Finally, when a **removal** is reported, we always decrease the counter for `memberOf`, but should only do the same for `activeMemberOf` if the detached member was active!

<sup>1</sup> We should also check if callers register `MemberChangeKind.CHANGED` items for this reason only!

apeteri (changes requested):

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/index/update/ReferenceSetMembershipUpdater.java
apeteri: The code related to footnote 1 works as expected: https://github.com/b2ihealthcare/snow-owl/blob/92045733c6033c377dd9a2a5fee12cca4eaf347a/snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/index/change/ReferringMemberChangeProcessor.java#L50-L58

cmark: > This `if` block is still required.
> 
> `memberOf` and `activeMemberOf` also keep track of how many members in the same reference set refer to this component, so if there are eg. 5 members in R1 that have this component set as the referenced component, and 3 are active, the two multisets will have entries `R1: 5` and `R1: 3`, respectively.
> 
> * When an **addition** is reported, we always increase the counter for the reference set in `memberOf`, but only increase `activeMemberOf` if the new member is active;
> * When a **change** in status1 is reported, we check if the member was inactivated or re-activated, in which case we modify the counter in `activeMemberOf` accordingly, but do not touch `memberOf` at all;
> * Finally, when a **removal** is reported, we always decrease the counter for `memberOf`, but should only do the same for `activeMemberOf` if the detached member was active!
> 
> 1 We should also check if callers register `MemberChangeKind.CHANGED` items for this reason only!

Right, then @AAAlinaaa please revert this change.

snomed/com.b2international.snowowl.snomed.core.rest.tests/src/com/b2international/snowowl/snomed/core/rest/components/SnomedDescriptionApiTest.java
```diff
@@ -190,6 +187,26 @@ public void deleteDescription() {
...
+		updateComponent(branchPath, SnomedComponentType.DESCRIPTION, descriptionId, requestBody)
+				.statusCode(204);
+		
+		SnomedDescription description = getComponent(branchPath, SnomedComponentType.DESCRIPTION, descriptionId, "acceptabilities()")
+			.statusCode(200)
+			.extract().as(SnomedDescription.class);
+		
+		assertThat(!description.isActive());
+		assertThat(description.getAcceptabilities().isEmpty());
+	}
```
apeteri, line 209: This test is passing without any other changes on 7.x. I suspect something on the client side, as the supposedly retired description appears in `SnomedConceptUpdateRequest` with a non-empty acceptability map. 🔍

cmark: Yes, I told exactly this to Alina a few hours ago 😄

AAAlinaaa: As mentioned earlier, those acceptability values are filled based on the values returned by the server.

cmark:
Closing in favor of #776.

PR #778, opened 2021-03-19
Exclude language refset members from inactivation change processing

AAAlinaaa:
... change processing


AAAlinaaa:
Closed as there is no need for selective language refset member inactivation.

PR #783, opened 2021-03-24
SO-4737 reindex on alias field changes

cmark:
This PR adds the capability to migrate existing indices to a new mapping schema when the new schema has field alias changes (and additionally extra new fields). 

The following schema changes are still considered as breaking and require a major Snow Owl version bump:
- deleting a field from a mapping
- renaming a field on a mapping


commons/com.b2international.index/src/com/b2international/index/es/admin/EsIndexAdmin.java
```diff
@@ -391,7 +412,9 @@ private void waitForYellowHealth(String... indices) {
 						if (!keywordMapping.index()) {
 							prop.put("index", false);
 						}
-						prop.put("doc_values", keywordMapping.index());
+						if (!keywordMapping.index()) {
+							prop.put("doc_values", keywordMapping.index());
+						}
```
apeteri: The two `if (!keywordMapping.index())` tests could be merged together; the property value could also be converted to a `false` literal for consistency.

apeteri (approved):
LGTM!

PR #788, opened 2021-03-28
[taxonomy] Update OWL ontology building process

apeteri:
This PR changes `DelegateOntology` so that the `expression` content of each OWL axiom reference set member is loaded directly, without splitting them into stated relationship fragments.


nagyo (changes requested):
Further investigation is needed due to higher amount of classification results after applying the proposed changes

apeteri:
Closing in favor of a future PR that incorporates these changes.

PR #790, opened 2021-04-03
SO-4751: support nested object and array/collection property auto merging

cmark:
This PR introduces automatic merging of non-conflicting nested objects, arrays, and collections during rebase/merge operations.

### DIFF
--- a/commons/com.b2international.index/src/com/b2international/index/es/admin/EsIndexAdmin.java
+++ b/commons/com.b2international.index/src/com/b2international/index/es/admin/EsIndexAdmin.java
@@ -54,7 +54,6 @@ import org.elasticsearch.script.ScriptType;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import com.b2international.commons.ClassUtils;
 import com.b2international.commons.CompareUtils;
 import com.b2international.commons.ReflectionUtils;
 import com.b2international.index.*;
@@ -65,16 +64,12 @@ import com.b2international.index.mapping.DocumentMapping;
 import com.b2international.index.mapping.Mappings;
 import com.b2international.index.query.Expression;
 import com.b2international.index.query.Expressions;
+import com.b2international.index.util.JsonDiff;
 import com.b2international.index.util.NumericClassUtils;
 import com.fasterxml.jackson.annotation.JsonValue;
-import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
-import com.fasterxml.jackson.databind.node.ArrayNode;
 import com.fasterxml.jackson.databind.node.ObjectNode;
-import com.flipkart.zjsonpatch.DiffFlags;
-import com.flipkart.zjsonpatch.JsonDiff;
 import com.google.common.collect.ImmutableMap;
-import com.google.common.collect.Iterables;
 import com.google.common.collect.Sets;
 import com.google.common.primitives.Primitives;
 
@@ -98,12 +93,6 @@ public final class EsIndexAdmin implements IndexAdmin {
 		IndexClientFactory.COMMIT_WATERMARK_LOW_KEY,
 		IndexClientFactory.COMMIT_WATERMARK_HIGH_KEY
 	);
-	
-	/**
-	 * Important DIFF flags required to produce the JSON patch needed for proper compare and branch merge operation behavior, for proper index schema migration, and so on.
-	 * Should be used by default when using jsonpatch.
-	 */
-	public static final EnumSet<DiffFlags> DIFF_FLAGS = EnumSet.of(DiffFlags.ADD_ORIGINAL_VALUE_ON_REPLACE, DiffFlags.OMIT_COPY_OPERATION, DiffFlags.OMIT_MOVE_OPERATION);
 	
 	private static final int DEFAULT_MAX_NUMBER_OF_VERSION_CONFLICT_RETRIES = 5;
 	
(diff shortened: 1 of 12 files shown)


commons/com.b2international.index/src/com/b2international/index/es/admin/EsIndexAdmin.java
```diff
@@ -192,24 +181,15 @@ public void create() {
...
-						case "replace":
-							uncompatibleChanges.add(prop);
-							break;
-						default:
-							break;
+					JsonDiff.diff(currentTypeMapping, newTypeMapping).forEach(change -> {
+						if (change.isAdd()) {
+							compatibleChanges.add(change.getFieldPath());
+						} else if (change.isMove() || change.isReplace()) {
+							uncompatibleChanges.add(change.getFieldPath());
```
apeteri: 👓 No idea why, but the prefix for this word is usually _in-_ instead of _un-_. The entry in Wiktionary says this form is "not conforming to the language as accepted by its majority of speakers": https://en.wiktionary.org/wiki/uncompatible

commons/com.b2international.index/src/com/b2international/index/revision/RevisionConflictProcessor.java
```diff
@@ -91,12 +105,104 @@ default Conflict convertConflict(Conflict conflict) {
...
+		public RevisionPropertyDiff handleChangedInSourceAndTarget(String revisionId, DocumentMapping mapping, RevisionPropertyDiff sourceChange, RevisionPropertyDiff targetChange, ObjectMapper mapper) {
+			String property = sourceChange.getProperty();
+			if (mapping.isObject(property)) {
+				// in case of Object properties, allow changes touching different properties to be merged without reporting conflict 
+				return handleObjectConflict(mapping, sourceChange, targetChange, mapper);
+			} else if (mapping.isCollection(property)) {
+				// in case of Collection/Array properties, allow subsets to be merged together 
+				// eg. [1,2] -> [1,2,3] vs. [1,2] -> [1,2,4] should not produce conflicts, but merge the array into [1,2,3,4]
+				return handleCollectionConflict(mapping, sourceChange, targetChange, mapper, mapping.isSet(property));
+			} else if (Objects.equals(sourceChange.getNewValue(), targetChange.getNewValue())) {
```
apeteri, line 120: Isn't this check universal enough that it also applies to objects and collections? It could be moved to the top before asking the document mapping to return the property's type.

commons/com.b2international.index/src/com/b2international/index/revision/RevisionConflictProcessor.java
```diff
@@ -91,12 +105,104 @@ default Conflict convertConflict(Conflict conflict) {
...
 				return sourceChange;
 			} else {
-				return null; 
+				return null;
+			}
+		}
+
+		private RevisionPropertyDiff handleObjectConflict(DocumentMapping mapping, RevisionPropertyDiff sourceChange, RevisionPropertyDiff targetChange, ObjectMapper mapper) {
+			try {
+				ObjectNode oldObject = CompareUtils.isEmpty(sourceChange.getOldValue()) ? mapper.createObjectNode() : (ObjectNode) mapper.readTree(sourceChange.getOldValue());
```
apeteri, line 130: Is `oldObject` loaded from the same point in time (most recent revision visible from both source and target branches)?

commons/com.b2international.index/src/com/b2international/index/revision/RevisionConflictProcessor.java
```diff
@@ -91,12 +105,104 @@ default Conflict convertConflict(Conflict conflict) {
...
+				ArrayDiff sourceDiff = ArrayDiff.diff(oldArray, sourceArray);
+				ArrayDiff targetDiff = ArrayDiff.diff(oldArray, targetArray);
+				
+				// TODO nested item conflict resolution???
+				// create a big union of all changes, treat collection as set if needed
+				// this eliminates most of the unnecessary conflicts between two collection properties
+				sourceDiff.getAddedItems().forEach(oldArray::add);
+				Set<JsonNode> newSourceItems = Set.copyOf(sourceDiff.getAddedItems());
+				targetDiff.getAddedItems().forEach(newTargetItem -> {
+					if (!isSet || !newSourceItems.contains(newTargetItem)) {
```
apeteri: Maybe I'm overly cautious, but to be _really_ sure that we uphold `Set` semantics, `oldArray` could be checked here instead of `newSourceItems`.

```suggestion
					if (!isSet || !oldArray.contains(newTargetItem)) {
```

commons/com.b2international.index/src/com/b2international/index/revision/StagingArea.java
```diff
@@ -452,15 +449,25 @@ private Commit doCommit(String commitGroupId, long timestamp, String author, Str
...
+				
+				final String prop = change.getFieldPath();
+				final String value = change.serializeValue();
+				if (change.isRemove()) { // fully clear property
+					// split by object type
+					objectIdsByType.keySet().forEach(type -> {
+						details.add(CommitDetail.changedProperty(prop, value, "", type, objectIdsByType.get(type)));
+					});
+				} else if (change.isAdd() || change.isReplace()) { // set or replace value
+					final String from = change.serializeFromValue();
```
apeteri: 📎 It looks like this code change could be avoided if `serializeFromValue` would return `""` for removed properties.

apeteri (approved):

commons/com.b2international.index/src/com/b2international/index/revision/RevisionConflictProcessor.java
cmark: No, it's not. The `getNewValue()` method returns the String representation of the value. In case of simple types like primitives and Strings this works fine, but when it comes to arrays/objects it needs extra care if the array/object JSON node list changes, but semantically the same (eg. order change). We don't want to report a conflict when an array changes from [1, 2] -> [2, 1] or when an object property list order changes, but holds the same values.

commons/com.b2international.index/src/com/b2international/index/revision/RevisionConflictProcessor.java
cmark: Yes, see here https://github.com/b2ihealthcare/snow-owl/blob/7.x/commons/com.b2international.index/src/com/b2international/index/revision/RevisionCompareDetail.java#L188

commons/com.b2international.index/src/com/b2international/index/revision/StagingArea.java
cmark: Please note: the order of arguments is different in the two branches, `from`, `value` vs. `value`, `from`.

commons/com.b2international.index/src/com/b2international/index/revision/RevisionConflictProcessor.java
cmark: Not sure, `oldArray` is an ArrayNode and that does not have a contains method. Also, `sourceDiff` items need to be checked as well if we are overly cautious.

commons/com.b2international.index/src/com/b2international/index/revision/RevisionConflictProcessor.java
```diff
@@ -184,13 +185,13 @@ protected RevisionPropertyDiff handleCollectionConflict(DocumentMapping mapping,
...
-				Set<JsonNode> newSourceItems = Set.copyOf(sourceDiff.getAddedItems());
-				targetDiff.getAddedItems().forEach(newTargetItem -> {
-					if (!isSet || !newSourceItems.contains(newTargetItem)) {
-						oldArray.add(newTargetItem);
-					}
-				});
+				Set<JsonNode> oldArraySet = Sets.newHashSet(oldArray);
+				Streams.concat(sourceDiff.getAddedItems().stream(), targetDiff.getAddedItems().stream())
+					.forEach(newItem -> {
+						if (!isSet || !oldArraySet.contains(newItem)) {
```
apeteri: Using `oldArraySet.add(newItem)` (no negation, returns `true` if the set changed) instead of `!oldArraySet.contains(newItem)` would both update the backing `Set` and check if `newItem` is a duplicate.

cmark: True, updated, thanks!

PR #791, opened 2021-04-07
SO-4751: report conflict if the an item added on both branches...

cmark:
...with the same ID but with different contents (same content is allowed
and merged automatically).


apeteri (approved):
👍

PR #794, opened 2021-04-19
SO-4728: extract ql packages

cmark:
This PR moves the SNOMED CT Query Language implementation modules to an external repository: https://github.com/b2ihealthcare/snomed-ql


apeteri (approved):
LGTM!

PR #802, opened 2021-05-03
SO-4800: RF2 import incorrect release flag

molnarlaura:


snomed/com.b2international.snowowl.snomed.core.rest.tests/src/com/b2international/snowowl/snomed/core/rest/io/SnomedImportApiTest.java
```diff
@@ -448,6 +445,38 @@ public void import28ImportConceptAsInactive() {
...
+		importArchive("SnomedCT_Release_INT_20210502_concept_w_eff_time.zip");
+		SnomedConcept conceptBefore = getComponent(branchPath, SnomedComponentType.CONCEPT, "100005").statusCode(200).extract().as(SnomedConcept.class);
+		
+		importArchive("SnomedCT_Release_INT_20210502_concept_wo_eff_time.zip");
+		SnomedConcept conceptAfter = getComponent(branchPath, SnomedComponentType.CONCEPT, "100005").statusCode(200).extract().as(SnomedConcept.class);
+		
+		assertEquals(EffectiveTimes.parse("2021-05-02"), conceptBefore.getEffectiveTime());
+		assertTrue(conceptBefore.isReleased());
+		assertEquals(EffectiveTimes.toDate(EffectiveTimes.UNSET_EFFECTIVE_TIME), conceptAfter.getEffectiveTime());
+		assertFalse(conceptAfter.isReleased());
```
apeteri: This doesn't seem right. When a component has been released, it should stay released &ndash; it is a one-way switch to indicate that a downstream consumer of terminology content might have already "seen" the component as part of an RF2 export, and authors should/may not delete it completely (and later, possibly, recycle the ID for a completely different concept).

apeteri (changes requested):

snomed/com.b2international.snowowl.snomed.core.rest.tests/src/com/b2international/snowowl/snomed/core/rest/io/SnomedImportApiTest.java
molnarlaura: Changed the test case to check that the component stays released if it was released before import. And changed the import logic based on this as well.

nagyo (approved):
LGTM!

PR #807, opened 2021-05-06
fix(index): use segment diff to compute accurate base/head timestamps...

cmark:
...when merging empty branch (with non-empty parent branch segments) to
another branch. Related use case: use CodeSystem version branch to
create a Code System upgrade instead of using the HEAD of the
CodeSystem's working branch.


apeteri (approved):
Non-review approval

PR #815, opened 2021-05-25
SO-4809: Make inclusion of non-property type changes optional during ...

AAAlinaaa:
... compare, exclude them from branch compare results, count total
commit details in new compare result content

### DIFF
--- a/commons/com.b2international.index/src/com/b2international/index/revision/DefaultRevisionIndex.java
+++ b/commons/com.b2international.index/src/com/b2international/index/revision/DefaultRevisionIndex.java
@@ -142,27 +142,27 @@ public final class DefaultRevisionIndex implements InternalRevisionIndex, Hooks 
 	
 	@Override
 	public RevisionCompare compare(final String branch, final int limit) {
-		return compare(getBaseRef(branch), getBranchRef(branch), limit, true);
+		return compare(getBaseRef(branch), getBranchRef(branch), limit, false);
 	}
 	
 	@Override
 	public RevisionCompare compare(final String baseBranch, final String compareBranch) {
-		return compare(baseBranch, compareBranch, COMPARE_DEFAULT_LIMIT, true);
+		return compare(baseBranch, compareBranch, COMPARE_DEFAULT_LIMIT, false);
 	}
 	
 	@Override
-	public RevisionCompare compare(final String baseBranch, final String compareBranch, final int limit, boolean preserveComponentChanges) {
-		return compare(getBranchRef(baseBranch), getBranchRef(compareBranch), limit, preserveComponentChanges);
+	public RevisionCompare compare(final String baseBranch, final String compareBranch, final int limit, boolean excludeComponentChanges) {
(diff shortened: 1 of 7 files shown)


commons/com.b2international.index/src/com/b2international/index/revision/DefaultRevisionIndex.java
```diff
@@ -142,27 +142,27 @@ public RevisionCompare compare(final String branch) {
...
 	@Override
 	public RevisionCompare compare(final String baseBranch, final String compareBranch) {
-		return compare(baseBranch, compareBranch, COMPARE_DEFAULT_LIMIT);
+		return compare(baseBranch, compareBranch, COMPARE_DEFAULT_LIMIT, true);
 	}
 	
 	@Override
-	public RevisionCompare compare(final String baseBranch, final String compareBranch, final int limit) {
-		return compare(getBranchRef(baseBranch), getBranchRef(compareBranch), limit);
+	public RevisionCompare compare(final String baseBranch, final String compareBranch, final int limit, boolean preserveComponentChanges) {
```
cmark: I think a better configuration option would be the opposite, `excludeComponentChanges`. The default functionality is to always include them, so allowing optional exclusion of them is the functionality we need. Please rename.

cmark: Alternatively, one could define the set of Operation literal to exclude.

commons/com.b2international.index/src/com/b2international/index/revision/RevisionCompare.java
```diff
@@ -138,14 +144,16 @@ public RevisionCompare build() {
 	private final int totalAdded;
 	private final int totalChanged;
 	private final int totalRemoved;
+	private final int total;
```
cmark: What is total here? Total `added` + `changed` + `removed`?

cmark (changes requested):

commons/com.b2international.index/src/com/b2international/index/revision/RevisionCompare.java
AAAlinaaa: No, it is the total, possibly more than the sum of the three if it is more than the limit.

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/branch/compare/BranchCompareResult.java
```diff
@@ -79,7 +79,7 @@ public Builder totalDeleted(int totalDeleted) {
 			this.totalDeleted = totalDeleted;
 			return this;
 		}
-		
+				
 		public BranchCompareResult build() {
```
cmark, line 83: I think you can safely remove the `build(Set changedContainers)` method and keep the zero-arg build method as the one that actually builds the result.

cmark (changes requested):

cmark (approved):
LGTM!

PR #823, opened 2021-06-03
SO-4780: Test that import succeeds for member with valid reference ...

AAAlinaaa:
... in the same refset as member with invalid reference


cmark (approved):
LGTM!

PR #825, opened 2021-06-04
SO-3927: Ensure component changes are not enumerated when the exclude...

AAAlinaaa:
... component changes flag is set to true


cmark (approved):
LGTM!

PR #827, opened 2021-06-04
Add case sensitivity flag to term filter

kovacstomi:


core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/request/TermFilter.java
```diff
@@ -47,6 +48,7 @@ public TermFilter(final String term, final Integer minShouldMatch, final boolean
 		this.exact = exact;
 		this.parsed = parsed;
 		this.ignoreStopwords = ignoreStopwords;
+		this.isCaseSensitive = isCaseSensitive;
```
cmark: Please rename this field to `caseSensitive`. `is` prefix is only applicable when you create an accessor for the field.

cmark (changes requested):

cmark (approved):
LGTM!

cmark (approved):
LGTM!

PR #832, opened 2021-06-14
Add bundle resource

kovacstomi:


core/com.b2international.snowowl.core.rest.tests/src/com/b2international/snowowl/core/bundle/BaseBundleApiTest.java
```diff
@@ -0,0 +1,130 @@
...
+import com.b2international.snowowl.test.commons.TestMethodNameRule;
+import com.b2international.snowowl.test.commons.rest.RestExtensions;
+
+/**
+ * @since 8.0
+ */
+abstract class BaseBundleApiTest {
+	
+	static final String USER = RestExtensions.USER;
+	static final String ROOT = "-1";
```
cmark: Use `IComponent.ROOT_ID`.

core/com.b2international.snowowl.core.rest.tests/src/com/b2international/snowowl/core/bundle/BaseBundleApiTest.java
```diff
@@ -0,0 +1,130 @@
...
+	@After
+	public void cleanUp() {
+		BundleRequests.prepareSearchBundle()
+			.buildAsync()
+			.getRequest()
+			.execute(Services.context())
+			.getItems()
+			.forEach(bundle -> {
+				ResourceRequests.prepareDelete(bundle.getId())
+					.build(USAGE, String.format("Delet bundle: %s", bundle.getId()))
```
cmark: ⚠️ Typo!

core/com.b2international.snowowl.core.rest.tests/src/com/b2international/snowowl/core/bundle/BundleApiTest.java
```diff
@@ -0,0 +1,383 @@
...
+ * @since 8.0
+ */
+public final class BundleApiTest extends BaseBundleApiTest {
+	
+	@Test
+	public void createWithoutId() {
+		final String id = BundleRequests.prepareNewBundle()
+				.setUrl(URL)
+				.setTitle(TITLE)
+				.build(USER, String.format("Create bundle"))
```
cmark: No format argument, either add one or remove the `String.format` call.

core/com.b2international.snowowl.core.rest.tests/src/com/b2international/snowowl/core/bundle/BundleApiTest.java
```diff
@@ -0,0 +1,383 @@
...
+		assertThat(bundle.getId()).isEqualTo(id);
+	}
+
+	@Test(expected = BadRequestException.class)
+	public void createWithoutNullId() {
+		BundleRequests.prepareNewBundle()
+			.setId(null)
+			.setUrl(URL)
+			.setTitle(TITLE)
+			.build(USER, String.format("Create bundle"));
```
cmark: Same here.

core/com.b2international.snowowl.core.rest.tests/src/com/b2international/snowowl/core/bundle/BundleApiTest.java
```diff
@@ -0,0 +1,383 @@
...
+	}
+	
+	@Test(expected = BadRequestException.class)
+	public void createWithoutTitle() {
+		BundleRequests.prepareNewBundle()
+			.setId(id)
+			.setUrl(URL)
+			.build(USER, String.format("Create bundle"));
```
cmark: And here.

core/com.b2international.snowowl.core.rest.tests/src/com/b2international/snowowl/core/bundle/BundleApiTest.java
```diff
@@ -0,0 +1,383 @@
...
+	}
+	
+	@Test(expected = BadRequestException.class)
+	public void createWithoutUrl() {
+		BundleRequests.prepareNewBundle()
+			.setId(id)
+			.setTitle(TITLE)
+			.build(USER, String.format("Create bundle"));
```
cmark: And here.

core/com.b2international.snowowl.core.rest.tests/src/com/b2international/snowowl/core/bundle/BundleApiTest.java
```diff
@@ -0,0 +1,383 @@
...
+	}
+
+	@Test
+	public void searchById() {
+		final String id2 = "id_2";
+		
+		createBundle();
+		createBundle(id2);
+		
+		final List<String> bundleIds = build(BundleRequests.prepareSearchBundle().filterById(id))
```
cmark: You can assert with `assertThat` on `Iterable` and `Stream` instances as well, so you can make this a bit nicer like this:

```
assertThat(build(BundleRequests.prepareSearchBundle().filterById(id)))
     .hasSameElementsAs(List.of(id));
```

cmark: The same applies to all test methods in this test class.

core/com.b2international.snowowl.core.rest.tests/src/com/b2international/snowowl/core/rest/bundle/BundleRestApiTest.java
```diff
@@ -0,0 +1,120 @@
...
+						ResourceDocument.Fields.ID, "b1",
+						ResourceDocument.Fields.TITLE, "Bundle title b2"))
+		.statusCode(400).body("violations", hasItem("'url' may not be empty (was 'null')"));
+	}
+	
+	@Test
+	public void createBundle() throws JsonProcessingException {
+		final Json body = prepareCreateRequestBody("b1");
+		assertBundleCreated(body);
+		assertBundleGet("b1").and()
```
cmark: The first assert should be a `statusCode` assert in any REST API test. Please add it to here and all other methods.

core/com.b2international.snowowl.core.rest.tests/src/com/b2international/snowowl/core/rest/codesystem/CodeSystemApiTest.java
```diff
@@ -363,6 +363,19 @@ public void codesystem20_UpdateInvalidBranchPath() {
 		assertCodeSystemUpdatedWithStatus(codeSystemId, updateRequestBody, 400);
 	}
 	
+	@Test
+	public void codesystem21_UpdateBundleId() {
```
cmark: The number in the test case is already taken, please use a new one and move the method to its correct place in this class.

core/com.b2international.snowowl.core.rest/src/com/b2international/snowowl/core/rest/bundle/BundleRestService.java
```diff
@@ -0,0 +1,217 @@
...
+import io.swagger.annotations.Api;
+import io.swagger.annotations.ApiOperation;
+import io.swagger.annotations.ApiParam;
+import io.swagger.annotations.ApiResponse;
+import io.swagger.annotations.ApiResponses;
+
+/**
+ * @since 8.0
+ */
+@Api(value = "Bundles", tags = { "resources" })
```
cmark: This should be part of the Resources API collection as well, please use `Resources` in the value annotation property.

core/com.b2international.snowowl.core.rest/src/com/b2international/snowowl/core/rest/bundle/BundleRestService.java
```diff
@@ -0,0 +1,217 @@
...
+			@ApiParam(value="The bundle identifier")
+			@PathVariable(value="bundleId") 
+			final String bundleId,
+			
+			@RequestBody
+			final Bundle bundle,
+			
+			@RequestHeader(value = X_AUTHOR, required = false)
+			final String author) {
+		final String commitComment = String.format("Updated bunle %s", bundleId);
```
cmark: Typo! And use `Update` instead of `Updated`.

core/com.b2international.snowowl.core.rest/src/com/b2international/snowowl/core/rest/bundle/BundleRestService.java
```diff
@@ -0,0 +1,217 @@
...
+				.execute(getBus());
+	}
+	
+	@ApiOperation(
+			value="Create a bundle",
+			notes="Create a new bundle with the given parameters"
+		)
+		@ApiResponses({
+			@ApiResponse(code = 201, message = "Created", response = Void.class),
+			@ApiResponse(code = 400, message = "Bundle already exists in the system", response = RestApiError.class)
```
cmark: Already exists results in `409` error, not `400`, please update.
Also, add the missing `400` ApiResponse, which is for Bad Requests.

core/com.b2international.snowowl.core.rest/src/com/b2international/snowowl/core/rest/bundle/BundleRestService.java
```diff
@@ -0,0 +1,217 @@
...
+				.execute(getBus())
+				.getSync(COMMIT_TIMEOUT, TimeUnit.MINUTES);
+	}
+	
+	@ApiOperation(
+			value="Delete a bundle",
+			notes="Delete a bundle with the given parameters")
+	@ApiResponses({
+		@ApiResponse(code = 204, message = "Deletion successful", response = Void.class),
+		@ApiResponse(code = 400, message = "Bundle cannot be deleted", response = RestApiError.class)
```
cmark: Should be `409` if a conflict occurs and the resource cannot be deleted.

core/com.b2international.snowowl.core.rest/src/com/b2international/snowowl/core/rest/bundle/BundleRestService.java
```diff
@@ -0,0 +1,217 @@
...
+			
+			return ResponseEntity.created(getResourceLocationURI(codeSystemId)).build();
+		}
+	
+	@ApiOperation(
+			value="Update a bundle",
+			notes="Update a bundle with the given parameters")
+	@ApiResponses({
+		@ApiResponse(code = 204, message = "No content", response = Void.class),
+		@ApiResponse(code = 400, message = "Bundle cannot be updated", response = RestApiError.class)
```
cmark:
```suggestion
		@ApiResponse(code = 400, message = "Bad Request", response = RestApiError.class)
```

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/bundle/BundleConverter.java
```diff
@@ -0,0 +1,87 @@
...
+	}
+
+	private void expandContents(final List<Bundle> results) {
+		if (!expand().containsKey(Bundle.Expand.RESOURCES)) {
+			return;
+		}
+		
+		results.forEach(result -> {
+			final Resources resources = ResourceRequests.prepareSearch()
+				.setLimit(100)
```
cmark: Use `limit` from the expand parameter instead of hardcoding it.

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/bundle/BundleDeleteRequest.java
```diff
@@ -0,0 +1,111 @@
...
+	 * 	|_ cs3
+	 * </blockquote></pre>
+	 * @param context 
+	 */
+	private void updateResourceBundles(TransactionContext context, final String id, final String bundleId) {
+		new SearchResourceRequestIterator<>(
+				ResourceRequests.prepareSearch()
+					.setLimit(10_000)
+					.filterByBundleId(id),
+				req -> req.buildAsync().getRequest().execute(context)
```
cmark: Why async? This should be executed synchronously.

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/bundle/BundleDeleteRequest.java
```diff
@@ -0,0 +1,111 @@
...
+			).forEachRemaining(resourceBatch -> {
+				final BulkRequestBuilder<ServiceProvider> bulkRequestBuilder = BulkRequest.<ServiceProvider>create();
+
+				resourceBatch.stream().forEach(resource -> bulkRequestBuilder.add(toRequest(context, resource, bundleId)));
+				
+				bulkRequestBuilder.build().execute(context);
```
cmark: This will create a bunch of unnecessary small commits if a Bundle has thousands of resources associated. It would be much better to add all child resource bundle ID change with the Delete bundle commit, so it can be tracked and retrieved in history APIs, etc.

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/bundle/BundleRequests.java
```diff
@@ -0,0 +1,44 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.b2international.snowowl.core.bundle;
+
+/**
+ * @since 8.0
+ */
+public final class BundleRequests {
```
cmark, line 21: Convert this to an instantiatable class and add a new method `ResourceRequests.bundles()` that creates it, so you can have an API like this:
```
ResourceRequests.bundles().prepareCreate()...
ResourceRequests.bundles().prepareSearch()...
ResourceRequests.bundles().prepareGet(id)...
ResourceRequests.bundles().prepareDelete(id)...
ResourceRequests.bundles().prepareUpdate(id)...
```

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/bundle/BundleSearchRequest.java
```diff
@@ -0,0 +1,104 @@
...
+	implements RepositoryAccessControl {
+
+	private static final long serialVersionUID = 1L;
+
+	/**
+	 * @since 8.0
+	 */
+	public enum OptionKey {
+		/** Search bundles by title */
+		TITLE,
```
cmark: It would be great to introduce an abstract base class for all resource search requests.

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/request/BaseResourceCreateRequest.java
```diff
@@ -0,0 +1,99 @@
...
+import com.b2international.snowowl.core.events.Request;
+
+/**
+ * @since 8.0
+ */
+public abstract class BaseResourceCreateRequest implements Request<TransactionContext, String> {
+
+	protected static final long serialVersionUID = 1L;
+
+	// the new codesystem's ID, if not specified, it will be auto-generated
```
cmark: Remove CodeSystem words from abstract resource base classes.

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/request/BaseResourceCreateRequest.java
```diff
@@ -0,0 +1,99 @@
...
+	@NotEmpty
+	protected String id;
```
cmark: Protected fields are considered a code smell (at least in larger class hierarchies). If possible, try to avoid them. You can use an abstract method to create a concrete builder instance then call it from here and allow specialization in another method.

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/request/BaseResourceCreateRequestBuilder.java
```diff
@@ -0,0 +1,126 @@
...
+
+/**
+ * @since 8.0
+ */
+public abstract class BaseResourceCreateRequestBuilder <RB extends BaseResourceCreateRequestBuilder<RB, R>, R extends BaseResourceCreateRequest>
+		extends BaseRequestBuilder<RB, TransactionContext, String> 
+		implements ResourceRepositoryTransactionRequestBuilder<String> {
+	
+	private String id = IDs.randomBase64UUID();
+	private String bundleId = "-1";
```
cmark: Use `IComponent.ROOT_ID`.

cmark (changes requested):

core/com.b2international.snowowl.core.rest.tests/src/com/b2international/snowowl/core/bundle/BaseBundleApiTest.java
```diff
@@ -127,4 +129,12 @@ Bundles build(final BundleSearchRequestBuilder builder) {
 				.getRequest()
 				.execute(Services.context());
 	}
+	
+	Stream<String> buildAsIds(final BundleSearchRequestBuilder builder) {
```
cmark: Let's rename this method to reflect its purpose better. How about `executeThenExtractIds`?

core/com.b2international.snowowl.core.rest.tests/src/com/b2international/snowowl/core/rest/BunldeApiAssert.java
```diff
@@ -41,13 +42,15 @@ public static ValidatableResponse assertBundleSearch() {
...
+				.then().statusCode(200).assertThat();
 	}
 	
 	public static ValidatableResponse assertBundleGet(final String resourceId) {
 		return givenAuthenticatedRequest(BUNDLE_API)
 				.when().get("/{id}", resourceId)
-				.then().assertThat();
+				.then().assertThat().statusCode(200)
+				.and().body("id", equalTo(resourceId))
+				.assertThat();
```
cmark: Use `and()` to allow the caller to attach more assertions to the chain.

core/com.b2international.snowowl.core.rest.tests/src/com/b2international/snowowl/core/rest/BunldeApiAssert.java
```diff
@@ -68,8 +71,25 @@ public static ValidatableResponse assertCreate(final Map<String, Object> request
...
+			.then().assertThat()
+			.statusCode(204);
+	}
+	
+	public static ValidatableResponse assertUpdateBundleField(final String uniqueId, final String field, final String value) {
+		updateBundle(uniqueId, Map.of(field, value));
+		
+		return assertBundleGet(uniqueId).and()
+			.body(field, equalTo(value))
+			.assertThat();
```
cmark: Same here.

core/com.b2international.snowowl.core.rest.tests/src/com/b2international/snowowl/core/rest/CodeSystemApiAssert.java
```diff
@@ -118,7 +118,8 @@ public static Json prepareCodeSystemCreateRequestBody(final String codeSystemId,
 			"toolingId", TOOLING_ID,
 			"oid", codeSystemId,
 			"language", "en",
-			"url", "link"
+			"url", "link",
+			"bundleId", IComponent.ROOT_ID
```
cmark: Is this required here? The default should be `ROOT`.

core/com.b2international.snowowl.core.rest.tests/src/com/b2international/snowowl/core/rest/bundle/BundleRestApiTest.java
```diff
@@ -117,4 +118,100 @@ public void serachBundleByTitle() {
...
+		final String id = "b12";
+		assertBundleCreated(prepareCreateRequestBody(id));
+		assertUpdateBundleField(id, "usage", "For testing");
+	}
+	
+	@Test
+	public void updateBundlePurpose() {
+		final String id = "b13";
+		assertBundleCreated(prepareCreateRequestBody(id));
+		assertUpdateBundleField(id, "purpose", "Test bundle REST API update endpoint");
```
cmark, line 208: There are no assertions in these tests, as minimum REST API tests should always explicitly check for status codes returned in the response. Could you please add them? Thanks!

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/bundle/BundleRequests.java
```diff
@@ -18,27 +18,25 @@
...
 		return new BundleCreateRequestBuilder();
 	}
 
-	public static BundleGetRequestBuilder prepareGetBundle(final String id) {
+	public static BundleGetRequestBuilder prepareGet(final String id) {
 		return new BundleGetRequestBuilder(id);
 	}
 	
-	public static BundleSearchRequestBuilder prepareSearchBundle() {
+	public static BundleSearchRequestBuilder prepareSearch() {
```
cmark: These are still static resources and you are referring to them like that. Could you please turn them into member methods and use them like this `ResourceRequests.bundles().prepareSearch()`.

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/bundle/BundleRequests.java
```diff
@@ -18,27 +18,25 @@
 /**
  * @since 8.0
  */
-public final class BundleRequests {
+public class BundleRequests {
```
cmark: You can keep the final keyword, classes like this are not meant to be subclassed.

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/codesystem/CodeSystemCreateRequest.java
```diff
@@ -101,7 +109,7 @@ private void checkBranchPath(final RepositoryContext context, final boolean crea
 		
 		// If the branch should be created, it branch should not exist, however 
 		if (create) {
-			final String newBranchPath = Branch.get(parentPath, title);
+			final String newBranchPath = Branch.get(parentPath, getTitle());
```
cmark: Title is definitely wrong here, it should be ID. Tests should break unless you use the same value for both.

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/codesystem/CodeSystemCreateRequest.java
```diff
@@ -112,25 +120,25 @@ private void checkBranchPath(final RepositoryContext context, final boolean crea
...
 		if (existingId) {
-			throw new AlreadyExistsException("Resource", id);
+			throw new AlreadyExistsException("Resource", getId());
 		}
 		
 		// title should be unique across all resources
 		final boolean existingTitle = ResourceRequests.prepareSearch()
 			.setLimit(0)
-			.filterByTitleExact(title)
+			.filterByExactTermIgnoreCase(getTitle())
```
cmark: It should be filterByTitleExactIgnoreCase not `term`. There is no such field in the Resource model.

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/request/BaseResourceCreateRequest.java
```diff
@@ -27,73 +29,66 @@
...
-	protected String id;
+	String id;
 	
 	@NotEmpty
-	protected String url;
+	String url;
 	
 	@NotEmpty
-	protected String title;
+	String title;
```
cmark: Now they are package-private, which is one step ahead, but not the ideal solution. Use private fields, please! 😄

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/request/BaseResourceCreateRequest.java
```diff
@@ -27,73 +29,66 @@
...
-		this.owner = owner;
-	}
-
-	public final void setContact(String contact) {
-		this.contact = contact;
+	String bundleId;
+	
+	@Override
+	public final String execute(TransactionContext context) {
+		executeAdditionalLogic(context);
```
cmark: Use `postExecute` or `preExecute` instead of this bad method name.

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/request/BaseResourceCreateRequest.java
```diff
@@ -27,73 +29,66 @@
...
-
-	public final void setPurpose(String purpose) {
-		this.purpose = purpose;
+	
+	public final String getTitle() {
+		return title;
 	}
+	
+	/**
+	 * Subclasses may override this method to add extra logic to request execution.
```
cmark: What is extra logic? A better alternative would be
```suggestion
	 * Subclasses may override this method to perform validation checks and/or attach additional resources to the transaction before creating the main resource.
```

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/request/BaseResourceCreateRequest.java
```diff
@@ -27,73 +29,66 @@
...
+	 */
+	protected void executeAdditionalLogic(final TransactionContext context) { }
 
-	public final void setBundleId(String bundleId) {
-		this.bundleId = bundleId;
+	protected abstract ResourceDocument.Builder setSpecializedFields(final ResourceDocument.Builder builder);
```
cmark: Another badly named method. How about
```suggestion
        TODO javadoc
	protected abstract ResourceDocument.Builder prepareResource(final ResourceDocument.Builder builder);
```

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/request/BaseResourceCreateRequest.java
```diff
@@ -27,73 +29,66 @@
...
+	
+	private ResourceDocument createResourceDocument() {
+		final Builder builder = ResourceDocument.builder();
+				
+		return setSpecializedFields(builder)
```
cmark: The order here is important. Superclasses should do their work first, then let subclasses perform theirs.

So the right order here:
1. Create `builder` and initialize with all common fields available in this base class.
2. Call template method `prepareResource` (or we could name it `completeResource`) to let the subclass do its work
3. Return the built document by calling `return builder.build()`.

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/request/BaseResourceSearchRequestBuilder.java
```diff
@@ -0,0 +1,44 @@
...
+		return addOption(OptionKey.TITLE, termFilter);
+	}
+	
+	/**
+	 * Filters matches by their title (exact match).
+	 * 
+	 * @param titles - at least one of these titles match
+	 * @return this builder
+	 */
+	public final RB filterByTermExact(Iterable<String> titles) {
```
cmark: Both `filterByTerm` method should be `filterByTitle` instead.

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/request/BaseResourceUpdateRequest.java
```diff
@@ -28,17 +28,17 @@
...
-	protected String language;
-	protected String description;
-	protected String status;
-	protected String copyright;
-	protected String owner;
-	protected String contact;
-	protected String usage;
-	protected String purpose;
-	protected String bundleId;
+	String url;
```
cmark: Fields should be private here as well, if you need access then create a protected accessor for it.

cmark: Also, add `JsonProperty` to all fields so we get them in the audit logs. Applies to the other requests as well.

cmark (changes requested):

core/com.b2international.snowowl.core.rest.tests/src/com/b2international/snowowl/core/rest/CodeSystemApiAssert.java
kovacstomi: Updated the create request builder to fall back to the previously set bundle ID on `null` value.

cmark (approved):
LGTM!

PR #844, opened 2021-07-08
SO-4446: Update to Java 17

apeteri:
As most of the tooling does not support version 17 yet, I have set `JavaSE-16` as the execution environment, compiling classes with version 16 source and binary compatibility, using a JDK 17 Early-Access build.


apeteri:
Closing in favor of #963 .

PR #850, opened 2021-07-13
SO-4125: Sync value set with query type and simple type reference set 

Henya14:


core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/codesystem/CodeSystem.java
```diff
@@ -55,6 +56,14 @@ public String getResourceType() {
 		return getSettings() == null ? null : (List<String>) getSettings().get(CommonSettings.LOCALES);
 	}
 	
+	
+	public static ResourceURI uriFromPath(String resourcePath) {
```
cmark: Please eliminate this method. Services should never ever try to calculate a `ResourceURI` from a branch path expression.

cmark (changes requested):

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/codesystem/CodeSystem.java
Henya14: Done

cmark (approved):
LGTM!

PR #857, opened 2021-07-19
SO- 4890: support includeNull and pretty query parameters

### LINKED ISSUE #486: Support pretty print option on REST API endpoints
Most of Snow Owl's endpoints return JSON data. 
It would be nice to have a `pretty` query parameter, that pretty prints the output.

cmark:
Both query parameters work with empty or `true` values. Any other value considered `false` and will not enable the feature.
* `includeNull` - includes all properties of the response model to be
serialized to the JSON representation even if the value is `null`. The
default behavior is to omit `null` values.
* `pretty` - pretty print the output to make it human-readable

Fixes #486.

### DIFF
--- a/core/com.b2international.snowowl.core.rest/src/com/b2international/snowowl/core/rest/SnowOwlApiConfig.java
+++ b/core/com.b2international.snowowl.core.rest/src/com/b2international/snowowl/core/rest/SnowOwlApiConfig.java
@@ -17,6 +17,7 @@ package com.b2international.snowowl.core.rest;
 
 import java.lang.reflect.AnnotatedElement;
 import java.lang.reflect.Method;
+import java.util.BitSet;
 import java.util.List;
 import java.util.Map;
 import java.util.TimeZone;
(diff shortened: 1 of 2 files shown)


core/com.b2international.snowowl.core.rest/src/com/b2international/snowowl/core/rest/SnowOwlApiConfig.java
```diff
@@ -120,6 +127,53 @@ public ObjectMapper objectMapper() {
...
+			return objectMappers.get("includeNonNullObjectMapperPretty");
+		} else if (!includeNulls && !pretty) {
+			return objectMappers.get("includeNonNullObjectMapper");
+		} else {
+			throw new NotImplementedException("Unexpected case that should not happen");
+		}
+	}
+
+	private boolean extractBooleanQueryParameterValue(HttpServletRequest request, String queryParameterKey) {
+		String[] values = request.getParameterMap().containsKey(queryParameterKey) ? request.getParameterMap().getOrDefault(queryParameterKey, null) : null;
```
apeteri, line 163: `getOrDefault` already checks for the existence of the key:

```suggestion
		String[] values = request.getParameterMap().getOrDefault(queryParameterKey, null);
```

apeteri (approved):

PR #858, opened 2021-07-19
SO-4892: field selection support

cmark:
This PR adds a new `field` query parameter to all list and get endpoint to restrict the payload size of the returned result and make fetching of documents as fast as possible.


core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/commit/CommitInfo.java
```diff
@@ -31,6 +32,8 @@
 
 	private static final long serialVersionUID = 1L;
 	
+	public static final List<String> DEAFULT_FIELD_SELECTION = List.of(Commit.Fields.ID, Commit.Fields.AUTHOR, Commit.Fields.BRANCH, Commit.Fields.COMMENT, Commit.Fields.TIMESTAMP, Commit.Fields.GROUP_ID);
```
apeteri: These kind of "default set of fields" constants should live in the corresponding, eg. `Commit.Fields` class instead. We already have `ALL`s there which are similar.

apeteri (approved):

PR #863, opened 2021-07-27
feat: support synonym filter in tokenized analyzer configurations

cmark:
New `SO_HOME/configuration/analysis/synonym.txt` file can be configured
to allow Snow Owl use the defined set of synonym rules when performing
full-text searches.
Update docker configuration as well to include an empty synonyms.txt
file (required by all Elasticsearch nodes now).

SO-4895 #resolve

### DIFF
--- a/commons/com.b2international.index/src/com/b2international/index/query/TextPredicate.java
+++ b/commons/com.b2international.index/src/com/b2international/index/query/TextPredicate.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2011-2020 B2i Healthcare Pte Ltd, http://b2i.sg
+ * Copyright 2011-2021 B2i Healthcare Pte Ltd, http://b2i.sg
  * 
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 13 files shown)


commons/com.b2international.index/src/com/b2international/index/query/TextPredicate.java
```diff
@@ -72,11 +72,18 @@ public String toString() {
 
 	public TextPredicate withIgnoreStopwords(boolean ignoreStopwords) {
 		if (ignoreStopwords) {
-			analyzer = Analyzers.TOKENIZED_IGNORE_STOPWORDS.getAnalyzer();
+			return withAnalyzer((analyzer == Analyzers.TOKENIZED_SYNONYMS || analyzer == Analyzers.TOKENIZED_SYNONYMS_IGNORE_STOPWORDS) ? Analyzers.TOKENIZED_SYNONYMS_IGNORE_STOPWORDS : Analyzers.TOKENIZED_IGNORE_STOPWORDS);
 		} else {
-			analyzer = null;
+			return withAnalyzer(null);
```
apeteri: Why don't we have to do the reverse of the code block above? If the analyzer is currently set to `TOKENIZED(_SYNONYMS)?_IGNORE_STOPWORDS`, shouldn't it be changed back to `TOKENIZED$1` (with pseudo-regexp syntax)?

commons/com.b2international.index/src/com/b2international/index/query/TextPredicate.java
```diff
@@ -72,11 +72,18 @@ public String toString() {
...
+		}
+	}
+	
+	public TextPredicate withSynonyms(boolean enableSynonyms) {
+		if (enableSynonyms) {
+			return withAnalyzer((analyzer == Analyzers.TOKENIZED_IGNORE_STOPWORDS || analyzer == Analyzers.TOKENIZED_SYNONYMS_IGNORE_STOPWORDS) ? Analyzers.TOKENIZED_SYNONYMS_IGNORE_STOPWORDS : Analyzers.TOKENIZED_SYNONYMS);
+		} else {
+			return withAnalyzer(null);
```
apeteri: Same with this branch.

apeteri (changes requested):

commons/com.b2international.index/src/com/b2international/index/query/TextPredicate.java
cmark: Right 😄

apeteri (approved):
LGTM!

PR #866, opened 2021-07-27
SO-4894: Use PUT for all update object endpoints

molnarlaura:


snomed/com.b2international.snowowl.snomed.core.rest/src/com/b2international/snowowl/snomed/core/rest/SnomedReferenceSetRestService.java
```diff
@@ -293,7 +293,7 @@ public void executeAction(
 		@ApiResponse(responseCode = "204", description = "No content"),
 		@ApiResponse(responseCode = "404", description = "Branch or reference set not found")
 	})
-	@PutMapping(value="/{id}/members", consumes = { AbstractRestService.JSON_MEDIA_TYPE })
```
cmark: This should use the original endpoint path because this is the Bulk RefSet Member API and not a single object update API. Please revert it back.

cmark (changes requested):

snomed/com.b2international.snowowl.snomed.core.rest/src/com/b2international/snowowl/snomed/core/rest/SnomedReferenceSetRestService.java
cmark: Resolved. 😃

cmark (approved):

PR #869, opened 2021-07-30
SO-4898: Move snomed.languages configuration options to CodeSystem.settings field

molnarlaura:


snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/validation/detail/SnomedValidationIssueDetailExtension.java
```diff
@@ -273,7 +274,9 @@ private void extendConceptIssueLabels(BranchContext context, Collection<Validati
 				.build()
 				.execute(context);
 			
-			Map<String, SnomedDescription> preferredDescriptions = SnomedDescriptionUtils.indexBestPreferredByConceptId(descriptions, locales);
+			@SuppressWarnings("unchecked")
+			ListMultimap<String, String> languageMap = (ListMultimap<String, String>) context.service(TerminologyResource.class).getSettings().get(SnomedTerminologyComponentConstants.CODESYSTEM_LANGUAGE_CONFIG_KEY);
```
apeteri: `SnomedDescriptionUtils` sounds like a good place for a new utility method that extracts the language setting from a terminology resource when a `context` parameter is given. Please replace all occurrences of this line with such a method, it should be reachable from all places where you had to add them.

apeteri (changes requested):

apeteri:
`snowowl.yml` configuration files for tests also need to be updated:

```
Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 4.329 s <<< FAILURE! - in com.b2international.snowowl.core.rest.AllSnowOwlApiTests
com.b2international.snowowl.core.rest.AllSnowOwlApiTests  Time elapsed: 4.328 s  <<< ERROR!
java.lang.RuntimeException: Unrecognized field 'languages' in config file 'snomed'
```

apeteri:
Two issues that we have found:

- the `ObjectMapper` configured for ES indexing does not support Guava data structures such as `ListMultimap`;
- if the above is avoided by using `Map<String, List<String>>` as the type, dynamic mapping on `settings.languages` would detect each language key as a separate property with type `keyword`, leading to a mapping explosion.

This requires changing the data structure of `languages` to a `List<Entry>`, and the removed `SnomedLanguageConfig` will be re-purposed as the entry type.

cmark:
> Two issues that we have found:
> 
> * the `ObjectMapper` configured for ES indexing does not support Guava data structures such as `ListMultimap`;
> * if the above is avoided by using `Map<String, List<String>>` as the type, dynamic mapping on `settings.languages` would detect each language key as a separate property with type `keyword`, leading to a mapping explosion.
> 
> This requires changing the data structure of `languages` to a `List<Entry>`, and the removed `SnomedLanguageConfig` will be re-purposed as the entry type.

How about using a list of special objects then?

molnarlaura:
> > Two issues that we have found:
> > 
> > * the `ObjectMapper` configured for ES indexing does not support Guava data structures such as `ListMultimap`;
> > * if the above is avoided by using `Map<String, List<String>>` as the type, dynamic mapping on `settings.languages` would detect each language key as a separate property with type `keyword`, leading to a mapping explosion.
> > 
> > This requires changing the data structure of `languages` to a `List<Entry>`, and the removed `SnomedLanguageConfig` will be re-purposed as the entry type.
> 
> How about using a list of special objects then?

I will use the original `SnomedLanguageConfig` class as the special object.

snomed/com.b2international.snowowl.snomed.core.rest/src/com/b2international/snowowl/snomed/core/rest/SnomedConceptRestService.java
```diff
@@ -62,6 +62,7 @@ public SnomedConceptRestService() {
 				.build());
 	}
 	
+	@SuppressWarnings("unchecked")
```
cmark: Please remove this leftover annotation.

snomed/com.b2international.snowowl.snomed.core.rest/src/com/b2international/snowowl/snomed/core/rest/domain/SnomedConceptRestSearch.java
```diff
@@ -62,7 +62,7 @@
 	private String[] statedAncestor;
 	@Parameter(description = "doi (degree-of-interest-based scoring)")
 	private Boolean doi = null;
-
+
```
cmark: Please remove the unnecessary whitespace changes.

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/config/SnomedLanguageConfig.java
```diff
@@ -27,27 +27,27 @@
  */
 public class SnomedLanguageConfig {
 
-	private String code;
-	private List<String> refSetIds;
+	private String languageTag;
+	private List<String> languageRefSetIds;
 	
 	public SnomedLanguageConfig(String code, String...refSetIds) {
```
cmark: Rename ctor argument names here as well to match the field names.

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/config/SnomedLanguageConfig.java
```diff
@@ -27,27 +27,27 @@
...
 		this(code, ImmutableList.copyOf(refSetIds));
```
cmark: Use `List.copyOf` instead.

cmark (changes requested):

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/request/SnomedConceptSearchRequest.java
```diff
@@ -116,7 +117,12 @@
 		/**
 		 * Enable score boosting using DOI field
 		 */
-		USE_DOI,
+		USE_DOI, 
+		
+		/**
+		 * Match concept descriptions where the description has language membership in on of the provided locales
```
apeteri: Accidentally a letter.

```suggestion
		 * Match concept descriptions where the description has language membership in one of the provided locales
```

apeteri (approved):
LGTM! Please fix the typo in the comment, and it is ready for merging.

cmark (approved):

PR #873, opened 2021-08-05
SO-4900: Improve messages, reorder path elements, allow source code ...

AAAlinaaa:
...system configuration during synchronization


cmark (approved):
LGTM!

PR #876, opened 2021-08-12
feat: automatically configure known language dialect aliases after...

cmark:
...successful RF2 import

Existing dialect configuration is not updated and kept as is so client
can act if they need to include additional refset IDs after an RF2
import.


snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/request/rf2/SnomedRf2ImportRequest.java
```diff
@@ -353,27 +360,52 @@ private DB createDb() {
...
 			.filterByType(SnomedRefSetType.LANGUAGE)
 			.filterByActive(true)
 			.setFields(SnomedConceptDocument.Fields.ID)
 			.sortBy(SortField.ascending(SnomedConceptDocument.Fields.ID))
 			.build()
-			.execute(context)
+			.execute(context);
+		
+		/*
+		 * XXX: The default language in locales is always "en", as there is no
```
apeteri, line 374: We could get the language code part from the dataset by looking at the language code of each description that is a member of the reference set in question. Something for the future.

apeteri (approved):
LGTM!

PR #878, opened 2021-08-17
Add suggest REST endpoint

kovacstomi:


core/com.b2international.snowowl.core.rest/src/com/b2international/snowowl/core/rest/suggest/SuggestRestService.java
```diff
@@ -0,0 +1,103 @@
...
+	}
+	
+	@Operation(
+			summary = "Concept suggestion", 
+			description = "Returns an actual concept of the specified code system based on the source term.")
+	@ApiResponses({ 
+		@ApiResponse(responseCode = "200", description = "OK"),
+		@ApiResponse(responseCode = "400", description = "Bad Request") 
+	})
+	@PostMapping
```
cmark: Specify `consumes` for both this and the bulk endpoint, please.

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/request/ConceptSuggestionRequest.java
```diff
@@ -68,7 +70,7 @@
 	@Min(1)
 	private int minOccurrenceCount;
 	
-	private transient List<String> topTokens;
+	private transient List<String> topTokens = Collections.emptyList();
```
cmark: I think the `emptyList()` assignment is not needed,

cmark (changes requested):

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/request/ConceptSuggestionRequest.java
kovacstomi: You are right, it's not needed 🎉

cmark (approved):
LGTM!

PR #880, opened 2021-08-17
SO-4113: authorization improvements

cmark:
This PR introduces new authorization options to allow/restrict access to server resources.
Access tokens describe the necessary scopes to give access to certain resources, either by:
* directly denoting the resource ID in the scope
* denoting the containing bundle ID in the scope (direct `bundleId` parent for now)
* or by allowing access to all resources (`*`)


core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/request/BaseResourceSearchRequest.java
```diff
@@ -85,6 +88,27 @@ protected final Expression prepareQuery(RepositoryContext context) {
...
 	
+	/**
+	 * Configures security filters to allow access to certain resources only. This method is no-op if the given {@link User} is an administrator or has read access to everything. 
+	 * 
+	 * @param queryBuilder - the query builder to append the clauses to
+	 * @param user - the user who's permissions will be applied to this resource search request
+	 */
+	protected final void addSecurityFilter(ExpressionBuilder queryBuilder, User user) {
+		if (!user.isAdministrator() && !user.hasPermission(Permission.requireAll(Permission.OPERATION_BROWSE, Permission.ALL))) {
+			Set<String> permittedResourceIds = user.getPermissions().stream().flatMap(p -> p.getResources().stream()).collect(Collectors.toSet());
```
apeteri, line 99: Does having any sort of permission on a resource imply `BROWSE` rights? With the currently used operations (edit, import, export, version, promote task, classify), it feels so.

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/identity/BasePermission.java
```diff
@@ -0,0 +1,84 @@
...
+	}
+	
+	@Override
+	public final String getOperation() {
+		return operation;
+	}
+	
+	@Override
+	public final String getPermission() {
+		if (permission == null) {
```
apeteri: For thread safety reasons, this should be put into a memoizing `Supplier` instead.

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/identity/User.java
```diff
@@ -50,7 +54,10 @@ public String getUsername() {
 	
 	@JsonIgnore
 	public List<Permission> getPermissions() {
-		return getRoles().stream().flatMap(role -> role.getPermissions().stream()).distinct().collect(Collectors.toList());
+		if (permissions == null) {
```
apeteri: Same here.

apeteri (changes requested):
`GetResourceRequest` should also implement `AccessControl`, the same way as `UpdateRequest` does; I think I saw some sub-classes in the other repository where the `implements` keyword was removed entirely.

cmark:
> `GetResourceRequest` should also implement `AccessControl`, the same way as `UpdateRequest` does; I think I saw some sub-classes in the other repository where the `implements` keyword was removed entirely.

Unfortunately, we cannot apply the same pattern there because certain subclasses (like `CodeSystemGetRequest`) should NOT check for access control. Scenarios where the token only states that the user has access to a bundle (and implicitly all its contained resources) would fail early because the permissions only describe the bundle and getting the nested resource requires evaluation of the described permissions, so this becomes a post-check as opposed to a pre-check. 
What I could do instead is to introduce a new subclass that implements the same logic as the UpgradeRequest, but with the `browse` scope, but certain get requests require other scopes like `ClassificationGetRequest`, which require a non-default scope, which would further increase the complexity of that hierarchy, so I have decided to ditch the idea.

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/request/BaseResourceSearchRequest.java
cmark: Yes, if you have at least one permission with other than the `browse` scope that includes the `browse` scope as well. For example, you would not be able to edit a resource without reading it first if this wasn't true.

cmark:
@apeteri fixed (or answered) all requested changes (comments).

apeteri:
`EclSerializerTest` failed in a previous CI build (`NullPointerException` was thrown in a `Promise` instead of concepts being returned), but it does not seem to be related to authorization changes. The concept search request invoked has a `User` instance registered with permissions in its context.

apeteri (approved):

PR #882, opened 2021-08-24
SO-4916: Add language to suggest API parameters

kovacstomi:


core/com.b2international.snowowl.core.rest/src/com/b2international/snowowl/core/rest/suggest/SuggestRestParameters.java
```diff
@@ -34,6 +34,9 @@
 	
 	@Parameter(description = "The preferred term display in case of SNOMED CT", example = "PT", schema = @Schema(allowableValues = {"FSN", "PT", "ID_ONLY"}, defaultValue = "PT"))
 	private String preferredDisplay = "PT";
+	
+	@Parameter(description = "Accepted language tags, in order of preference", example = "en-US;q=0.8,en-GB;q=0.6")
+	private String acceptLanguage = "en-US;q=0.8,en-GB;q=0.6";
```
cmark: This should be defined as a header, not as a query parameter, so it is consistent with the other endpoints.

cmark (changes requested):

cmark (approved):
LGTM!

PR #884, opened 2021-08-25
SNRAY-115: Concept suggestion API and related changes

apeteri:


core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/context/TerminologyResourceContentRequest.java
```diff
@@ -39,9 +40,9 @@ public TerminologyResourceContentRequest(final Request<BranchContext, R> next) {
 	public R execute(TerminologyResourceContext context) {
 		final ResourceURI resourceURI = context.resourceURI();
 		final TerminologyResource resource = context.resource();
-		final String branchPath = context.service(ResourceURIPathResolver.class).resolve(context, resourceURI, resource);
+		final PathWithVersion branchPathWithVersion = context.service(ResourceURIPathResolver.class).resolveWithVersion(context, resourceURI, resource);
 		return new RepositoryRequest<R>(resource.getToolingId(),
```
cmark, line 53: How about injecting the version's ResourceURI here instead of in the BranchRequest? It makes it simpler to read and reason about and only a single place in the code when we have to change it. 😉

cmark (changes requested):

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/context/TerminologyResourceContentRequest.java
apeteri: :done:

cmark (approved):
LGTM!

PR #885, opened 2021-08-25
Add test for faulty reparenting

molnarlaura:
This test will fail until the field ancestorBundleIds won't be introduced.
https://snowowl.atlassian.net/browse/SNRAY-283


apeteri:
Test case included in #891, closing PR.

PR #887, opened 2021-08-31
SNRAY-315: Add support for resource URIs with timestamp part

apeteri:
The format changes to `{type}/{name}[/{path}][@{timestamp}]`, where both `path` and `timestamp` are optional. The timestamp should be a valid 64-bit value. Resource names and paths may not contain any 'at' characters.


molnarlaura (commented):
Looks good to me! 👍

nagyo (approved):
LGTM!

PR #891, opened 2021-09-07
SO-4923: Index ancestor bundle identifiers on resources

apeteri:
Access control was also extended to allow or disallow access to nested bundles.


cmark (approved):
LGTM!

PR #892, opened 2021-09-15
Sort resource search result by type

molnarlaura:


core/com.b2international.snowowl.core.rest/src/com/b2international/snowowl/core/rest/resource/ResourceRestService.java
```diff
@@ -53,6 +59,10 @@ public ResourceRestService() {
 	})
 	@GetMapping
 	public Promise<Resources> searchByGet(@ParameterObject final ResourceRestSearch params) {
+		List<Sort> sortFields = Lists.newArrayList();
+		sortFields.add(SortScript.of("typeSort", Map.of(), true));
```
apeteri: Since we have the option of passing in parameters here, it might make sense to externalize the priority/order for each resource type as a Map:

```suggestion
		final Map<String, String> orderByType = Map.of(
			"bundles", "1",
			"codesystems", "2",
			"valuesets", "3",
			"conceptmaps", "4"
		);
		sortFields.add(SortScript.of("typeSort", Map.of("orderByType", orderByType), true));
```

Then on the scripting side, the chain of `if`s (unfortunately Painless does not support `switch` statements) could be replaced with a single call to `Map#getOrDefault`.

molnarlaura: 👍 Changed the script as well.

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/internal/ResourceDocument.java
```diff
@@ -60,6 +61,13 @@
 	}
 )
 @JsonDeserialize(builder = ResourceDocument.Builder.class)
+@Script(
+		name="typeSort", 
+		script=
+		 "if (params.orderByType.containsKey(doc.resourceType.value)) {"
+		+ "	return params.orderByType.get(doc.resourceType.value)"
+		+ "}"
+		+ "return '5'")
```
apeteri: The script should fall back to sorting by `doc.resourceType.value` if the map does not have an entry for it.

apeteri (approved):
LGTM, with one small change request. 🛩️

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/internal/ResourceDocument.java
molnarlaura: Fixed.

PR #898, opened 2021-09-16
feat(api): change `/snomed-ct/v3` endpoints to use just `/snomedct`

cmark:


apeteri (approved):
I couldn't find any other matches for `/v3` in my workspace either. 👍

nagyo (approved):
👍🏻

PR #900, opened 2021-09-20
RF2 export field selection

cmark:
This PR improves the RF2 export logic to load only the minimum set of fields instead of loading entire index documents. Also, this PR renames the `editorialGuideReference` field to `guideURL` to match the RF2 spec.


snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/request/SnomedRelationshipSearchRequest.java
```diff
@@ -63,6 +71,18 @@
...
+		if (fields().contains(SnomedRelationship.Fields.VALUE)) {
+			additionalFieldsToLoad.add(SnomedRelationshipIndexEntry.Fields.VALUE_TYPE);
+		}
+	}
+	
+	@Override
+	protected Multimap<String, String> collectFieldsToLoadReplacements() {
+		return REPLACE_VALUE_FIELD;
+	}
+
```
apeteri: How about adding `SnomedRelationshipIndexEntry.Fields.VALUE_TYPE` as a fourth value for key `SnomedRelationship.Fields.VALUE` in the Multimap? That should handle both cases.

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/request/SearchIndexResourceRequest.java
```diff
@@ -90,7 +74,54 @@ protected final B doExecute(C context) throws IOException {
...
+		final Multimap<String, String> fieldReplacements = collectFieldsToLoadReplacements();
+		if (!fieldReplacements.isEmpty()) {
+			fields = Lists.newArrayList();
+			for (String fieldToLoad : fields) {
+				Collection<String> replacements = fieldReplacements.get(fieldToLoad);
+				if (replacements.isEmpty()) {
+					fields.add(fieldToLoad);
+				} else {
+					fields.addAll(replacements);
+				}
```
apeteri: There is no duplicate check for these fields. A client could request a field that is a replacement of another requested field.

apeteri (changes requested):

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/request/SnomedRelationshipSearchRequest.java
cmark: Hmm, true 😃

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/request/SearchIndexResourceRequest.java
cmark: Hmm, yes, but that is unnecessary I think, the underlying system can handle multiple requests of the same field as well. Like requesting `id`, `id`, `id` should still be able to load the document without failing or broken. Partial loading (which is only supported in low-level requests at the moment) is another story, but we will figure that one out when we get there. I would leave this like this for now.

apeteri: Then line 85 and the block starting in 91 can be simplified greatly.

cmark: Not sure about line 85, but yes, the block starting in 91 can be simplified.

apeteri (approved):
🍃

AAAlinaaa (approved):

PR #901, opened 2021-09-20
SO-4935 Selective export

AAAlinaaa:


cmark (changes requested):
While the PR itself looks good, I would wait until we confirm that this would be useful in any actual user scenario. Putting this to on hold.

PR #907, opened 2021-09-29
SO-4942: Support delta RF2 exports with branch path range

apeteri:
(Also includes changes from 8.x, where support for snapshot exports using branch paths with a timestamp were introduced.)


cmark (approved):
LGTM!

nagyo (approved):
👍🏻

PR #929, opened 2021-10-21
feat(core): allow fully customizable indices settings in `snowowl.yml` configuration file

cmark:
Each index can receive its own custom settings object nested under `repository.index` via its index name. If not specified, index
initialization will fall back to the default values.

### DIFF
--- a/cis/com.b2international.snowowl.snomed.cis/src/com/b2international/snowowl/snomed/cis/SnomedIdentifierPlugin.java
+++ b/cis/com.b2international.snowowl.snomed.cis/src/com/b2international/snowowl/snomed/cis/SnomedIdentifierPlugin.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2011-2019 B2i Healthcare Pte Ltd, http://b2i.sg
+ * Copyright 2011-2021 B2i Healthcare Pte Ltd, http://b2i.sg
  * 
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -21,6 +21,7 @@ import java.nio.file.Files;
 import java.util.Set;
 import java.util.stream.Collectors;
 
+import org.elasticsearch.core.Map;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -28,6 +29,7 @@ import com.b2international.index.Index;
 import com.b2international.index.Indexes;
 import com.b2international.index.mapping.Mappings;
 import com.b2international.snowowl.core.config.IndexSettings;
+import com.b2international.snowowl.core.config.RepositoryConfiguration;
 import com.b2international.snowowl.core.config.SnowOwlConfiguration;
 import com.b2international.snowowl.core.plugin.Component;
 import com.b2international.snowowl.core.setup.ConfigurationRegistry;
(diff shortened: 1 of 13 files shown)


core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/repository/TerminologyRepository.java
```diff
@@ -83,9 +83,13 @@ public String id() {
...
 		
-		final Map<String, Object> indexSettings = newHashMap(context.service(IndexSettings.class));
-		indexSettings.put(IndexClientFactory.NUMBER_OF_SHARDS, context.service(SnowOwlConfiguration.class).getModuleConfig(RepositoryConfiguration.class).getIndexConfiguration().getNumberOfShards());
-		final IndexClient indexClient = Indexes.createIndexClient(repositoryId, mapper, mappings, indexSettings);
+		IndexConfiguration indexConfiguration = context.service(RepositoryConfiguration.class).getIndexConfiguration();
+		final IndexClient indexClient = Indexes.createIndexClient(
+			repositoryId, 
+			mapper, 
+			mappings, 
+			context.service(IndexSettings.class).forIndex(indexConfiguration, repositoryId)
```
apeteri, line 91: Same for `Indexes.createIndexClient`, although the client name here is already parameterized.

cis/com.b2international.snowowl.snomed.cis/src/com/b2international/snowowl/snomed/cis/SnomedIdentifierPlugin.java
```diff
@@ -116,7 +117,12 @@ private void registerSnomedIdentifierService(final SnomedIdentifierConfiguration
 
 		switch (conf.getStrategy()) {
 		case EMBEDDED:
-			final Index index = Indexes.createIndex(SNOMED_IDS_INDEX, env.service(ObjectMapper.class), new Mappings(SctId.class), env.service(IndexSettings.class));
+			final Index index = Indexes.createIndex(
+				SNOMED_IDS_INDEX, 
+				env.service(ObjectMapper.class), 
+				new Mappings(SctId.class), 
+				env.service(IndexSettings.class).forIndex(env.service(RepositoryConfiguration.class).getIndexConfiguration(), SNOMED_IDS_INDEX)
```
apeteri, line 124: The request to get the specific index configuration could be pushed down to `Indexes.createIndex` itself, as it already knows about the index name passed in the first arg. Then you don't have to change all invocations.

apeteri (changes requested):

nagyo (approved):
LGTM!

cis/com.b2international.snowowl.snomed.cis/src/com/b2international/snowowl/snomed/cis/SnomedIdentifierPlugin.java
cmark: The problem is a reference to the `IndexConfiguration` class, which is a Snow Owl specific class with Snow Owl defaults, while the index layer is treated like a library and unfortunately it does not have access to that class.

nagyo (approved):
👍🏻

apeteri (approved):

PR #931, opened 2021-10-25
Support index-level request caching

cmark:
This PR introduces index-level search request caching and potential configuration on request API level via [cacheHits(ServiceProvider)](https://github.com/b2ihealthcare/snow-owl/commit/163181e3982809171ef740013a72e73df31174d3#diff-b854641ff991e8157af0c445f61bebd4c94474b2434c9ac8ae2bb8c4d1fce435R157).
By default the system will cache:
* all requests that are being executed on versioned content
* certain ECL search requests (or partial ECL search requests)

### DIFF
--- a/core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/request/SearchIndexResourceRequest.java
+++ b/core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/request/SearchIndexResourceRequest.java
@@ -61,7 +61,6 @@ public abstract class SearchIndexResourceRequest<C extends ServiceProvider, B, D
 			fields = configureFieldsToLoad(fields);
 		}
 		
-		
 		final Hits<D> hits = searcher.search(Query.select(getSelect())
 				.from(getFrom())
 				.fields(fields)
@@ -155,7 +154,7 @@ public abstract class SearchIndexResourceRequest<C extends ServiceProvider, B, D
 	 * @return whether the search should cache the resulting hits in the underlying index system or not
 	 */
 	protected boolean cacheHits(C context) {
-		return context.optionalService(PathWithVersion.class).isPresent();
+		return Revision.class.isAssignableFrom(getFrom()) && context.optionalService(PathWithVersion.class).isPresent();
 	}
 	
 	/**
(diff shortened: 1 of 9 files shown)


core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/request/SearchIndexResourceRequest.java
```diff
@@ -147,6 +148,15 @@ protected boolean trackScores() {
 		return false;
 	}
 	
+	/**
+	 * Subclasses may override to configure caching. By default search requests that are executed against a version URI will be cached.  
+	 * @param context - the context that can be used to determine whether caching should be enabled for this search request or not
+	 * @return whether the search should cache the resulting hits in the underlying index system or not
+	 */
+	protected boolean cacheHits(C context) {
+		return Revision.class.isAssignableFrom(getFrom()) && context.optionalService(PathWithVersion.class).isPresent();
```
apeteri, line 157: `PathWithVersion` does not always contain a version URI; when the original request path does not correspond to a version identifier, eg. when it contains a branch separator character, only an absolute branch path will be returned from the resolver: https://github.com/b2ihealthcare/snow-owl/blob/eb3e28d471f2e18dd1a7455691bfdbfb1d32caca/core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/uri/DefaultResourceURIPathResolver.java#L76-L80

apeteri (changes requested):

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/request/SearchIndexResourceRequest.java
nagyo: I agree, it would be nice to introduce a more rigorous condition for this, given that it is quite hard to validate if a search result is indeed correct when it was cached.

nagyo (changes requested):

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/request/SearchIndexResourceRequest.java
cmark: I'm open to any suggestions. This was the easiest way at the moment to determine whether a request is being executed on a version or not (and I think it's not that bad 😅).

cmark: @apeteri see this piece of code, the PathWithVersion instance gets injected only when it has a version URI:
https://github.com/b2ihealthcare/snow-owl/pull/931/files#diff-a4877e9655543b6f3a4aeb80fb2df33cfb521470f823c273b36214aad0b087ecR47

apeteri (approved):

nagyo (approved):
👍🏻

PR #937, opened 2021-11-11
feat(auth): configurable OAuth access token verification and signing

cmark:
This PR adds support for proper OAuth access token (JWT) handling by supporting 
* HMAC (256/512 bits) and RSA (256/512 bits) based algorithms (EC and advanced algorithms are not supported).
  * `verificationKey` and `signingKey` loading in standard Base64 encoded PKCS#8 (PEM file) format is supported
* Support loading public verification keys (JSON Web Key Set) from external URLs
* Configuration of the `email` and `permissions` claims that are required to identify and authorize the user


core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/identity/IdentityPlugin.java
```diff
@@ -49,11 +77,16 @@ public void addConfigurations(ConfigurationRegistry registry) {
...
 		final IdentityConfiguration conf = configuration.getModuleConfig(IdentityConfiguration.class);
-		final List<IdentityProvider> providers = createProviders(env, conf.getProviderConfigurations());
+		final List<IdentityProvider> providers = createProviders(env, conf.getProviderConfigurations() == null ? List.of() : conf.getProviderConfigurations());
 		
-		IdentityProvider identityProvider = null; 
+		IdentityProvider identityProvider = null;
 		if (providers.isEmpty()) {
-			identityProvider = IdentityProvider.NOOP;
+			// if there are no providers, but the issuer is external and there is a JWS configured, then assume we are using access tokens and Snow Owl is a resource server
+			if (conf.isExternalIssuer() || !Strings.isNullOrEmpty(conf.getJws())) {
```
apeteri, line 130: Is it allowed/useful to set a non-Snow Owl issuer in the config, but leave the JWS property empty? We obtain a `DecodedJWT` via verification in `IdentityProvider`, and I'd think it will break when it tries to use the algorithm implementation that always throws `BadRequestException`s.

apeteri (changes requested):

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/identity/IdentityPlugin.java
cmark: If a non-Snow Owl issuer is configured, then a JWS must be specified, otherwise, it will throw the disabled BadRequest exception message, which is the indication of the misconfiguration (I did not want to use an HTTP 50x error).
I can add another check that the JWS parameter is required when non-Snow Owl issuer is configured. What do you think?

apeteri: Would it work if this check was reduced to:

```suggestion
			if (!Strings.isNullOrEmpty(conf.getJws())) {
```

cmark: In this case, Snow Owl would need to generate a JWT, but that would require an identity provider to fetch the user data from plus verify the user/pass. So, no unfortunately this is not an option here.

nagyo (approved):
👍🏻

apeteri (approved):

nagyo (approved):
👍🏻

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/identity/JWTGenerator.java
```diff
@@ -34,19 +37,27 @@
...
+	public JWTGenerator(final Algorithm algorithm, final RSAKeyProvider keyProvider, final String issuer, final String emailClaimProperty, final String permissionsClaimProperty) {
 		this.algorithm = algorithm;
+		this.keyProvider = keyProvider;
 		this.issuer = issuer;
 		this.emailClaimProperty = emailClaimProperty;
 		this.permissionsClaimProperty = permissionsClaimProperty;
 	}
 	
 	public String generate(String email, Map<String, Object> claims) {
+		if (algorithm == null || keyProvider != null && keyProvider.getPrivateKey() == null) {
```
apeteri: I'd split this check in two, with a different error message for each case:
- `algorithm == null` &rarr; `"No algorithm specified, cannot sign"`
- `keyProvider != null && keyProvider.getPrivateKey() == null` &rarr; `"Algorithm requires private key but none given, cannot sign"`

apeteri (changes requested):

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/identity/JWTGenerator.java
cmark: Hmm, both represent the same thing for me, token signing is not configured, aka disabled, aka user cannot generate a token. 😕

cmark: Another alternative would be to introduce an interface for JWTGenerator and use a different implementation when the service is disabled.

apeteri: Just to help out with deciphering the error message 😄

cmark: Refactored the relevant code statements and classes to be more expressive and easier to read and understand. 🎶

apeteri (approved):

cmark:
🙇

PR #942, opened 2021-11-23
Automatically remove stale remote jobs (8.x)

nagyo:

PR #943, opened 2021-11-23
Automatically remove stale remote jobs (7.x)

nagyo:


cis/com.b2international.snowowl.snomed.cis.rest/src/com/b2international/snowowl/snomed/cis/rest/CisBulkSctIdService.java
```diff
@@ -230,6 +230,7 @@
 				.setDescription(jobDescription)
 				.setRequest(request)
 				.setUser(User.SYSTEM.getUsername())
+				.setAutoClean(true)
```
apeteri: Note that jobs that run into a problem will not be removed: https://github.com/b2ihealthcare/snow-owl/blob/f602f1addc56dd466941121c77c3c88768022a0b/core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/jobs/RemoteJob.java#L102-L123

On the `CisBulkJobService` CIS API level, we are offering to return jobs with a `finished` status as well, these will not be available if they are auto-cleaned. The service relies directly on the remote job storage. In case of a bulk generate job, this means the resulting SCTIDs are not available (maybe for registers this can be done because we already know what is being registered).

apeteri (changes requested):

cis/com.b2international.snowowl.snomed.cis.rest/src/com/b2international/snowowl/snomed/cis/rest/CisBulkSctIdService.java
apeteri: Although someone who monitors progress by polling the CIS bulk jobs API will not see when their registration job finishes. 😕

cmark: Yeah agree, this is unfortunately against the current CIS API design, so as @nagyo already did, I suggest reverting the changes and using a better alternative like explicit clean-up tasks or GC style just in time cleanups (when a new request comes in the system will try to detect and clean any obsolete remote jobs automatically).

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/jobs/RemoteJobTracker.java
```diff
@@ -63,14 +70,18 @@ public void run() {
 							.from(RemoteJobEntry.class)
 							.where(
 								Expressions.builder()
-									.filter(RemoteJobEntry.Expressions.deleted(true))
-									.filter(RemoteJobEntry.Expressions.done())
+									.must(RemoteJobEntry.Expressions.done())
+									.must(Expressions.builder()
```
apeteri: `filter` is more efficient than `must`; the semantics are the same, but it will not score documents based on the sub-expressions, and we don't need to keep score here.

apeteri (changes requested):

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/jobs/RemoteJobTracker.java
apeteri: @nagyo pointed out that `must` clauses turn into `filter`s within `EsQueryBuilder` if there are no text matching queries or script scoring (in which case the `needsScoring` variable in line 280 in the first snippet remains `false`): 

https://github.com/b2ihealthcare/snow-owl/blob/311c96ec555a64e7762c8dd34c76ce0b35e66ee3/commons/com.b2international.index/src/com/b2international/index/es/query/EsQueryBuilder.java#L232-L283

https://github.com/b2ihealthcare/snow-owl/blob/4712e259a6dd22adb06f392da6ca0549520977ae/commons/com.b2international.index/src/com/b2international/index/es/query/EsQueryBuilder.java#L174-L186

nagyo: https://github.com/b2ihealthcare/snow-owl/blob/246460bb004abb6caaa6f3d474fbf46c659166ca/commons/com.b2international.index/src/com/b2international/index/es/query/EsQueryBuilder.java#L180-L185

This piece does the magic of converting musts to filters, in case no score is needed, so the end result will be a filter 🎉

apeteri (approved):
👍. I am hesitant to recommend changing `must`s to `filter`s; the functionality is unlikely to break in the future, on the other hand it might spare come cycles on the (human) reader's end if they don't know that `EsQueryBuilder` handles boolean clauses a bit differently from ES or Lucene.

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/jobs/RemoteJobTracker.java
```diff
@@ -310,7 +344,11 @@ public void done(IJobChangeEvent event) {
 				}
 				params.put("state", newState.name());
 				params.put("finishDate", System.currentTimeMillis());
+				
 				update(jobId, RemoteJobEntry.WITH_DONE, params.build());
+				
+				LOG.trace("Incrementing completed job counter to {}", completedJobCounter.incrementAndGet());
```
apeteri: Please separate the counter increment to a separate line; it is easy to mistake this logging statement as a side-effect free operation:

```suggestion
				final long completedJobs = completedJobCounter.incrementAndGet();
				LOG.trace("Incrementing completed job counter to {}", completedJobs);
```

apeteri (approved):
👍 with a tiny request below.

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/jobs/RemoteJobTracker.java
```diff
@@ -203,11 +168,84 @@ private void update(String jobId, String script, Map<String, Object> params) {
...
+					.build());
+			if (hits.getTotal() > 0) {
+				LOG.trace("Purging deleted / stale job entries {}", hits.getHits());
+				writer.remove(RemoteJobEntry.class, Set.copyOf(hits.getHits()));
+				writer.commit();
+			}
+			return null;
+		});
+		
+		completedJobCounter.set(getNumberOfCompletedJobs());
```
cmark: The counter here should be set to `0` and used for counting the number of completed jobs, then triggering the purge when at least X (configurable threshold) jobs have completed since the last purge.

cmark (changes requested):

AAAlinaaa (approved):
LGTM

cmark (approved):
LGTM!

apeteri:
The 8.x counterpart of this PR is https://github.com/b2ihealthcare/snow-owl/pull/942.

PR #944, opened 2021-11-23
SO-4965 Publish sct ids on version create

AAAlinaaa:


snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/core/version/SnomedVersioningRequest.java
```diff
@@ -145,60 +146,75 @@ protected void doVersionComponents(TransactionContext context) throws Exception
 	
 	private void versionComponents(TransactionContext context, Iterable<? extends SnomedDocument> componentsToVersion, Multimap<String, String> componentIdsByReferringModule) {
 		
-		for (SnomedDocument componentToVersion : componentsToVersion) {
-			// register IDs for publication
-			if (componentToVersion instanceof SnomedComponentDocument) {
-				componentIdsToPublish.add(componentToVersion.getId());
-			}
+		try (TransactionContext ctx = new RepositoryTransactionContext(context, context.author(), "Versioning components", context.parentLock())) {
```
apeteri: Did `context.openTransaction` not work? I'd prefer using it instead of instantiating a concrete context implementation.

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/core/version/SnomedVersioningRequest.java
```diff
@@ -145,60 +146,75 @@ protected void doVersionComponents(TransactionContext context) throws Exception
...
+			ctx.commit();
+		
+			if (!CompareUtils.isEmpty(componentIdsToPublish)) {
+				SnomedRequests.identifiers().preparePublish()
+					.setComponentIds(componentIdsToPublish)
+					.build()
+					.execute(ctx);
+			}
+		} catch (Exception e) {
+			throw new SnowowlRuntimeException(e);
```
apeteri: `ctx.rollback` should be called if an exception occurs. `TransactionalRequest` would handle this, but not for the manually managed "inner" transaction context instance.

apeteri (changes requested):

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/core/version/SnomedVersioningRequest.java
apeteri: Side note: this could be done in `VersioningRequest` as well, as the capped transaction context is also locally managed, but it gets a bit more complicated in the general case as it might have already committed some pieces successfully, in which case rollback will be of no use. But what can be applied here is the error handling pattern from `TransactionalRequest`: 

```java
TransactionContext tx = null;
try (final TransactionContext transaction = context.openTransaction(context, author, "Versioning components", parentLockContext)) {
  tx = transaction;
  // Versioning work happens here
} catch (ApiException e) {
  if (tx != null) { tx.rollback(); }
  throw e;
} catch (Exception e) {
  if (tx != null) { tx.rollback(); }
  throw SnowowlRuntimeException.wrap(e);
}
```

apeteri (approved):
👍

apeteri (approved):
👍

apeteri:
OK, maybe one thing 😄 `commit` itself could be overridden in `CappedTransactionContext` to call back to the `BiConsumer`, then there is no chance of missing a call to it anywhere else in the code.

cmark:
Done.

apeteri (approved):

apeteri:
The 8.x counterpart of this PR is https://github.com/b2ihealthcare/snow-owl/pull/942.

PR #945, opened 2021-11-23
SO-4964: Ensure id is registered in nested create requests

AAAlinaaa:


snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/request/SnomedConceptUpdateRequest.java
```diff
@@ -319,7 +319,7 @@ private boolean updateSubclassDefinitionStatus(final TransactionContext context,
 			.map(componentId -> {
 				if (!previousComponentIds.contains(componentId) && currentComponentsById.containsKey(componentId)) {
 					// new component
-					return currentComponentsById.get(componentId).toCreateRequest(conceptId);
+					return new IdRequest<>(currentComponentsById.get(componentId).toCreateRequest(conceptId));
```
apeteri: There are three existing uses of `IdRequest` in the current code:

1. Creating a simple type reference set while creating a query member: https://github.com/b2ihealthcare/snow-owl/blob/ec7fc8e590f391a7f466fab1480c55a0f9b01af7/snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/request/SnomedQueryMemberCreateDelegate.java#L58
2. Saving a batch of classification changes: https://github.com/b2ihealthcare/snow-owl/blob/ec7fc8e590f391a7f466fab1480c55a0f9b01af7/snomed/com.b2international.snowowl.snomed.reasoner/src/com/b2international/snowowl/snomed/reasoner/request/SaveJobRequest.java#L211-L218
3. Nesting incoming requests in the SNOMED CT-specific commit request builder: https://github.com/b2ihealthcare/snow-owl/blob/ec7fc8e590f391a7f466fab1480c55a0f9b01af7/snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/request/SnomedRepositoryCommitRequestBuilder.java#L51-L53

Given the existing usage in 1), I am OK with accepting the change, however it should be noted that 1) is not great either because it does not release the SCTID if the request fails, 2) seems to be redundant (or maybe even problematic as there are two `IdRequest` instances in the chain) and 3) has a problem with the order of nesting requests.

If we'd like `IdActionRecorder` to only commit changes (ie. register recorded reservations and _not_ roll back recorded registrations) only if the transaction in the index succeeds, requests would have to be encapsulated in the following order:

```
IdRequest
  ╰ TransactionalRequest
      ╰ <any business or bulk request>
```

A more optimal solution for the general problem (not the immediate issue) would have the following features:

- Moves `IdRequest` to become the parent of `TransactionalRequest` in the request chain
- Interrogates `SnomedConceptUpdateRequest` in advance for the SCTIDs it intends to use, via eg. the `Request#getNestedRequests` mechanism (currently only implemented in `SnomedConceptCreateRequest`)

apeteri (approved):

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/request/SnomedConceptUpdateRequest.java
cmark: I think this is a good time to refactor the SCT ID management code pieces, so I vote for a bigger refactor than just patching the bug. Similarly to the `ModuleRequest` it is possible to wrap a `TransactionalRequest` around with an `IdRequest` instance and handle all the id registration logic there and roll back when the transaction fails, etc. The only problem is that SnomedConceptUpdateRequest does not actually know at pre-execution whether the current list of description entries will result in create, update or delete requests, so registration (and generation) of SCT IDs will need to be either delayed to the point where we know all of them or we need a way to generate IDs pre-commit instead of pre-execution, but that requires temporary SCT IDs (or even UUIDs) to be assigned and then replaced before commit. We can chat about this and decide which way would be the best to handle all the use cases.

snomed/com.b2international.snowowl.snomed.core.rest.tests/src/com/b2international/snowowl/snomed/core/rest/SnomedConceptUpdateIDManagementTest.java
```diff
@@ -0,0 +1,108 @@
...
+import com.b2international.snowowl.snomed.datastore.request.SnomedRequests;
+import com.b2international.snowowl.test.commons.Services;
+
+/**
+ * @since 7.18.0
+ */
+public class SnomedConceptUpdateIDManagementTest extends AbstractSnomedApiTest {
+
+	@Test
+	public void assignReservedIdOnNestedCreateRequest() throws Exception {
```
cmark: I would add this request to the `SnomedConceptApiTest` class and use the REST API for testing the functionality.

cmark (changes requested):

cmark:
@apeteri please take a look again and review this PR. 🙇

apeteri (approved):
LGTM!

cmark (approved):

PR #947, opened 2021-11-26
fix(index): incorrect branch segment diff algorithm causing false positive conflicts

AAAlinaaa:


commons/com.b2international.index/src/com/b2international/index/revision/RevisionBranchRef.java
```diff
@@ -129,29 +128,25 @@ public Expression toRevisedInFilter() {
...
 				differenceSegments.add(difference);
 			}
 		}
 		
-		while (segmentsIterator.hasNext()) {
-			RevisionSegment nextSegment = segmentsIterator.next();
-			if (!nextSegment.isEmpty()) {
-				differenceSegments.add(nextSegment);
+		// add any remaining segments if they are not empty
+		for (Long branchId : Sets.difference(segmentsById.keySet(), otherSegmentsById.keySet())) {
```
apeteri, line 143: Do we need `Sets#symmetricDifference` here (segments on _either side_ that are not present on the other), or just the ones which are in `segmentsById` but not in `otherSegmentsById`?

apeteri (changes requested):

commons/com.b2international.index/src/com/b2international/index/revision/RevisionBranchRef.java
cmark: No, this method should always return the diff from the current `this` RevisionBranchRef, similar to the `Sets.difference` method.

apeteri (approved):

PR #948, opened 2021-12-02
feat(snomed): defer loading of expanded additional data...

cmark:
...to dedupe requests going to Elasticsearch and reduce load on both the
ES cluster and the termserver.

This effectively improves search performance of concepts with
additionally fetched concept data on multiple components (using the expand API), including
descriptions, relationships, members (deduped expanded fields are:
`module`, `type`, `source`, `destination`, `caseSignificance`,
`modifier`, `definitionStatus`, `languageRefSet` and others).

### DIFF
--- a/snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/converter/ModuleExpander.java
+++ b/snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/converter/ModuleExpander.java
@@ -16,14 +16,12 @@
 package com.b2international.snowowl.snomed.datastore.converter;
 
 import java.util.List;
-import java.util.SortedSet;
 
 import com.b2international.commons.http.ExtendedLocale;
 import com.b2international.commons.options.Options;
 import com.b2international.snowowl.core.domain.BranchContext;
 import com.b2international.snowowl.snomed.core.domain.SnomedComponent;
 import com.b2international.snowowl.snomed.datastore.request.SnomedConceptRequestCache;
-import com.google.common.collect.ImmutableSortedSet;
 
 /**
  * @since 7.4
@@ -44,7 +42,7 @@ final class ModuleExpander {
 		if (expand.containsKey(SnomedComponent.Expand.MODULE)) {
 			final Options moduleOptions = expand.get(SnomedComponent.Expand.MODULE, Options.class);
 
-			final SortedSet<String> moduleIds = results.stream().map(SnomedComponent::getModuleId).collect(ImmutableSortedSet.toImmutableSortedSet(String::compareTo));
+			final Iterable<String> moduleIds = results.stream().map(SnomedComponent::getModuleId)::iterator;
 			
 			context.service(SnomedConceptRequestCache.class)
 				.request(moduleIds, moduleOptions.getOptions("expand"), locales, modulesById -> {
--- a/snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/converter/ModuleExpander.java
+++ b/snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/converter/ModuleExpander.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2020 B2i Healthcare Pte Ltd, http://b2i.sg
+ * Copyright 2020-2021 B2i Healthcare Pte Ltd, http://b2i.sg
  * 
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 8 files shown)


cmark:
The performance gain is almost **70%**.

Without request deduping and caching (current 8.x code):
Avg of three fully expanded requests: `815ms`
Fourth fully expanded ES search request cached request: `57ms`

With request deduping and caching (this PR):
Avg of three fully expanded requests: `250ms`
Fourth fully expanded ES search request cached request: `40ms`

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/converter/ModuleExpander.java
```diff
@@ -46,21 +44,15 @@ void expand(List<? extends SnomedComponent> results) {
 		if (expand.containsKey(SnomedComponent.Expand.MODULE)) {
 			final Options moduleOptions = expand.get(SnomedComponent.Expand.MODULE, Options.class);
 
-			final Set<String> moduleIds = results.stream().map(SnomedComponent::getModuleId).collect(Collectors.toSet());
+			final SortedSet<String> moduleIds = results.stream().map(SnomedComponent::getModuleId).collect(ImmutableSortedSet.toImmutableSortedSet(String::compareTo));
```
apeteri: This one could also be an `Iterable` for consistency, since `SnomedConceptRequestCache` takes care of making the input unique.

In the converter classes below, `Iterable`s and `Stream`s are about equal in numbers as far as I can tell; would you like to harmonize those into a One True Way of collecting concept IDs off components? 😄 Then whichever type is chosen, that could be used as the first parameter in `SnomedConceptRequestCache#request`.

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/request/SnomedConceptRequestCache.java
```diff
@@ -0,0 +1,124 @@
...
+		requestedFetches.add(new FetchConfig(ids, expand, locales, onConceptsReady));
+	}
+	
+	public Map<String, SnomedConcept> get(BranchContext context, Iterable<String> ids, Options expand, List<ExtendedLocale> locales) {
+		FetchConfig config = new FetchConfig(ids, expand, locales, concepts -> {});
+		if (cache.containsKey(config)) {
+			return cache.get(config);
+		} else {
+			requestedFetches.add(config);
+			compute(context);
```
apeteri: Would it make sense to restrict computation to this config or configs with the same `expand` and `locale` parameters only, and not compute all pending requests?

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/request/SnomedConceptRequestCache.java
```diff
@@ -0,0 +1,124 @@
...
+			
+			// populate the cache for each fetch config and call the callback
+			configsToFetch.forEach(cfg -> {
+				ImmutableMap.Builder<String, SnomedConcept> configConcepts = ImmutableMap.builder();
+				cfg.ids.forEach(idToCache -> {
+					configConcepts.put(idToCache, fetchedConcepts.get(idToCache));
+				});
+				ImmutableMap<String, SnomedConcept> concepts = configConcepts.build();
+				cache.put(cfg, concepts);
+				cfg.onConceptsReady.accept(concepts);
```
apeteri: Callbacks could also use `fetchedConcepts` directly, which contains a superset of the concepts they have originally requested. Then you don't need to build a map for each consumer.

apeteri (changes requested):

nagyo (approved):
LGTM!

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/request/SnomedConceptRequestCache.java
cmark: That was the initial implementation 😄 I can make an alternative, where there is a limit to the computed configs.

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/request/SnomedConceptRequestCache.java
cmark: Yeah, thought about that, it would work in certain cases where we only use the map by requested concepts via their ID from it, but in case you iterate over all of the entries of the returned map, this would give you inconsistent results. `config.ids` and `map.keys` should be equal.

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/request/SnomedConceptRequestCache.java
cmark: Done

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/converter/ModuleExpander.java
cmark: Done

apeteri (approved):

nagyo (approved):
👍🏻

PR #949, opened 2021-12-07
Add missing expand and field selector to relationship GET by ID endpoint

nagyo:
- add unit test for simple expand use-case
- add missing accept language tag to description GET by ID endpoint
- rephrase description of status and module filters

### DIFF
--- a/snomed/com.b2international.snowowl.snomed.core.rest.tests/src/com/b2international/snowowl/snomed/core/rest/components/SnomedRelationshipApiTest.java
+++ b/snomed/com.b2international.snowowl.snomed.core.rest.tests/src/com/b2international/snowowl/snomed/core/rest/components/SnomedRelationshipApiTest.java
@@ -163,15 +163,10 @@ public class SnomedRelationshipApiTest extends AbstractSnomedApiTest {
 			.statusCode(200)
 			.extract().as(SnomedRelationship.class);
 		
-		assertNotNull(relationship.getCharacteristicType());
 		assertEquals(Concepts.INFERRED_RELATIONSHIP, relationship.getCharacteristicType().getId());
-		assertNotNull(relationship.getSource());
 		assertEquals(Concepts.ROOT_CONCEPT, relationship.getSource().getId());
-		assertNotNull(relationship.getType());
 		assertEquals(Concepts.PART_OF, relationship.getType().getId());
-		assertNotNull(relationship.getDestination());
 		assertEquals(Concepts.NAMESPACE_ROOT, relationship.getDestination().getId());
-		assertNotNull(relationship.getModifier());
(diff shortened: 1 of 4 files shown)


snomed/com.b2international.snowowl.snomed.core.rest.tests/src/com/b2international/snowowl/snomed/core/rest/components/SnomedRelationshipApiTest.java
```diff
@@ -152,13 +153,27 @@ public void createDuplicateRelationship() {
...
+			.extract().as(SnomedRelationship.class);
+		
+		assertNotNull(relationship.getCharacteristicType());
+		assertEquals(Concepts.INFERRED_RELATIONSHIP, relationship.getCharacteristicType().getId());
+		assertNotNull(relationship.getSource());
+		assertEquals(Concepts.ROOT_CONCEPT, relationship.getSource().getId());
+		assertNotNull(relationship.getType());
+		assertEquals(Concepts.PART_OF, relationship.getType().getId());
+		assertNotNull(relationship.getDestination());
+		assertEquals(Concepts.NAMESPACE_ROOT, relationship.getDestination().getId());
```
apeteri: For concept-valued properties, a minimal `SnomedConcept` instance is always present, even without expanding (the example is for `setDestinationId` but other setters are also put together this way): https://github.com/b2ihealthcare/snow-owl/blob/2db98207d4ab37615c3fcabfe25dfddd22897be6/snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/core/domain/SnomedRelationship.java#L299-L302

 I'd recommend adding a non-null check for eg. the `active` status or `moduleId` property for each one to be sure that they were fully populated with information.

apeteri (approved):
👍

snomed/com.b2international.snowowl.snomed.core.rest.tests/src/com/b2international/snowowl/snomed/core/rest/components/SnomedRelationshipApiTest.java
cmark: I would remove the non-null checks completely, as they don't add that much to the test itself.

cmark (approved):
LGTM!

snomed/com.b2international.snowowl.snomed.core.rest.tests/src/com/b2international/snowowl/snomed/core/rest/components/SnomedRelationshipApiTest.java
nagyo: Old habits of getPt() throwing NPE tricked me 😊, removed the null checks 🎉

PR #955, opened 2021-12-16
[6.x] add missing log4j2 system properties to embedded mode

cmark:
While neither Snow Owl nor the underlying Elasticsearch is affected by
the CVE-2021-44228 security vulnerability, to mitigate any future risk
around this Snow Owl will now require a minimum of 6.8.21 Elasticsearch
version in both embedded and external modes (Snow Owl 6.x).
Both the `log4j2.formatMsgNoLookups` system property is ensured and an
altered version of the `elasticsearch-log4j.jar` is used that removes
the JNDI lookup feature from log4j2.


nagyo (approved):
LGTM!

PR #959, opened 2022-01-05
Filter out hidden indices when checking cluster health (7.x)

nagyo:


apeteri (approved):

apeteri:
The corresponding PR on 8.x that includes a change compatible with the addition of `lenientExpand` done here is https://github.com/b2ihealthcare/snow-owl/pull/1053:

https://github.com/b2ihealthcare/snow-owl/blob/8d86b1e163d3f5bcb0c958316bc8b3d04a0153a9/commons/com.b2international.index/src/com/b2international/index/es/client/EsClientBase.java#L63-L72

PR #961, opened 2022-01-07
refactor(core): Clarify that resources store bundle ancestor IDs...

apeteri:
...in order of depth, ie. a bundle path/absolute path can be computed by
concatenating all ancestor IDs, the container bundle ID and optionally,
the resource ID.

The only exception are resources that reside in the resource root, where
both the container bundle ID is set to -1 and the ancestor ID array
contains a single element with -1 as the value.

Additionally, the descendant updating step in BaseResourceUpdateRequest
was refactored to process siblings in a single iteration, and reduce
memory usage slightly.


apeteri:
~Let me know if I should add a derived property that puts together the computed path! I think it is better to compile this information on the consumer side, as I can imagine lots of variations (sometimes only the bundle IDs leading to the resource are needed, other times the resource ID should be appended to the end, path separators may differ, etc.)~

Bundle path segments and an expand option for labels were added to `Resource`.

cmark (approved):
LGTM!

PR #962, opened 2022-01-10
feat(core): Introduce streaming helper method that uses the...

apeteri:
...ServiceProvider of the currently executing request.

The second argument is an AsyncRequest-building function, which can be
used for extracting a fully wrapped request suitable for a
ServiceProvider context. This can be useful when eg. a repository or
branch needs to be targeted that differs from the context of the current
request.


cmark (approved):
LGTM!

PR #965, opened 2022-01-13
SO-4985: Update extension namespace-module assignment logic

apeteri:
The current code system is derived from the `BranchContext` path, and extension module IDs are retrieved from code system metadata. 

For all inferred or redundant components where the source/referenced concept's module is not in the code system's module list, the module ID will be set to `defaultModuleId`, otherwise the source concept's module will be used. 

The same logic applies to the namespace portion of relationship SCTIDs as well (note that the decision is still made based on the source concept's *module*, not its namespace).


cmark (approved):
LGTM!

apeteri:
On 8.x, commit https://github.com/b2ihealthcare/snow-owl/commit/53288636a4b260746edb1e7ac0692c2b49e91383 makes semi-compatible changes to the module-namespace assigner infrastructure. Leaving the label up, but this PR might end up unused.

PR #966, opened 2022-01-19
feat(api): support expansion of `commits()` in Resource API

cmark:
This PR introduces a new expand parameter to the Resource API, called `commits` to support single HTTP calls when listing resources along with their relevant commits.

### DIFF
--- a/core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/request/ResourceConverter.java
+++ b/core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/request/ResourceConverter.java
@@ -18,18 +18,20 @@ package com.b2international.snowowl.core.request;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
 
 import com.b2international.commons.http.ExtendedLocale;
 import com.b2international.commons.options.Options;
-import com.b2international.snowowl.core.Resource;
-import com.b2international.snowowl.core.ResourceTypeConverter;
-import com.b2international.snowowl.core.Resources;
-import com.b2international.snowowl.core.TerminologyResource;
+import com.b2international.snowowl.core.*;
+import com.b2international.snowowl.core.commit.CommitInfo;
 import com.b2international.snowowl.core.domain.IComponent;
 import com.b2international.snowowl.core.domain.RepositoryContext;
+import com.b2international.snowowl.core.events.util.Promise;
 import com.b2international.snowowl.core.internal.ResourceDocument;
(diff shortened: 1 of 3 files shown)


core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/request/ResourceConverter.java
```diff
@@ -124,4 +127,32 @@ private void expandVersions(List<Resource> results) {
...
+			// commit searches must be performed individually on each resource to provide correct results
+			var commitSearchRequests = results.stream()
+				.filter(TerminologyResource.class::isInstance)
+				.map(TerminologyResource.class::cast)
+				.map(res -> {
+					return RepositoryRequests.commitInfos().prepareSearchCommitInfo()
+						.filterByBranch(res.getBranchPath())
+						.setLimit(getLimit(expandOptions))
+						.setFields(expandOptions.containsKey("fields") ? expandOptions.getList("fields", String.class) : CommitInfo.Fields.DEAFULT_FIELD_SELECTION)
+						.sortBy(expandOptions.containsKey("sort") ? expandOptions.getString("sort") : null)
```
apeteri, line 142: It's OK to leave it like this for better readability, but `expandOptions.getString("sort")` would end up doing the same thing.

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/request/ResourceConverter.java
```diff
@@ -124,4 +127,32 @@ private void expandVersions(List<Resource> results) {
...
+						.setLocales(locales())
+						.build(res.getToolingId())
+						.execute(context().service(IEventBus.class))
```
apeteri, line 145: Do we want to run all commit search requests in parallel? Maybe a `filterByBranches` method could be added later, accepting an `Iterable<String>` of branch paths, so we end up with a single request per tooling.

Is the `commits()` expansion expected to be used in single terminology resource ("get") scenarios right now?

apeteri (approved):
LGTM!

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/request/ResourceConverter.java
cmark: It is expected to be used when you perform a search request or a single get by ID request. Both use cases should be covered.

cmark: To properly expand commits with sorting options (like `timestamp:desc`) we need to expand them parallel for each resource.

PR #967, opened 2022-01-20
feat(core): support resource expansion via plugins

cmark:


core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/request/ResourceExpander.java
```diff
@@ -15,47 +15,41 @@
...
 		return expandOptions.containsKey("limit") ? expandOptions.get("limit", Integer.class) : DEFAULT_LIMIT;
 	}
-
+	
+	/**
+	 * @return the class of the target type this expander can expand
+	 */
+	@SuppressWarnings("unchecked")
+	default Class<R> getType() {
+		final Class<?>[] types = TypeResolver.resolveRawArguments(BaseResourceConverter.class, getClass());
```
apeteri: Since the type parameter has been lifted up to this interface: if typetools can check implemented interfaces as well, not just the base class, it should do so.

```suggestion
		final Class<?>[] types = TypeResolver.resolveRawArguments(ResourceExpander.class, getClass());
```

apeteri (approved):
👍

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/request/ResourceExpander.java
apeteri: LGTM might be complaining about the javadoc tag not looking like this: `@param <R>`

PR #968, opened 2022-01-20
SO-4984: Add CUD support to FHIR codesystems

AAAlinaaa:


apeteri:
LGTM is complaining about this Javadoc comment: https://github.com/b2ihealthcare/snow-owl/blob/31618cdcc13f1a346b3efc5478372bdd98b5463a/fhir/com.b2international.snowowl.fhir.rest/src/com/b2international/snowowl/fhir/rest/FhirCodeSystemController.java#L58

https://github.com/b2ihealthcare/snow-owl/blob/31618cdcc13f1a346b3efc5478372bdd98b5463a/fhir/com.b2international.snowowl.fhir.rest/src/com/b2international/snowowl/fhir/rest/FhirCodeSystemController.java#L74

Please update it so the character case matches the formal parameter of the method!

fhir/com.b2international.snowowl.fhir.core/src/com/b2international/snowowl/fhir/core/model/dt/Uri.java
```diff
@@ -53,6 +54,14 @@ public String getUriValue() {
 	public boolean isSnomedUri() {
 		return uriValue != null && uriValue.startsWith(SNOMED_BASE_URI_STRING);
 	}
+	
+	public boolean isOid() {
+		return uriValue != null && uriValue.startsWith(OID_SUBSTRING);
+	}
+	
+	public String getOid() {
+		return isOid() ? uriValue.replaceAll(OID_SUBSTRING, "") : "";
```
apeteri: It would be better to remove `OID_SUBSTRING.length()` worth of characters from the beginning of the string with `substring` instead of using regex to replace all occurrences via `replaceAll` here.

apeteri: ⚠️ Copyright date needs adjustment on this file

fhir/com.b2international.snowowl.fhir.core/src/com/b2international/snowowl/fhir/core/request/FhirCodeSystemRequests.java
```diff
@@ -42,4 +48,8 @@ public FhirValidateCodeRequestBuilder prepareValidateCode() {
 		return new FhirValidateCodeRequestBuilder();
 	}
 	
+	public ResourceDeleteRequestBuilder prepareDelete(String codeSystemId) {
+		return new ResourceDeleteRequestBuilder(codeSystemId, ResourceDocument.class);
+	}
```
apeteri, line 53: `ResourceRequests#prepareDelete` makes this request builder redundant, I think 🤔

fhir/com.b2international.snowowl.fhir.core/src/com/b2international/snowowl/fhir/core/request/codesystem/FhirCodeSystemPutRequest.java
```diff
@@ -0,0 +1,57 @@
...
+	
+	private final CodeSystem codeSystem;
+	
+	public FhirCodeSystemPutRequest(CodeSystem codeSystem) {
+		this.codeSystem = codeSystem;
+	}
+
+	@Override
+	public Boolean execute(RepositoryContext context) {
+		checkArgument(codeSystem.getConcepts() == null || codeSystem.getConcepts().size() < CONCEPT_LIMIT, "Maintanence of code systems with more than %d codes is not supported.", CONCEPT_LIMIT);
```
apeteri: ⚠️ Typo

```suggestion
		checkArgument(codeSystem.getConcepts() == null || codeSystem.getConcepts().size() < CONCEPT_LIMIT, "Maintenance of code systems with more than %d codes is not supported.", CONCEPT_LIMIT);
```

fhir/com.b2international.snowowl.fhir.core/src/com/b2international/snowowl/fhir/core/request/codesystem/FhirCodeSystemPutRequest.java
```diff
@@ -0,0 +1,57 @@
...
+		
+		FhirCodeSystemCUDSupport cudSupport =	context.service(RepositoryManager.class)
+				.repositories()
+				.stream()
+				.filter(r -> r.optionalService(FhirCodeSystemCUDSupport.class).isPresent())
+				.map(r -> r.service(FhirCodeSystemCUDSupport.class))
+				.collect(Collectors.toList())
+				.get(0);
```
apeteri, line 51: Some sort of routing is needed here; with the current implementation, you assign the task to the first repository that claims to support code system creation/updates. See `FhirCodeSystemSearchRequest` for an example how to handle this:

```java
FhirCodeSystemResourceConverter converter = context.service(RepositoryManager.class)
  .get(resource.getToolingId())
  .optionalService(FhirCodeSystemResourceConverter.class)
  .orElse(FhirCodeSystemResourceConverter.DEFAULT);
```

You should be able to grab the tooling ID from the `CodeSystem` in the request.

The `orElse` part of the example is not applicable in this case since there is no default implementation &ndash; an exception should be thrown instead if the tooling does not have a service that implements `FhirCodeSystemCUDSupport`.

fhir/com.b2international.snowowl.fhir.rest/src/com/b2international/snowowl/fhir/rest/FhirCodeSystemController.java
```diff
@@ -237,4 +158,45 @@
...
+			final String id,
+			
+			@Parameter(description = "Force deletion flag")
+			@RequestParam(defaultValue="false", required=false)
+			final Boolean force,
+
+			@RequestHeader(value = X_AUTHOR, required = true)
+			final String author) {
+		try {
+			FhirRequests.codeSystems().prepareDelete(id)
```
apeteri, line 188:
```suggestion
			ResourceRequests.prepareDelete(id)
```

apeteri (changes requested):

fhir/com.b2international.snowowl.fhir.core/src/com/b2international/snowowl/fhir/core/request/codesystem/FhirCodeSystemPutRequest.java
AAAlinaaa: 1. Unfortunately we can't count on the tooling id being present on the codesystem, especially not in the format expected by Snow Owl
2. In previous discussions on the topic we agreed there will only be one supporting terminology, and to use the first (and only) implementor

fhir/com.b2international.snowowl.fhir.core/src/com/b2international/snowowl/fhir/core/request/FhirCodeSystemRequests.java
apeteri: We have agreed to keep this in place, so that all operations are reachable from the same request builder factory.

apeteri (approved):
LGTM!

PR #971, opened 2022-01-27
fix(auth): prevent access to resources when permissions claim is empty..

cmark:
...for the current user or in the given JWT accessToken


apeteri (approved):
LGTM.

PR #976, opened 2022-02-11
SO-4080: optimize bool index queries

cmark:
This PR optimizes bool index queries to reduce (and hopefully eliminate) the chance of hitting search errors:

1. Merge term filters specified on the same field to reduce the number of bool query clauses
2. Flatten deep bool query hierarchies to reduce the chance of getting HTTP 400 error due to `indices.query.bool.max_nested_depth` setting, which defaults to depth of `20`.

### DIFF
--- a/commons/com.b2international.index/src/com/b2international/index/es/EsDocumentSearcher.java
+++ b/commons/com.b2international.index/src/com/b2international/index/es/EsDocumentSearcher.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017-2021 B2i Healthcare Pte Ltd, http://b2i.sg
+ * Copyright 2017-2022 B2i Healthcare Pte Ltd, http://b2i.sg
  * 
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -28,11 +28,13 @@ import java.util.*;
 import org.apache.lucene.search.TotalHits;
 import org.apache.lucene.search.TotalHits.Relation;
 import org.apache.solr.common.util.JavaBinCodec;
+import org.elasticsearch.ElasticsearchStatusException;
 import org.elasticsearch.action.get.GetRequest;
 import org.elasticsearch.action.get.GetResponse;
 import org.elasticsearch.action.search.SearchRequest;
 import org.elasticsearch.action.search.SearchResponse;
 import org.elasticsearch.index.query.QueryBuilder;
+import org.elasticsearch.rest.RestStatus;
 import org.elasticsearch.search.SearchHit;
 import org.elasticsearch.search.SearchHits;
 import org.elasticsearch.search.aggregations.AggregationBuilders;
@@ -183,6 +185,9 @@ public class EsDocumentSearcher implements Searcher {
 		try {
 			response = client.search(req);
 		} catch (Exception e) {
+			if (e instanceof ElasticsearchStatusException && ((ElasticsearchStatusException) e).status() == RestStatus.BAD_REQUEST) {
+				throw new IllegalArgumentException(e.getMessage(), e);
+			}
 			admin.log().error("Couldn't execute query", e);
 			throw new IndexException("Couldn't execute query: " + e.getMessage(), null);
 		}
--- a/commons/com.b2international.index/src/com/b2international/index/query/AbstractExpressionBuilder.java
+++ b/commons/com.b2international.index/src/com/b2international/index/query/AbstractExpressionBuilder.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2011-2017 B2i Healthcare Pte Ltd, http://b2i.sg
+ * Copyright 2011-2022 B2i Healthcare Pte Ltd, http://b2i.sg
  * 
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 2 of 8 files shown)


commons/com.b2international.index/src/com/b2international/index/es/EsDocumentSearcher.java
```diff
@@ -183,6 +185,9 @@ public EsDocumentSearcher(EsIndexAdmin admin, ObjectMapper mapper) {
 		try {
 			response = client.search(req);
 		} catch (Exception e) {
+			if (e instanceof ElasticsearchStatusException && ((ElasticsearchStatusException) e).status() == RestStatus.BAD_REQUEST) {
+				throw new IllegalArgumentException(e.getMessage(), e);
+			}
```
apeteri, line 190: Please add this block to a separate `catch (ElasticSearchStatusException e) { ... }` clause above the existing "catch-all" block.

commons/com.b2international.index/src/com/b2international/index/query/AbstractExpressionBuilder.java
```diff
@@ -72,13 +78,90 @@ public B setMinimumNumberShouldMatch(int minShouldMatch) {
 	public Expression build() {
 		if (mustClauses.isEmpty() && mustNotClauses.isEmpty() && shouldClauses.isEmpty() && filterClauses.isEmpty()) {
 			return Expressions.matchAll();
-		} else if (mustClauses.isEmpty() && mustNotClauses.isEmpty() && shouldClauses.isEmpty() && filterClauses.size() == 1) {
+		} else if (isSingleFilter()) {
 			// shortcut to reduce number of nested Boolean clauses
 			return filterClauses.get(0);
```
apeteri, line 82: We could support `isSingleShould()` here as well, as you write in the final `else` block below (single "shoulds" _must_ match). 

Also if we want to be really precise, the extracted single filter clause should be wrapped in a non-scoring query wrapper to achieve the same effect as a boolean filter &ndash; which is not contributing to the final document score.

```suggestion
		} else if (isSingleFilter()) {
			// shortcut to reduce number of nested Boolean clauses
			return filterClauses.get(0);
			// return new NoScoreExpression(filterClauses.get(0)); <-- this class does not exist yet
		} else if (isSingleShould()) {
			return filterClauses.get(0);
```

commons/com.b2international.index/src/com/b2international/index/query/AbstractExpressionBuilder.java
```diff
@@ -72,13 +78,90 @@ public B setMinimumNumberShouldMatch(int minShouldMatch) {
...
+
+	protected final void flattenShoulds() {
+		if (!mustClauses.isEmpty() || !mustNotClauses.isEmpty() || !filterClauses.isEmpty() || minShouldMatch != 1) {
+			return;
+		}
+		for (Expression expression : List.copyOf(shouldClauses)) {
+			if (expression instanceof BoolExpression) {
+				BoolExpression bool = (BoolExpression) expression;
+				if (bool.isShouldOnly() && bool.minShouldMatch() == 1) {
+					shouldClauses.addAll(bool.filterClauses());
```
apeteri:
```suggestion
					shouldClauses.addAll(bool.shouldClauses());
```

apeteri: (I understand that starting with ES 8.0 we don't have a limit on the number of "should"-clauses. Don't we want to restrain ourselves regardless? 😃 So `addAll` would only be called if the resulting clause list remains under the previous 1024 limit.)

commons/com.b2international.index/src/com/b2international/index/query/AbstractExpressionBuilder.java
```diff
@@ -72,13 +78,90 @@ public B setMinimumNumberShouldMatch(int minShouldMatch) {
...
+				for (Expression expression : termExpressions) {
+					if (expression instanceof SingleArgumentPredicate<?>) {
+						values.add(((SingleArgumentPredicate<?>) expression).getArgument());
+					} else if (expression instanceof SetPredicate<?>) {
+						values.addAll(((SetPredicate<?>) expression).values());
+					}
+				}
+				// replace all clauses with a single expression
+				clauses.add(Expressions.matchAnyObject(field, values));
+				clauses.removeAll(termExpressions);
```
apeteri, line 207: Hopefully the `equals` + `hashCode` implementations of `SingleArgumentPredicate` and `SetPredicate` do not interfere with this remove operation. On another note, they could be used to remove duplicate expressions!

apeteri (changes requested):

commons/com.b2international.index/src/com/b2international/index/es/EsDocumentSearcher.java
cmark: Leaving as is now, as the current catch-all block handles the logging and generic index exception in case of any other ES exception thrown by the search request.

apeteri: Following discussion with @cmark, non-400 errors should still be logged, so the `catch` block can be left as is.

commons/com.b2international.index/src/com/b2international/index/query/AbstractExpressionBuilder.java
cmark: Yeah, I'll swap the remove and add call order to remove the old expressions first then add the new one, it can happen that the new expression is the same as one of the removed ones and in that case, the resulting boolean expression will be empty, which is wrong.

commons/com.b2international.index/src/com/b2international/index/query/AbstractExpressionBuilder.java
cmark: Not sure, I would rely on the ES to tell us if a query is too big to execute, (has too deep bool clause structure or too many clauses or whatevs). We can monitor the executed queries and add this later on if needed.

apeteri (approved):
LGTM!

PR #981, opened 2022-02-21
Add generic taxonomy change processor

apeteri:
The PR also removes `LongBidiMapWithInternalId`, which previously provided an SCTID 🡘 long storage key mapping in addition to an integer index. Since it was only used for an SCTID 🡘 integer index mapping, it is replaced with a `LongOrderedSet` implementation.


commons/com.b2international.commons/src/com/b2international/commons/collect/LongOrderedSet.java
```diff
@@ -0,0 +1,65 @@
...
+ * limitations under the License.
+ */
+package com.b2international.commons.collect;
+
+import com.b2international.collections.longs.LongSet;
+
+/**
+ * A set storing primitive long values that preserves insertion order.  
+ * 
+ * @since 8.0.1
```
cmark, line 23: Change since tag to 8.1.0

cmark (approved):
Looks really impressive @apeteri! I'm going to digest the SimpleTaxonomyGraph and ChangeProcessor classes tomorrow.

cmark (approved):
LGTM!

PR #982, opened 2022-02-22
fix(fhir): Validate system URL against version given in subsumption request

apeteri:
Re-enabled test case where the system URL has a version suffix, but it does not match the version parameter of the request.


cmark (approved):
LGTM!

PR #990, opened 2022-03-02
SO-5362: Support retrieval of historical state for resources

apeteri:
Setting the `timestamp` parameter will return content that represents the state of each resource at the specified point in time.


cmark (approved):
LGTM!

PR #1001, opened 2022-04-01
Wrap Groovy 3.0.9 and extensions into a bundle

apeteri:
While all JARs include an OSGi-compatible manifest, extensions such as XML, JSON and text template helper classes were converted to bundle fragments by the authors.

This causes problems when we want to refer to eg. an XML builder class in a script, as we can only require the "core" Groovy bundle to be present, we can't ask for a fragment to be added to the classpath.

The pull request wraps Groovy into a single bundle that can be referenced in consuming bundles (similar to the earlier bundle provided by Codehaus for 2.4).


cmark (approved):
LGTM!

PR #1002, opened 2022-04-05
fix(api): commits endpoint authorization now correctly restricts...

cmark:
...the responses based on the incoming access token permission list
(user info).

If resources are not available to the user, then commit endpoints will
return empty results.
If a resource is available to the user, then it always returns the main
commits of that resource and depending on additional permissions it can
be return commits made on child branches.

Supported permission formats:
 * RESOURCE_ID
 * RESOURCE_ID*
 * RESOURCE_ID/*
 * RESOURCE_ID/RELATIVE_BRANCH


apeteri (approved):
👍

PR #1006, opened 2022-04-12
chore(commons): Update H2 from 1.4.199 to 2.1.210

apeteri:
Existing scripts are using Groovy's built-in SQL facilities so H2 only serves as a JDBC driver for an embedded database. We no longer rely on its low-level functionality.


apeteri:
Related to https://github.com/b2ihealthcare/snow-owl/pull/969 which can also be merged.

cmark (approved):
LGTM!

PR #1007, opened 2022-04-13
SO-5406 freeze retired resources

cmark:
This PR ensures that retired resources cannot be versioned and their contents cannot be edited unless their status changes back to something else than `retired`.


apeteri (approved):
👍

PR #1012, opened 2022-04-26
SO-5412: make upload and download buffer size configurable (8.x)

nagyo:
- set default upload buffer size to 10 Mb
- set default download buffer size to 1 Mb

When large enough files were being uploaded with a small upload buffer
size, promise evaluation halted the upload process possibly of the
reason: the number of upload chunks is equal to the number of promises
wrapped into each other in AttachmentRegistryClient.upload()


apeteri (approved):
LGTM!

PR #1015, opened 2022-05-17
fix(index): Defer "must" and "filter" clause reduction to a later...

apeteri:
...point, when document mapping is available. This allows us to discern between collection and scalar fields; the former can take up multiple values at the same time and as such this optimization step is not applicable for them.


cmark (approved):
LGTM!

PR #1021, opened 2022-05-20
feat(eventbus): Enable call-time stack trace recording with JVM option

apeteri:
Adding `-Deventbus.record.stack=true` to the JVM options will add a header with the call-time stack trace for each sent message, even if it originates on another machine. Responding with an exception prints the stack to the standard error output. This can help in diagnosing issues related to distributed messaging.


cmark (approved):
LGTM!

PR #1022, opened 2022-05-20
fix(core): Make derived field on User transient

apeteri:
The corresponding property is already &commat; JsonIgnore-d; it needs to be absent from the binary serialization as well.


cmark (approved):
LGTM!

PR #1026, opened 2022-05-27
Authorization API improvements

cmark:
This PR improves the available authorization APIs to support third-party extensions and configuration.

It also 
* removes support for the old toolingId/repositoryId:branch resource format in the permission model.
* removes Role representation of a User
* renames the `User.username` property to `userId` to better reflect its purpose
* moves `testcontainers` dependencies to test.dependencies to let any test plugin to use them


core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/identity/IdentityProvider.java
```diff
@@ -19,11 +19,12 @@
 import java.util.List;
 import java.util.stream.Collectors;
 
+import javax.management.relation.Role;
```
apeteri: This import statement probably took the place of the original `Role` class during the remove.

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/identity/User.java
```diff
@@ -20,54 +20,54 @@
...
 	public List<Permission> getPermissions() {
-		return permissions.get();
+		return permissions;
+	}
+	
+	public void setAuthorizationContext(Map<String, Object> authorizationContext) {
+		this.authorizationContext = authorizationContext == null ? Collections.emptyMap() : Map.copyOf(authorizationContext);
+	}
+	
+	public Map<String, Object> getAuthorizationContext() {
```
apeteri, line 66: Do you want this property to appear in JSON serialized form? I can not tell right of the bat if it can ever happen that it is populated and returned where it is visible externally, but just to be on the safe side, a &commat; `JsonIgnore` might be warranted here.

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/request/version/VersionCreateRequest.java
```diff
@@ -317,19 +317,13 @@ private void doTag(ServiceProvider context, TerminologyResource codeSystem, IPro
 	
 	@Override
 	public List<Permission> getPermissions(ServiceProvider context, Request<ServiceProvider, ?> req) {
-		if (resourcesById == null) {
-			resourcesById = fetchResources(context);
-		}
-		List<Permission> permissions = new ArrayList<>(resourcesById.size());
-		for (ResourceURI accessedResourceURI : resourcesById.keySet()) {
```
apeteri, line 324: How come this step can be skipped in the new form? It looks like there is a similar bit of code in `execute`, but it is also only looking at a single resource: https://github.com/b2ihealthcare/snow-owl/blob/d0e037d88f8e0a3a3593fef529fa0639723d6ac9/core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/request/version/VersionCreateRequest.java#L109-L132

apeteri (approved):
LGTM!

Some classes were only changed by automated refactoring steps; the copyright date should be updated on them as well:

- CisBulkSctIdService
- RepositoryApiTest
- RepositoryRestService
- AbstractBranchChangeRequest
- CodeSystemUpgradeSynchronizationRequest
- CommandLineAuthenticator
- DefaultJWTGenerator
- RemoteJob
- ScheduleJobRequest
- DefaultOperationLockManager
- Locks
- LocksCommand
- RateLimitingRequest
- RepositoryTransactionContext
- SnomedBranchRequestTest
- ClassificationCreateRequest
- ClassificationSaveRequest
- SaveJobRequest
- BundleApiAssert
- RepositoryApiAssert

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/identity/User.java
cmark: No, this is an internal property, will add the `JsonIgnore` annotation.

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/request/version/VersionCreateRequest.java
cmark: The only resource we can version now is the one that the user specifies in the Version Create request body. Multi-resource versioning is not possible at the moment and it won't be for a while. I can eliminate the Map-like handling of resources and replace it with a single transient Resource field, if that's better.

PR #1029, opened 2022-06-02
SO-5443: Add support for new refset type; simple map to SNOMEDCT

AAAlinaaa:


snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/core/domain/refset/SnomedRefSetType.java
```diff
@@ -34,8 +34,16 @@ public enum SnomedRefSetType {
...
+	 * @deprecated
 	 */
 	SIMPLE_MAP(1, "SIMPLE_MAP", "SIMPLE_MAP"),
+	
+	/**
+	 * Simple map from SNOMEDCT type. 
+	 * @see #SIMPLE_MAP_FROM_SNOMEDCT_VALUE
+	 * @ordered
+	 */
+	SIMPLE_MAP_FROM_SNOMEDCT(1, "SIMPLE_MAP_FROM_SNOMEDCT", "SIMPLE_MAP_FROM_SNOMEDCT"),
```
cmark: I think it would be great to change this enum to a basic enum literal instead of using the old EMF style format. Also, the index here is incorrect, please move it to the end of the list.

cmark (changes requested):

cmark (approved):
LGTM!

cmark:
@AAAlinaaa Make sure you port this to 8.x as well. Thanks!

apeteri:
The 8.x counterpart PR is https://github.com/b2ihealthcare/snow-owl/pull/1043 which introduces broader support for the new reference set type (including read/write capability). I'm leaving the `port-to-8.x` label up as the constant for the original "from SNOMED CT"-type concept was not renamed on 8.x &ndash; we might want to harmonize names across the two branches to avoid merge conflicts in the future.

PR #1043, opened 2022-07-13
SO-5498: Add support for "Simple map to SNOMED CT" reference sets

apeteri:


cmark (approved):
This looks really great, thanks @apeteri!

PR #1044, opened 2022-07-19
INFRA-286: Update docker files (8.x)

nagyo:
- minor updates to the default Dockerfile
- introduce .dockerignore to exclude unnecessary files from the docker build context
- made compose example easily executable, no configuration is needed. The Elasticsearch data files are bind mounted to the example folder.

PR #1045, opened 2022-07-19
INFRA-286: Update docker files (7.x)

nagyo:
- minor updates to the default Dockerfile
- introduce .dockerignore to exclude unnecessary files from the docker build context
- made compose example easily executable, no configuration is needed. The Elasticsearch data files are bind mounted to the example folder.


apeteri (approved):
LGTM!

Why do the two `snowowl.yml` files have different content, though?

nagyo:
One is used for the docker image (with embedded ES), and the other is used for the compose example ("remote" ES cluster)

apeteri:
The 8.x counterpart of this PR is https://github.com/b2ihealthcare/snow-owl/pull/1044.

PR #1047, opened 2022-07-25
Elasticsearch 8 compatible high-level client integration

cmark:
This PR integrates the Elasticsearch 8 compatible high-level client in the index layer. Next to the current obsolete TCP and HTTP clients the new client is available via the IndexAdmin interface (`IndexAdmin#es8Client()`). This client is only available when the connected cluster is version 8.0 or higher. The currently supported client version is `8.3.2`.

Related to https://snowowl.atlassian.net/browse/SO-4997


commons/com.b2international.index.es8/src/com/b2international/index/es8/Es8Client.java
```diff
@@ -0,0 +1,95 @@
...
+
+	private final HttpHost host;
+	
+	private final ElasticsearchTransport transport;
+	private final ElasticsearchClient client;
+
+	public Es8Client(String clusterName, String clusterUrl, String username, String password, int connectTimeout, int socketTimeout, SSLContext sslContext, ObjectMapper mapper) {
+		this.host = HttpHost.create(clusterUrl);
+		
+		final RequestConfigCallback requestConfigCallback = requestConfigBuilder -> requestConfigBuilder
```
apeteri, line 64: Is the 100 MB &rarr; 1 GB request limit increase also applicable to this client somehow? What is the default here, and can it be changed via this callback or the one below?

This is the part I'm thinking of:

https://github.com/b2ihealthcare/snow-owl/blob/871a777ade4fdc85f6a209771f05b9d07b4e9c8b/commons/com.b2international.index/src/com/b2international/index/es/client/http/EsHttpClient.java#L59-L72

commons/com.b2international.index/src/com/b2international/index/Searcher.java
```diff
@@ -72,6 +73,18 @@ public interface Searcher {
...
+	/**
+	 * Perform k nearest neighbor search on the select index using the provided search parameters.
+	 * 
+	 * @param <T> - the response type
+ 	 * @param knn - the knn query to run
+	 * @return a {@link Hits} containing all hits returned by the knn search
+	 * @throws IOException
+	 */
+	default <T> Hits<T> knn(Knn<T> knn) throws IOException {
+		throw new UnsupportedOperationException("This feature is only supported is specific searcher implementations");
```
apeteri: ⚠️ Typo alert
```suggestion
		throw new UnsupportedOperationException("This feature is only supported in specific searcher implementations");
```

commons/com.b2international.index/src/com/b2international/index/es/EsDocumentSearcher.java
```diff
@@ -532,5 +541,41 @@ public int resultWindow() {
...
+		
+		Es8QueryBuilder q = new Es8QueryBuilder(mapping, admin.settings(), admin.log());
+		co.elastic.clients.elasticsearch._types.query_dsl.Query esQuery = q.build(knn.getFilter());
+		
+		// TODO consider adding support for double primitive lists
+		FloatIterator it = knn.getQueryVector().iterator();
+		final List<Double> queryVector = new ArrayList<>(knn.getQueryVector().size());
+		while (it.hasNext()) {
+			queryVector.add((double) it.next());
+		}
```
apeteri, line 562: This is a bit strange on the ES Java Client's part, as the internal representation is certainly limited to 32 bits in precision, as mentioned in [Dense vector field type](https://www.elastic.co/guide/en/elasticsearch/reference/current/dense-vector.html) and [Functions for vector fields](https://www.elastic.co/guide/en/elasticsearch/reference/master/query-dsl-script-score-query.html#vector-functions):

> The dense_vector field type stores dense vectors of float values.

> `doc[<field>].vectorValue` – returns a vector’s value as an array of floats

Doubles are most likely used to overcome some JSON serialization limitations. I don't think our Knn search vector should adopt this type and let users think that higher precision can be used than what is actually available.

commons/com.b2international.index/src/com/b2international/index/es/query/Es8QueryBuilder.java
```diff
@@ -0,0 +1,499 @@
...
+		} else if (expression instanceof ScriptQueryExpression){
+			visit((ScriptQueryExpression)expression);
+		} else {
+			throw new IllegalArgumentException("Unexpected expression: " + expression);
+		}
+	}
+
+	private void visit(ScriptQueryExpression expression) {
+		throw new UnsupportedOperationException();
+//		deque.push(QueryBuilders.scriptQuery(expression.toEsScript(mapping)));
```
apeteri: Please add a ticket in Jira for all the commented sections, so we know what gaps in functionality we need to fill in later.

apeteri (approved):

commons/com.b2international.index.es8/src/com/b2international/index/es8/Es8Client.java
cmark: Right, that needs to be applied here as well. I think it would be better to create only a single HTTP client, will take a look whether that is possible or not. The new ES8 client library uses a newer version of the Apache HTTP, hopefully we can migrate to that.

commons/com.b2international.index/src/com/b2international/index/es/EsDocumentSearcher.java
cmark: Yes, that's why I have used FloatList instead of introducing doubles. In our API we will index an array of float values and will support searching using an array of float values, no doubles.

commons/com.b2international.index.es8/src/com/b2international/index/es8/Es8Client.java
cmark: Done

commons/com.b2international.index/src/com/b2international/index/es/query/Es8QueryBuilder.java
cmark: Implemented all remaining expression to query transformations.

apeteri (approved):
Review comments are resolved! 👍

PR #1049, opened 2022-07-28
Suggest API 2.0

cmark:
Introduce new Suggest API format:
* from - an URI with optional ECL query part that specifies where the suggestions should come from
* like - an array of like items (either simple terms or URIs with ECL query part)
* unlike - an array of like items (either simple terms or URIs with ECL query part)

Deprecate the `GET /suggest` endpoint as the new API has nested JSON configuration options, which are not meant to be specified via query parameters.
Deprecate old query parameters:
* `term` - replaced by the `term` suggester and `like` parameter
* `minOccurenceCount` - replaced by specific suggester configuration
* `codeSystemPath` - replaced by `from` resource URI with optional ECL query part
* `query`/`mustNotQuery` - replaced by the `like`/`unlike` parameters

Extract lexical suggestion algorithm into `TermConceptSuggester` implementation.
Use the same generic algorithm for both previous term-based approaches (single term query, and multi-term based on a selected corpus of concept terms).
Allow customization options like stemming, fuzzy matching, ignore stopwords, and token frequency and length settings.

This PR also
* adds support for [MoreLikeThis](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-mlt-query.html) queries and MLT based suggestions via the suggestion API.
* adds support for knn query based filtering in generic and snomed concept APIs
* bumps Jackson to 2.12.7


core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/request/KnnFilterSupport.java
```diff
@@ -0,0 +1,59 @@
...
+	/**
+	 * Filter results by returning semantically similar matches compared to the given query vector.
+	 * 
+	 * @param queryVector - the query vector to match against
+	 * @param numCandidates - can be used to configure the underlying Elasticsearch knn search to improve accuracy
+	 * @return
+	 * @see #filterByKnn(FloatList)
+	 */
+	default T filterByKnn(FloatList queryVector, Integer numCandidates) {
+		return filterByKnn(queryVector == null ? null : new KnnFilter(queryVector, numCandidates));
```
apeteri, line 46: What do we expect of implementations that receive a `null` vector here and pass it on to the method below as a `KnnFilter` argument of `null`? If they work the same way (throw an exception or do nothing), we should handle it here.

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/request/search/MatchTermFilter.java
```diff
@@ -0,0 +1,201 @@
...
+		private Integer maxExpansions;
+		
+		Builder() {
+		}
+		
+		Builder(MatchTermFilter from) {
+			this.term = from.getTerm();
+			this.minShouldMatch = from.getMinShouldMatch();
+			this.ignoreStopwords = from.isIgnoreStopwords();
+			this.caseSensitive = from.isCaseSensitive();
```
apeteri, line 134: Fuzziness settings could be copied to the builder here 🤔

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/request/search/TermFilter.java
```diff
@@ -0,0 +1,160 @@
...
+	
+	public Expression toExpression(String field) {
+		return toExpression(field, "text", "exact", "prefix");
+	}
+	
+	public abstract Expression toExpression(String field, String textFieldSuffix, String exactFieldSuffix, String prefixFieldSuffix);
+	
+	protected final String fieldAlias(String field, String textFieldSuffix) {
+		return String.join(".", field, textFieldSuffix);
+	}
```
apeteri: Second argument name should be more general:

```suggestion
	protected final String fieldAlias(String field, String fieldSuffix) {
		return String.join(".", field, fieldSuffix);
	}
```

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/request/search/TermFilter.java
```diff
@@ -0,0 +1,160 @@
...
+//	
+//	/**
+//	 * @param term
+//	 * 			- the term to apply the expressions on
+//	 * @return {@link TermFilter}
+//	 */
+//	@JsonIgnore
+//	public static final TermFilter parsedTermMatch(final String term) {
+//		return builder().term(term).parsed(true).build();
+//	}
```
apeteri: If we no longer need the commented section of code above, please remove it. 🗑️

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/request/suggest/ConceptSuggestionRequest.java
```diff
@@ -0,0 +1,110 @@
...
+
+	@Override
+	protected Suggestions doExecute(ServiceProvider context) throws IOException {
+		final ConceptSuggester.Registry suggesterRegistry = context.service(ConceptSuggester.Registry.class);
+		final Set<String> availableSuggesters = suggesterRegistry.getSuggesterTypes();
+		
+		// suggester is mandatory, otherwise we can't suggest
+		if (suggester == null) {
+			throw new BadRequestException("'suggester' is required. Available suggesters are: %s", availableSuggesters);
+		}
```
apeteri: Minor nitpick &ndash; we don't need to retrieve these keys if a suggester was specified:

```suggestion
		// suggester is mandatory, otherwise we can't suggest
		if (suggester == null) {
			final Set<String> availableSuggesters = suggesterRegistry.getSuggesterTypes();		
			throw new BadRequestException("'suggester' is required. Available suggesters are: %s", availableSuggesters);
		}
```

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/request/suggest/TermConceptSuggester.java
```diff
@@ -0,0 +1,161 @@
...
+	 * Whether to ignore any stopwords in the selected text corpus or not. Default value is <code>true</code>.
+	 */
+	@JsonProperty
+	private boolean ignoreStopwords = true;
+	
+	/**
+	 * Whether to perform english word stemming on the tokens or not. Default value is false.
+	 */
+	@JsonProperty
+	private boolean stemming = false;
```
apeteri, line 81: The previous default was to remove affixes from the end of word stems, in other words `true`. We might want to keep it for backwards compatibility.

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/request/SnomedConceptSearchRequest.java
```diff
@@ -215,9 +222,12 @@ protected Expression prepareQuery(BranchContext context) {
 			
 			if (termFilter != null) {
 				try {
-					final ComponentCategory category = SnomedIdentifiers.getComponentCategory(termFilter.getTerm());
-					if (category == ComponentCategory.CONCEPT) {
-						conceptScoreMap.put(termFilter.getTerm(), Float.MAX_VALUE);
+					final String singleTerm = termFilter.getSingleTermOrNull();
```
apeteri, line 226: Don't we need scores for >1 terms as well? `getSingleTermOrNull` returns the term if there was exactly one given.

apeteri (approved):

apeteri:
(I forgot to add this to the review: `Collections3` needs a copyright row update.)

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/request/KnnFilterSupport.java
cmark: We can't handle it here, this is an API that accepts many arguments coming from client-side calls or rest requests, where the argument might be null when you don't want to filter by a query vector. This is similar to all other `filterBy` methods.

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/request/search/MatchTermFilter.java
cmark: Right, thanks!

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/request/search/TermFilter.java
cmark: Yup, we can safely remove them now.

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/request/suggest/TermConceptSuggester.java
cmark: Yeah, this is not meant to be fully backward compatible and the former API was in an experimental state, so this is fine. Stemming was only used by default when you have selected a range of concepts via a query. It was impossible to request stemming when you were only passing in a `term` parameter. This now allows the caller to decide whether the basis of the suggestion should be stemmed or not, regardless of where the source terms came from (defined externally or selected via ECL).

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/request/SnomedConceptSearchRequest.java
cmark: Yeah, that is not supported yet here, not sure if we need it, this is backward compatible with the API functionality. I can add a note here for future selves.

cmark: It is already handled by returning null in case of zero or more than 1 terms specified. I'll add a note though that multi-term ID matching is not supported and `filterById` should be used instead.

apeteri (approved):

PR #1061, opened 2022-09-20
fix(sorting): Favor snomed resources over other toolings

AAAlinaaa:


core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/internal/ResourceDocument.java
```diff
@@ -66,6 +66,9 @@
 @Script(
 		name="typeRank", 
 		script="return params.ranks.getOrDefault(doc.resourceType.value, Integer.MAX_VALUE)")
+@Script(
+		name="snomedFirst", 
+		script="return doc.toolingId.value.equals(\"snomed\") ? \"1\" : \"0\"")
```
cmark: Please change the ranking to 0 for snomed and 1 for everything else.

cmark (changes requested):

cmark (approved):

PR #1063, opened 2022-09-26
feat(fhir.core): Add generic "constraint" filter

apeteri:
It is handled the same way as SNOMED CT's "expression" filter defined on the FHIR code system resource:

```json
{
  "code" : "expression",
  "description" : "The result of the filter is the result of executing the given SNOMED CT Expression Constraint",
  "operator" : ["="],
  "value" : "A SNOMED CT ECL expression (see http://snomed.org/ecl)"
}
```

...but is applicable to any code system tooling with ECL support.


cmark (approved):
🥓

PR #1068, opened 2022-10-12
SO-5535: MRCM evaluations

AAAlinaaa:
Related ticket: https://snowowl.atlassian.net/browse/SO-5501


snomed/com.b2international.snowowl.snomed.core.rest.tests/src/com/b2international/snowowl/snomed/core/request/SnomedMrcmTest.java
```diff
@@ -0,0 +1,59 @@
+package com.b2international.snowowl.snomed.core.request;
```
cmark, line 16: ⚠️ Missing license!

snomed/com.b2international.snowowl.snomed.core.rest.tests/src/com/b2international/snowowl/snomed/core/request/SnomedMrcmTest.java
```diff
@@ -0,0 +1,59 @@
...
+
+import org.junit.Test;
+
+import com.b2international.snowowl.snomed.common.SnomedConstants.Concepts;
+import com.b2international.snowowl.snomed.core.domain.refset.SnomedReferenceSetMembers;
+import com.b2international.snowowl.snomed.datastore.request.SnomedRequests;
+import com.b2international.snowowl.test.commons.Services;
+import com.google.common.base.Objects;
+
+public class SnomedMrcmTest {
```
cmark, line 43: ⚠️ Missing since tag!

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/request/SnomedRequests.java
```diff
@@ -386,5 +397,97 @@ public static SnomedRepositoryCommitRequestBuilder prepareCommit() {
 	public static SnomedConceptSearchRequestBuilder prepareGetSynonyms() {
 		return prepareSearchConcept().all().filterByActive(true).filterByEcl("<<"+Concepts.SYNONYM);
 	}
-
+	
+	public static Promise<Collection<String>> getApplicableTypes(final IEventBus bus, final String resourcePath,
```
cmark: Let's push this completely to the server side as one request to form a new Java API.

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/request/SnomedRequests.java
```diff
@@ -386,5 +397,97 @@ public static SnomedRepositoryCommitRequestBuilder prepareCommit() {
...
+				SnomedConceptSearchRequestBuilder requestBuilder = SnomedRequests.prepareSearchConcept()
+						.filterByActive(true)
+						.filterByIds(typeIds)
+						.setFields(SnomedConceptDocument.Fields.ID);
+				
+				if (needsDataAttributes && !needsObjectAttributes) {
+					requestBuilder.filterByAncestor(SnomedConstants.Concepts.CONCEPT_MODEL_DATA_ATTRIBUTE);
+				} else if (!needsDataAttributes && needsObjectAttributes) {
+					requestBuilder.filterByAncestor(SnomedConstants.Concepts.CONCEPT_MODEL_OBJECT_ATTRIBUTE);
+				}
```
cmark: Add another else block with a meaningful comment for the missing case (both boolean parameters are false).

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/request/SnomedRequests.java
```diff
@@ -386,5 +397,97 @@ public static SnomedRepositoryCommitRequestBuilder prepareCommit() {
...
+					.all()
+					.filterByActive(true)
+					.filterByRefSetType(SnomedRefSetType.MRCM_ATTRIBUTE_DOMAIN)
+					.filterByRefSet(inScopeRefSetIds)
+					.filterByProps(Options.from(Map.of(SnomedRf2Headers.FIELD_MRCM_DOMAIN_ID, domainIds)))
+					.build(resourcePath)
+					.execute(bus);
+			}).thenWith(members -> {
+				Set<String> typeIds = members.stream().map(m -> m.getReferencedComponentId()).collect(Collectors.toSet());
+				if ((needsDataAttributes && needsObjectAttributes) || typeIds.isEmpty()) {
```
cmark: What happens if I specify both boolean parameters with the value `false`?

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/request/SnomedRequests.java
```diff
@@ -386,5 +397,97 @@ public static SnomedRepositoryCommitRequestBuilder prepareCommit() {
...
+				if (refSetIds != null) {
+					domainIds.addAll(refSetIds);
+				}
+				
+				return SnomedRequests.prepareSearchMember()
+					.all()
+					.filterByActive(true)
+					.filterByRefSetType(SnomedRefSetType.MRCM_ATTRIBUTE_DOMAIN)
+					.filterByRefSet(inScopeRefSetIds)
+					.filterByProps(Options.from(Map.of(SnomedRf2Headers.FIELD_MRCM_DOMAIN_ID, domainIds)))
```
cmark: Field selection could speed this up a bit.

cmark: Also, what happens if `domainIds` variable ends up empty?

snomed/com.b2international.snowowl.validation.snomed/src/main/resources/scripts/rule_mrcm_constraint.groovy
```diff
@@ -50,14 +68,18 @@ searcher.search(mrcmRangeMemberQuery).each { hit ->
 	String typeId = hit[1];
 	String rangeConstraint = hit[2];
 	
-	if (allowedRanges.containsKey(typeId)) {
-		if (allPrecoordinatedContent.equals(contentType)) {
-			allowedRanges.put(typeId, rangeConstraint);
+	if (rangeConstraint.startsWith("int") || rangeConstraint.startsWith("dec")) {
```
cmark: Magic string values here. Please move them to constants.

snomed/com.b2international.snowowl.validation.snomed/src/main/resources/scripts/rule_mrcm_constraint_type.groovy
```diff
@@ -240,14 +254,90 @@ def searchRelationshipsWithUnregulatedTypeIds =  {
...
 		.limit(50_000)
 		.build();
 	
-	searcher.stream(owlMemberQuery).forEach({ hits -> });
+	searcher.stream(owlMemberQuery).forEach({ hits ->
+		hits.each { hit ->
+			def id = hit[0];
+			def owlExpression = hit[1];
+			if (!owlExpression.contains("TransitiveObjectProperty") &&
+				!owlExpression.contains("ReflexiveObjectProperty") &&
```
cmark: These should be moved to constants.

snomed/com.b2international.snowowl.validation.snomed/src/main/resources/scripts/rule_mrcm_constraint_type.groovy
```diff
@@ -240,14 +254,90 @@ def searchRelationshipsWithUnregulatedTypeIds =  {
...
+				!owlExpression.contains("ObjectPropertyChain")) {
+				//Skip axiom member with no generated axiom relationships
+				issues.add(ComponentIdentifier.of(SnomedReferenceSetMember.TYPE, id))
+			}
+		}
+	});
+}
+
+def searchRelationshipsInUnregulatedDomains =  {
+	String regulatedDomainSpace = Joiner.on(" OR ")
```
cmark: If this is an ECL OR operator, then use `Ecl.or` helper instead.

snomed/com.b2international.snowowl.validation.snomed/src/main/resources/scripts/rule_mrcm_constraint_type.groovy
```diff
@@ -240,14 +254,90 @@ def searchRelationshipsWithUnregulatedTypeIds =  {
...
+		.join(FluentIterable.from(domainMembers.values())
+			.transform({ m -> m.getProperties().get(SnomedRf2Headers.FIELD_MRCM_DOMAIN_CONSTRAINT)})
+			.transform({ c -> "(${c})"}));
```
cmark: This becomes unnecessary if you use that `Ecl.or` helper I think.

snomed/com.b2international.snowowl.validation.snomed/src/main/resources/scripts/rule_mrcm_constraint_type.groovy
```diff
@@ -240,14 +254,90 @@ def searchRelationshipsWithUnregulatedTypeIds =  {
...
+				
+	Set<String> unregulatedDomainSpace = getApplicableConcepts("* MINUS (${regulatedDomainSpace})");
```
cmark: Use `Ecl.exclude` helper.

snomed/com.b2international.snowowl.validation.snomed/src/main/resources/scripts/rule_mrcm_constraint_type.groovy
```diff
@@ -240,14 +254,90 @@ def searchRelationshipsWithUnregulatedTypeIds =  {
...
+	
+	//Find non-IsA relationships in domains where no MRCM rule is defined
+	ExpressionBuilder relationshipQueryBuilder = Expressions.builder()
+			.filter(SnomedRelationshipIndexEntry.Expressions.active())
+			.filter(SnomedRelationshipIndexEntry.Expressions.sourceIds(unregulatedDomainSpace))
```
cmark, line 291: Is regulated/unregulated the same as sanctioned/unsanctioned? If yes, please rename these variables to the latter.

snomed/com.b2international.snowowl.validation.snomed/src/test/java/com/b2international/snowowl/validation/snomed/GenericValidationRuleTest.java
```diff
@@ -1296,22 +1300,26 @@ public void rule_mrcm_constraint_type() throws Exception {
 		//First MRCM rule
 		final SnomedRefSetMemberIndexEntry mrcmDomainMember1 = member(Concepts.CONCEPT_MODEL_ATTRIBUTE, Concepts.REFSET_MRCM_DOMAIN_INTERNATIONAL)
 				.referenceSetType(SnomedRefSetType.MRCM_DOMAIN)
+				.refsetId(Concepts.REFSET_MRCM_DOMAIN_INTERNATIONAL)
 				.domainConstraint(String.format("<<%s", Concepts.CONCEPT_MODEL_ATTRIBUTE))
 				.build();
 		
 		final SnomedRefSetMemberIndexEntry mrcmAttributeDomainMember1 = member(Concepts.FINDING_SITE, Concepts.REFSET_MRCM_ATTRIBUTE_DOMAIN_INTERNATIONAL)
 				.referenceSetType(SnomedRefSetType.MRCM_ATTRIBUTE_DOMAIN)
+				.refsetId(Concepts.REFSET_MRCM_ATTRIBUTE_DOMAIN_INTERNATIONAL)
```
cmark, line 1309: Do we have test cases for failing/passing OWL Axiom MRCM relationships?

cmark (changes requested):

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/request/SnomedRequests.java
AAAlinaaa: Currently, it wouldn't return any rules, but it would be probably better if it returned them all instead.

snomed/com.b2international.snowowl.validation.snomed/src/main/resources/scripts/rule_mrcm_constraint_type.groovy
AAAlinaaa: They might, I was trying to make a distinction between cases that break an explicit rule, vs cases that aren't covered by any rules. To me, an attribute that breaks an explicit rule is an unsanctioned attribute, attributes in domains that have no rules defined for them are unregulated.

snomed/com.b2international.snowowl.validation.snomed/src/test/java/com/b2international/snowowl/validation/snomed/GenericValidationRuleTest.java
AAAlinaaa: Yes we do.

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/request/SnomedRequests.java
AAAlinaaa: It now returns all rules if domains is empty

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/request/SnomedRequests.java
AAAlinaaa: Changed this to use an enum instead, so this wouldn't happen.

snomed/com.b2international.snowowl.snomed.core.rest.tests/src/com/b2international/snowowl/snomed/core/request/SnomedMrcmTest.java
AAAlinaaa: Fixed

snomed/com.b2international.snowowl.snomed.core.rest.tests/src/com/b2international/snowowl/snomed/core/request/SnomedMrcmTest.java
AAAlinaaa: Fixed

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/request/SnomedRequests.java
AAAlinaaa: No longer applicable.

snomed/com.b2international.snowowl.snomed.core.rest.tests/src/com/b2international/snowowl/snomed/core/request/SnomedMrcmTest.java
cmark: Update this to 8.8.0 please.

snomed/com.b2international.snowowl.snomed.core.rest/src/com/b2international/snowowl/snomed/core/rest/SnomedMrcmRestService.java
```diff
@@ -0,0 +1,133 @@
...
+import io.swagger.v3.oas.annotations.Parameter;
+import io.swagger.v3.oas.annotations.media.Schema;
+import io.swagger.v3.oas.annotations.responses.ApiResponse;
+import io.swagger.v3.oas.annotations.responses.ApiResponses;
+import io.swagger.v3.oas.annotations.tags.Tag;
+
+/**
+ * @since 8.7.0
+ */
+@Tag(description="Rules", name = "rules")
```
cmark: Let's use `MRCM` or `MRCM Rules` tag instead.

snomed/com.b2international.snowowl.snomed.core.rest/src/com/b2international/snowowl/snomed/core/rest/SnomedMrcmRestService.java
```diff
@@ -0,0 +1,133 @@
...
+	
+	@Operation(
+		summary="Retrieve MRCM attribute domain type refset members filtered by domain and attribute type", 
+		description="Retrieve MRCM relationship type rules ")
+	@ApiResponses({
+		@ApiResponse(responseCode = "200", description = "OK"),
+		@ApiResponse(responseCode = "400", description = "Bad Request"),
+		@ApiResponse(responseCode = "404", description = "Not found"),
+	})
+	@GetMapping(value = "/{path:**}/types", produces = { AbstractRestService.JSON_MEDIA_TYPE })
```
cmark: Not sure about the endpoint's name, I'd like to see the word mrcm in it. `mrcm-rules` perhaps, but I'm not a fan of dashes in endpoint names. Another option would be `{path:**}/mrcm/types`. @nagyo @apeteri thoughts?

snomed/com.b2international.snowowl.snomed.core.rest/src/com/b2international/snowowl/snomed/core/rest/SnomedMrcmRestService.java
```diff
@@ -0,0 +1,133 @@
...
+	
+	@Operation(
+			summary="Retrieve MRCM attribute domain range refset members filtered by domain and attribute type", 
+			description="Retrieve MRCM relationship range rules ")
+		@ApiResponses({
+			@ApiResponse(responseCode = "200", description = "OK"),
+			@ApiResponse(responseCode = "400", description = "Bad Request"),
+			@ApiResponse(responseCode = "404", description = "Not found"),
+		})
+		@GetMapping(value = "/{path:**}/ranges", produces = { AbstractRestService.JSON_MEDIA_TYPE })
```
cmark: Same endpoint name change ideas here. See above.

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/request/MrcmRangeRequest.java
```diff
@@ -0,0 +1,128 @@
...
+import com.b2international.snowowl.core.domain.BranchContext;
+import com.b2international.snowowl.core.events.Request;
+import com.b2international.snowowl.snomed.core.domain.refset.SnomedReferenceSetMember;
+import com.b2international.snowowl.snomed.core.domain.refset.SnomedReferenceSetMembers;
+import com.b2international.snowowl.snomed.datastore.request.MrcmTypeRequest.ATTRIBUTE_TYPE;
+
+/**
+ * @since 8.7.0
+ */
+public class MrcmRangeRequest implements Request<BranchContext, SnomedReferenceSetMembers> {
```
cmark: Please prefix the classes with `Snomed`.

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/request/MrcmTypeRequest.java
```diff
@@ -0,0 +1,160 @@
...
+import com.b2international.snowowl.snomed.datastore.index.entry.SnomedConceptDocument;
+
+/**
+ * @since 8.7.0
+ */
+public class MrcmTypeRequest implements Request<BranchContext, SnomedReferenceSetMembers> {
+	
+	private static final long serialVersionUID = 1L;
+
+	public enum ATTRIBUTE_TYPE {
```
cmark: Invalid naming pattern of enum type. Please Rename this to `MrcmAttributeType`.

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/request/MrcmTypeRequest.java
```diff
@@ -0,0 +1,160 @@
...
+import com.b2international.snowowl.snomed.common.SnomedRf2Headers;
+import com.b2international.snowowl.snomed.core.domain.SnomedConcept;
+import com.b2international.snowowl.snomed.core.domain.refset.SnomedRefSetType;
+import com.b2international.snowowl.snomed.core.domain.refset.SnomedReferenceSetMembers;
+import com.b2international.snowowl.snomed.datastore.index.entry.SnomedConceptDocument;
+
+/**
+ * @since 8.7.0
+ */
+public class MrcmTypeRequest implements Request<BranchContext, SnomedReferenceSetMembers> {
```
cmark: Update since tags on all newly introduced classes to 8.8.0 and also ensure that request implementations have package-private visibility.

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/request/MrcmTypeRequest.java
```diff
@@ -0,0 +1,160 @@
...
+	public SnomedReferenceSetMembers execute(BranchContext context) {
+		Set<String> inScopeRefSetIds = SnomedRequests.prepareSearchMember()
+				.all()
+				.filterByActive(true)
+				.filterByRefSetType(SnomedRefSetType.MRCM_MODULE_SCOPE)
+				.filterByReferencedComponent(moduleIds)
+				.setFields(ID, MRCM_RULE_REFSET_ID)
+				.build()
+				.execute(context)
+				.stream()
```
cmark: Please don't use all() anymore unless you have a really good reason to do so. Also, using all + stream together is a no go. Please set the limit correctly.

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/request/MrcmTypeRequest.java
```diff
@@ -0,0 +1,160 @@
...
+				.collect(Collectors.toSet());
+		
+		if (inScopeRefSetIds.isEmpty()) {
+			return new SnomedReferenceSetMembers(0, 0);
+		}
+		
+		String eclConstraint = "*";
+		
+		switch (attributeType) {
+		case DATA: eclConstraint = String.format("<%s", CONCEPT_MODEL_DATA_ATTRIBUTE);
```
cmark: Add line breaks after switch cases, so the switch statement is easier to read.

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/request/MrcmTypeRequest.java
```diff
@@ -0,0 +1,160 @@
...
+			break;
+		case OBJECT: eclConstraint = String.format("<%s", CONCEPT_MODEL_OBJECT_ATTRIBUTE);
+			break;
+		case ALL: eclConstraint = String.format("<%s AND <%s", CONCEPT_MODEL_OBJECT_ATTRIBUTE, CONCEPT_MODEL_DATA_ATTRIBUTE );
+			break;		
+		};
+		
+		final Set<String> typeIds = SnomedRequests.prepareSearchConcept()
+			.all()
```
cmark: Same paging issue here.

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/request/MrcmTypeRequest.java
```diff
@@ -0,0 +1,160 @@
...
+		if (ruleParentIds != null) {
+			domainIds.addAll(ruleParentIds);
+		}
+		
+		if (refSetIds != null) {
+			domainIds.addAll(refSetIds);
+		}
+		
+		SnomedRefSetMemberSearchRequestBuilder requestBuilder = SnomedRequests.prepareSearchMember()
+			.all()
```
cmark: Are we going to return all matching members all the time? What happens if we have hundreds of results? I would like to see a limit parameter introduced in the API to limit the number of types returned here and also a searchAfter parameter to control the paging.

cmark (changes requested):

snomed/com.b2international.snowowl.snomed.core.rest/src/com/b2international/snowowl/snomed/core/rest/SnomedMrcmRestService.java
apeteri: "Attributes" and "attribute ranges" should be incorporated into each path in some form, this is how the documentation refers to relationship types in this context. No strong feelings about including "mrcm".

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/request/MrcmRangeRequest.java
```diff
@@ -0,0 +1,128 @@
...
+			.filterByReferencedComponent(typeIds)
+			.filterByRefSet(inScopeRefSetIds)
+			.setFields(ID, REFERENCED_COMPONENT_ID, MRCM_RANGE_CONSTRAINT, MRCM_CONTENT_TYPE_ID)
+			.build()
+			.execute(context)
+			.forEach(m -> {
+				String contentType = (String) m.getProperties().get(FIELD_MRCM_CONTENT_TYPE_ID);
+				if (!rangeConstraintMembers.containsKey(m.getReferencedComponentId()) || ALL_PRECOORDINATED_CONTENT.equals(contentType)) {
+					rangeConstraintMembers.put(m.getReferencedComponentId(), m);
+				}
```
apeteri: An explicit sort order should be defined for the search request if we are relying on the `Map` entry overwriting behavior this `if` statement enables.

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/request/MrcmTypeRequest.java
```diff
@@ -0,0 +1,160 @@
...
+				.execute(context)
+				.stream()
+				.map(m -> (String) m.getProperties().get(SnomedRf2Headers.FIELD_MRCM_RULE_REFSET_ID))
+				.collect(Collectors.toSet());
+		
+		if (inScopeRefSetIds.isEmpty()) {
+			return new SnomedReferenceSetMembers(0, 0);
+		}
+		
+		String eclConstraint = "*";
```
apeteri: Switching on an enum-typed expression means there is really no way of _not_ entering one of the cases listed below. I'd set this variable to `final` and add a default case that throws an `IllegalStateException` or initializes the constraint to this catch-all expression if it still makes sense to do so.

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/request/MrcmTypeRequest.java
```diff
@@ -0,0 +1,160 @@
...
+		
+		switch (attributeType) {
+		case DATA: eclConstraint = String.format("<%s", CONCEPT_MODEL_DATA_ATTRIBUTE);
+			break;
+		case OBJECT: eclConstraint = String.format("<%s", CONCEPT_MODEL_OBJECT_ATTRIBUTE);
+			break;
+		case ALL: eclConstraint = String.format("<%s AND <%s", CONCEPT_MODEL_OBJECT_ATTRIBUTE, CONCEPT_MODEL_DATA_ATTRIBUTE );
```
apeteri: The AND operator forms an intersection, and the two set of descendants don't have common elements, so the query will return 0 concepts.

```suggestion
		case ALL: eclConstraint = String.format("<%s OR <%s", CONCEPT_MODEL_OBJECT_ATTRIBUTE, CONCEPT_MODEL_DATA_ATTRIBUTE );
```

snomed/com.b2international.snowowl.validation.snomed/src/main/resources/scripts/rule_mrcm_constraint.groovy
```diff
@@ -50,14 +70,18 @@ searcher.search(mrcmRangeMemberQuery).each { hit ->
...
-			allowedRanges.put(typeId, rangeConstraint);
+	if (rangeConstraint.startsWith(integerTypeRangePrefix) || rangeConstraint.startsWith(decimalTypeRangePrefix)) {
+		//Do nothing, skip concrete value type range validation for now
+	} else {
+		if (allowedRanges.containsKey(typeId)) {
+			if (allPrecoordinatedContent.equals(contentType)) {
+				allowedRanges.put(typeId, rangeConstraint);
+			} else {
+				//Do nothing, the already mapped member should be chosen over this
+			}
```
apeteri: Same concern as above wrt. "all pre-coordinated content" mode and overwriting the Map entry.

snomed/com.b2international.snowowl.validation.snomed/src/main/resources/scripts/rule_mrcm_constraint_type.groovy
```diff
@@ -240,14 +259,82 @@ def searchRelationshipsWithUnregulatedTypeIds =  {
...
 		.where(owlMemberExpressionBuilder.build())
 		.limit(50_000)
 		.build();
 	
-	searcher.stream(owlMemberQuery).forEach({ hits -> });
+	searcher.stream(owlMemberQuery).forEach({ hits ->
+		hits.each { hit ->
+			def id = hit[0];
+			def owlExpression = hit[1];
+			def hasAxiomRelationships = !unsupportedAxiomMarkers.stream().filter({ marker -> owlExpression.contains(marker)}).findAny();
```
apeteri, line 273: Hmm. It might be easier to check whether the expression starts with `SubClassOf` or `EquivalentClasses` 😃

snomed/com.b2international.snowowl.validation.snomed/src/main/resources/scripts/rule_mrcm_constraint_type.groovy
apeteri: We didn't make that distinction earlier, but I'm not against it if it describes things better. 

Previously any type not covered by an MRCM rule was considered unsanctioned, regardless of whether it was added to a concept living in a regulated or an unregulated region, in exchange for a warning. Breaking the rules is indeed an error (or "sacrilege" if we stick with the terminology adopted for this subject). I don't know if we even offer "clearly illegal" attributes for selection on the UI, but they can be part of an import.

apeteri (changes requested):

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/request/MrcmTypeRequest.java
apeteri: This request queries a small set of module scope members, which should be kept in practice at a low member count (the current tally is 3, and the expected count is 3 × number of modules declaring MRCM rules). `stream()` is called on the returned collection resource and processes individual members, not batches.

snomed/com.b2international.snowowl.validation.snomed/src/main/resources/scripts/rule_mrcm_constraint_type.groovy
apeteri: My comment doesn't hold true for the object and data property hierarchies, see section "Representation of |is a| Relationships" in the SNOMED CT OWL Guide: https://confluence.ihtsdotools.org/display/DOCOWL/2.4.+Content+for+the+OWL+Axiom+Refset

Only property roots receive a `SubClassOf` axiom that connects them to the `410662002|Concept model attribute (attribute)|`; the rest of each hierarchy is modeled using `SubObjectPropertyOf` and `SubDataPropertyOf` axioms, respectively.

snomed/com.b2international.snowowl.validation.snomed/src/main/resources/scripts/rule_mrcm_constraint_type.groovy
AAAlinaaa: We do offer both sanctioned and unsanctioned properties for selection.

snomed/com.b2international.snowowl.validation.snomed/src/main/resources/scripts/rule_mrcm_constraint.groovy
```diff
@@ -29,12 +29,32 @@ RevisionSearcher searcher = ctx.service(RevisionSearcher.class);
 Set<ComponentIdentifier> issues = Sets.newHashSet();
 
 final String allPrecoordinatedContent = "723594008";
```
apeteri: Please change this SCTID so that it points to `SnomedConstants.Concepts`.

apeteri: ...by which I mean "remove the variable and use the constant directly" 😄

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/request/MrcmRangeRequest.java
apeteri: This is also a non-issue as `ALL_PRECOORDINATED_CONTENT` is allowed to displace other entries. I mis-read the constant as `ALL_CONTENT` when reviewing the code earlier.

snomed/com.b2international.snowowl.validation.snomed/src/main/resources/scripts/rule_mrcm_constraint.groovy
AAAlinaaa: Done

snomed/com.b2international.snowowl.snomed.core.rest/src/com/b2international/snowowl/snomed/core/rest/SnomedMrcmRestService.java
```diff
@@ -48,7 +48,7 @@ public class SnomedMrcmRestService extends AbstractRestService {
 		@ApiResponse(responseCode = "400", description = "Bad Request"),
 		@ApiResponse(responseCode = "404", description = "Not found"),
 	})
-	@GetMapping(value = "/{path:**}/types", produces = { AbstractRestService.JSON_MEDIA_TYPE })
+	@GetMapping(value = "/{path:**}/mrcm/attribute/domains", produces = { AbstractRestService.JSON_MEDIA_TYPE })
```
apeteri: Either `mrcm` or `attribute` should be enough as the path segment.

snomed/com.b2international.snowowl.snomed.core.rest/src/com/b2international/snowowl/snomed/core/rest/SnomedMrcmRestService.java
```diff
@@ -93,7 +102,7 @@ public class SnomedMrcmRestService extends AbstractRestService {
 			@ApiResponse(responseCode = "400", description = "Bad Request"),
 			@ApiResponse(responseCode = "404", description = "Not found"),
 		})
-		@GetMapping(value = "/{path:**}/ranges", produces = { AbstractRestService.JSON_MEDIA_TYPE })
+		@GetMapping(value = "/{path:**}/mrcm/attribute/ranges", produces = { AbstractRestService.JSON_MEDIA_TYPE })
```
apeteri: Same here.

apeteri (approved):
One small observation below. I think the rest of my requests were addressed, so LGTM otherwise!

snomed/com.b2international.snowowl.snomed.core.rest/src/com/b2international/snowowl/snomed/core/rest/SnomedMrcmRestService.java
AAAlinaaa: I guess I misunderstood, I thought you said it has to be attribute-domain, at the same time Mark asked to include the word mrcm

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/request/SnomedMrcmTypeRequest.java
```diff
@@ -81,47 +65,48 @@ public void setModuleIds(List<String> moduleIds) {
...
+		case OBJECT: 
+			eclConstraint = String.format("<%s", CONCEPT_MODEL_OBJECT_ATTRIBUTE);
 			break;
-		case ALL: eclConstraint = String.format("<%s AND <%s", CONCEPT_MODEL_OBJECT_ATTRIBUTE, CONCEPT_MODEL_DATA_ATTRIBUTE );
-			break;		
+		case ALL: 
+			eclConstraint = String.format("<%s OR <%s", CONCEPT_MODEL_OBJECT_ATTRIBUTE, CONCEPT_MODEL_DATA_ATTRIBUTE );
+			break;
+		default: 
+			eclConstraint = "*";
```
cmark: There is an Ecl.ANY constant, please use that.

cmark (changes requested):

cmark (approved):
🥓

PR #1071, opened 2022-10-26
fix(EsIndexAdmin): Use annotation index field on map type fields too

apeteri:
The previous PR #1070 had to be rolled back, created this one for another round of review.

See also PR #1054 with overlapping changes.


cmark (approved):
This looks okay, but a test case would be nice to have.

PR #1073, opened 2022-10-27
Feature: accessing versioned resource via GET endpoints using URIs

cmark:
This PR adds the ability to access a versioned form of a resource via the following endpoints:
* `GET /bundles/:id`
* `GET /codesystems/:id`

The id argument now accepts a complete URI, including both version paths and timestamp parts.
Examples:
* `GET /codesystems/SNOMEDCT/2022-07-31`
* `GET /codesystems/SNOMEDCT@1659225600`

### DIFF
--- a/core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/bundle/Bundle.java
+++ b/core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/bundle/Bundle.java
@@ -95,6 +95,10 @@ public final class Bundle extends Resource {
 		return bundleUri.startsWith(RESOURCE_TYPE) ? new ResourceURI(bundleUri) : ResourceURI.of(RESOURCE_TYPE, bundleUri);
 	}
 	
+	public static ResourceURI uri(String bundleUri, String versionId) {
+		return uri(bundleUri).withPath(versionId);
+	}
+	
 	public static Bundle from(ResourceDocument doc) {
 		final Bundle bundle = new Bundle();
 		bundle.setId(doc.getId());
--- a/core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/bundle/Bundle.java
+++ b/core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/bundle/Bundle.java
@@ -16,6 +16,7 @@
 package com.b2international.snowowl.core.bundle;
 
 import com.b2international.snowowl.core.Resource;
+import com.b2international.snowowl.core.ResourceURI;
 import com.b2international.snowowl.core.Resources;
 import com.b2international.snowowl.core.internal.ResourceDocument;
 import com.b2international.snowowl.core.internal.ResourceDocument.Builder;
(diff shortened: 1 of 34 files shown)


core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/bundle/Bundle.java
```diff
@@ -89,6 +90,10 @@ public Builder toDocumentBuilder() {
 				.bundleId(getBundleId())
 				.settings(getSettings());
 	}
+
+	public static ResourceURI uri(String bundleUri) {
+		return bundleUri.startsWith(RESOURCE_TYPE) ? new ResourceURI(bundleUri) : ResourceURI.of(RESOURCE_TYPE, bundleUri);
```
apeteri: We might want to add a trailing `/` to the constant in these `startsWith` checks. As an example, `"bundlesort"` would be a resource-type-less URI, but the code treats it as one that has a type because it starts with `"bundles"`.

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/internal/ResourceRepository.java
```diff
@@ -140,10 +141,10 @@ public void run(StagingArea staging) {
...
 				
 				ResourceRequests.prepareSearchVersion()
-					.all()
+					.setLimit(10_000)
 					.filterByResources(resources)
-					.build()
-					.execute(context)
+					.stream(context)
+					.flatMap(Versions::stream)
 					.forEach(version -> {
```
apeteri: As this became a `Stream`, `forEachOrdered` should be used so we don't get caught surprised by a future version of Java that enables wider parallel processing.

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/request/resource/BaseGetResourceRequest.java
```diff
@@ -0,0 +1,66 @@
...
+	
+	@Override
+	public final Long getReadTimestamp(ServiceProvider context) {
+		if (!Strings.isNullOrEmpty(resourceUri.getTimestampPart())) {
+			return Long.parseLong(resourceUri.getTimestampPart().substring(1));
+		} else if (!resourceUri.isHead() && !resourceUri.isNext()) {
+			return ResourceRequests.prepareSearchVersion()
+					.one()
+					.filterByResource(resourceUri.withoutPath())
+					.filterByVersionId(resourceUri.getPath())
```
apeteri: We might be able to come up with a filter in the future that matches any descendant of a version branch path. Bookmarking! 🔖

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/request/resource/BaseGetResourceRequestBuilder.java
```diff
@@ -0,0 +1,36 @@
...
+import com.b2international.snowowl.core.request.SearchResourceRequestBuilder;
+
+/**
+ * @since 8.7
+ */
+public abstract class BaseGetResourceRequestBuilder<
+	B extends GetResourceRequestBuilder<B, SB, RepositoryContext, SR, R>, 
+	SB extends SearchResourceRequestBuilder<SB, RepositoryContext, SR>, 
+	SR, 
+	R> extends GetResourceRequestBuilder<B, SB, RepositoryContext, SR, R> implements ResourceRepositoryRequestBuilder<R> {
```
apeteri: 😖 Please move this last `extends` to its own line, so there is a separation between the type parameters and the base class of this builder.

apeteri (approved):
👍

nagyo (approved):
semi-RSLGTM!

PR #1074, opened 2022-11-07
feat(api): support effective time range filtering in SCT Concept Search

cmark:


apeteri (approved):
LGTM! 

For brownie points 🍪 a test case for the pre-existing single effective time query parameter can also be added.

nagyo (approved):
LGTM!

PR #1077, opened 2022-11-11
SO-5544: Log errors encountered during remote job execution

AAAlinaaa:


cmark (approved):
🥓 please also apply this to the 8.x branch.

PR #1083, opened 2022-11-11
SO-5544: Log errors encountered during remote job execution (8.x) 

nagyo:


cmark (approved):
🥓

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/jobs/RemoteJob.java
```diff
@@ -119,6 +123,7 @@ protected final IStatus run(IProgressMonitor monitor) {
 					.build();
 			}
 			
+			LOG.error("Error encountered while executing remote job", e);
```
apeteri: Some identifying information about the job should be added to the message. `id`, `description`, class name of `request` are all good candidates.

apeteri (changes requested):

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/jobs/RemoteJob.java
nagyo: +1 for that

apeteri (approved):
👍

PR #1086, opened 2022-11-17
fix(fhir.core): Use "createdAt" for point-in-time resource document...

apeteri:
...queries instead of "created"'s timestamp portion for backwards compatibility in datasets where no FHIR resource metadata is stored on version documents.


cmark (approved):
🥓

nagyo (approved):
LGTM!

PR #1089, opened 2022-11-18
SNRAY-1117: Make "version" field mandatory to load as well

apeteri:
The value is used in `fillResourceOnlyProperties` to distinguish between the two possible sources of a `ResourceFragment` instance (they are created either from a resource or a version document). If it is not part of the set of fields to be loaded, we can no longer determine the source correctly.


cmark (approved):
🥓

nagyo (approved):
LGTM!

PR #1095, opened 2022-12-05
SO-3802: improve JWT configuration

cmark:
This PR introduces a few changes to the existing access token signing and verification module.

Authentication changes:
* The system still accepts externally signed access token but instead of configuring the externally available JWKS as a global config, administrators should use the new `jwks` identity provider module. It can have its own issuer, email and permissions claim property configuration different from the global one.
* Snow Owl is now able to generate access tokens even when using external JWKS identity providers, no longer limited to just verification

Configuration changes:
* Deprecated `identity.jwksUrl` in favor of the new `jwks` identity provider


core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/identity/AuthorizationHeaderVerifier.java
```diff
@@ -107,51 +78,18 @@ public User auth(String authorizationHeaderValue) {
 	 */
 	public User authJWT(final String token) {
 		try {
-			final DecodedJWT jwt = verifier.verify(token);
-			return toUser(jwt);
-		} catch (JWTVerificationException e) {
-			throw new UnauthorizedException("Incorrect authorization token");
+			return jwtSupport.authJWT(token);
+		} catch (Exception e) {
+			// try to authorize JWT using the global provider first, then fall back to the dedicated identity provider specific JWT authentication
```
apeteri, line 83: We could log these exceptions with a low enough log level 🤔

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/identity/IdentityConfiguration.java
```diff
@@ -62,73 +51,20 @@ public void setProviderConfigurations(List<IdentityProviderConfig> providerConfi
...
-	
-	public String getJws() {
-		return jws;
-	}
-	
+	/**
+	 * @return
+	 * @deprecated - use dedicated jwks identity provider to configure external JWKS based authentication
+	 */
 	public String getJwksUrl() {
```
apeteri, line 58: This value doesn't seem to be retrieved any more &ndash; I thought a minimal `JwksIdentityProvider` will be built out of top-level values if present, for backwards compatibility. 

Is it kept here only so that existing configuration files don't break?

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/identity/JWTSupport.java
```diff
@@ -0,0 +1,281 @@
...
+	 * Converts the given JWT access token to a {@link User} representation using the configured email and permission claims. This method does not
+	 * verify the given access token, it only decodes it and uses the publicly available claims to construct the {@link User} object. To verify a
+	 * token and create a user object use the {@link #authJWT(String)} method.
+	 * 
+	 * @param token
+	 * @return a {@link User} instance created from the given token
+	 */
+	public User toUser(String token) {
+		return toUser(JWT.decode(token), config);
+	}
```
apeteri: Can you double check that all current callers use these two methods as intended (ie. no one takes JWTs at face value unless they were verified at an earlier point in time)?

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/identity/JWTSupport.java
```diff
@@ -0,0 +1,281 @@
...
+						.replace(PKCS8_FOOTER, "");
+				PKCS8EncodedKeySpec keySpec = new PKCS8EncodedKeySpec(Base64.getDecoder().decode(extractedKey));
+				return (RSAPrivateKey) KeyFactory.getInstance("RSA").generatePrivate(keySpec);
+			} catch (InvalidKeySpecException | NoSuchAlgorithmException e) {
+				throw new SnowOwl.InitializationException("Invalid signingKey. Only Base64 encoded PKCS#1, PKCS#8 or JWK keys are supported. Error: " + e.getMessage());
+			}			
+		} else {
+			// TODO support PKCS#1
+			// TODO support JWK strings
+			throw new SnowOwl.InitializationException(String.format(""));
```
apeteri: An exception message needs to be added here.

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/identity/JWTSupport.java
```diff
@@ -0,0 +1,281 @@
...
+		}
+
+		final String permissionsClaimProperty = config.getPermissionsClaimProperty();
+		Claim permissionsClaim = jwt.getClaim(permissionsClaimProperty);
+		if (permissionsClaim == null || permissionsClaim.isNull()) {
+			throw new BadRequestException("'%s' JWT access token field is required for permissions access, but it was missing.",
+					permissionsClaimProperty);
+		}
+
+		final List<Permission> permissions = jwt.getClaim(permissionsClaimProperty).asList(String.class).stream().map(Permission::valueOf)
```
apeteri: The claim already seems to be extracted to a local variable `permissionsClaim`. I'd recommend using it again in this expression like `emailClaim` is re-used below.

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/identity/file/FileIdentityProvider.java
```diff
@@ -49,11 +50,19 @@ final class FileIdentityProvider implements IdentityProvider, IdentityWriter {
...
-	private final Path usersFile;
-	private final Map<String, String> verifiedTokens;
+	private final FileIdentityProviderConfig configuration;
 	
-	public FileIdentityProvider(Path usersFile) throws IOException {
+	private Path usersFile;
+	private Map<String, FileUser> users;
+	private Map<String, String> verifiedTokens;
+	
+	public FileIdentityProvider(FileIdentityProviderConfig configuration) throws IOException {
```
apeteri: This constructor should no longer throw its declared exception 😄

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/identity/file/FileIdentityProvider.java
```diff
@@ -71,7 +80,9 @@ public User auth(String username, String token) {
 			final FileUser user = getFileUser(username);
 			boolean success = user != null && BCrypt.checkpw(token, user.getHashedPassword());
 			if (success) {
-				verifiedTokens.put(username, token);
+				synchronized (verifiedTokens) {
+					verifiedTokens.put(username, token);
```
apeteri, line 84: The second parameter will be the user's password according to the superinterface. Is this map meant to cache passwords or tokens?

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/identity/IdentityProvider.java
```diff
@@ -22,105 +22,106 @@
...
+
 		@Override
 		public User auth(String username, String password) {
 			return new User(username, List.of(Permission.ADMIN));
 		}
-		
+
+		@Override
+		public User authJWT(String token) {
+			final DecodedJWT decodedToken = JWT.decode(token);
```
apeteri, line 53: Please add a comment here that tells readers that this identity provider has no means of verifying JWT contents.

apeteri (changes requested):
I have a few small requests, but otherwise changes look good.

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/identity/AuthorizationHeaderVerifier.java
cmark: Trace? 😅

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/identity/IdentityConfiguration.java
cmark: You are right, I forgot to add this as an identity provider when set and raise a deprecation log. Also, raise an error when there is a separate `jwks` provider set along with this configuration.

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/identity/JWTSupport.java
cmark: The second method is not used at all, we can safely remove it.

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/identity/file/FileIdentityProvider.java
cmark: Cache to speed up tests, but it would be great to disable this in PROD. Thoughts?

apeteri: Fully agree 😄

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/identity/IdentityConfiguration.java
cmark: Done. @apeteri please review again, thanks!

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/identity/file/FileIdentityProvider.java
cmark: Let's raise an issue about this, the FILE provider is not meant to be used in most production environments anyway, so I'm not seeing this as a major issue right now.

apeteri (approved):
LGTM!

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/DeprecationLogger.java
```diff
@@ -0,0 +1,38 @@
...
+ */
+package com.b2international.snowowl.core;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * @since 8.8.0
+ */
+public final class DeprecationLogger {
```
nagyo, line 24: ❤️‍🔥

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/identity/file/FileIdentityProvider.java
nagyo: 👍🏻

nagyo (approved):
LGTM, with one minor request. I would love to see a good configuration example (best if multiple is present) e.g. in here: https://github.com/b2ihealthcare/snow-owl/blob/8.x/docker/config/snowowl.yml

PR #1097, opened 2023-01-06
SO-5580: Fix effective time restoration following status changes (7.x)

apeteri:


nagyo (approved):
LGTM! This boxed Boolean bug rings a bell, but I cannot recall what it was exactly...

cmark (approved):
LGTM!

PR #1100, opened 2023-01-10
Use ubuntu as base image for docker containers (8.x)

nagyo:
- simplified builder stage
- more secure file ownership settings
- use tini to init

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -15,16 +15,6 @@ RUN chmod 0660 configuration/snowowl.yml
 
 FROM ${BASE_IMAGE}
 
-ARG BUILD_TIMESTAMP
-ARG TAG
-ARG GIT_REVISION
-ARG GIT_BRANCH
-ARG PRODUCT_NAME="Snow Owl OSS"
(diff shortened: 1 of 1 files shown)


docker/Dockerfile
```diff
@@ -23,48 +21,58 @@ ARG GIT_REVISION
 ARG GIT_BRANCH
 ARG PRODUCT_NAME="Snow Owl OSS"
 ARG REPOSITORY_URL="https://github.com/b2ihealthcare/snow-owl"
+ARG CI_PASSED=""
+ARG WIN_CLIENT_URL=""
+ARG MAC_CLIENT_URL=""
```
cmark: Do we really need the client urls here?

cmark (changes requested):

docker/Dockerfile
nagyo: Nope, good catch :thumbsup:

cmark (approved):
🥓

apeteri (approved):
👍

PR #1107, opened 2023-02-07
SO-5616: Test new scenarios covered by the acceptability updater

AAAlinaaa:


apeteri (approved):
LGTM, all listed cases are covered. The test class also needs a date bump.

nagyo (approved):
RSLGTM!

snomed/com.b2international.snowowl.snomed.core.rest.tests/src/com/b2international/snowowl/snomed/core/rest/components/SnomedDescriptionApiTest.java
```diff
@@ -449,6 +453,132 @@ public void updateAcceptabilityAndInactivate() {
 			.statusCode(200)
 			.body("pt.id", equalTo(ptDescriptionId));
 	}
+	
+	@Test
+	public void updateAcceptabilityWithMultipleAcceptabilityMembersForSameRefset1() {
```
cmark: Please use a valid test method name instead of a number suffix. If this test case verifies a certain scenario then pls add it to the name and also add a comment above describing the scenario. Same applies to the other test cases.

cmark (changes requested):

snomed/com.b2international.snowowl.snomed.core.rest.tests/src/com/b2international/snowowl/snomed/core/rest/components/SnomedDescriptionApiTest.java
AAAlinaaa: The scenario comments are inside the body of the test case

cmark (approved):
🥓

PR #1108, opened 2023-02-07
SNRAY-1091: Support reindex operation for Elasticsearch indices

nagyo:
### Introduces:
- the ability to `refresh` specific indices if requested
- the ability to `reindex` specific indices either locally or from a remote source
- ReindexResult object to store results of a given reindex operation
- ReindexQuery object to represent queries executed against a reindex source index
- custom de/serializer for org.elasticsearch.index.query.QueryBuilder
- de/serialization tests for repository info, server info and reindex queries
- all required dependencies to build custom docker containers with testcontainers
### Fixes:
- potential NPE when creating ApiExceptions with additional information
- deserialization of server info, repository info, ES cluster status and ES index status
- async execution of the following REST endpoints:
  - `POST /upgrade/{id}/complete`
  - `POST /upgrade/sync`
### Improves:
- handling of ObjectMapper instances with custom configuration


commons/com.b2international.commons/src/com/b2international/commons/exceptions/ApiException.java
```diff
@@ -87,7 +88,9 @@ public final Map<String, Object> getAdditionalInfo() {
 	 * @return this instance for method chaining
 	 */
 	public final ApiException withDeveloperMessage(String message, Object...args) {
-		this.developerMessage = String.format(message, args);
+		if (!Strings.isNullOrEmpty(message)) {
```
cmark: Is this really something we have to check here? Supplying a `null` message is clearly a programmer error and not something I would allow in the system. Could you please point me to the place where this caused an error?

commons/com.b2international.index.tests/src/com/b2international/index/ReindexResultSerializationTest.java
```diff
@@ -0,0 +1,86 @@
+/*******************************************************************************
+ * Copyright (c) 2023 B2i Healthcare. All rights reserved.
+ *******************************************************************************/
```
cmark: Ensure that all copyright headers in this repository use the OSS header.

commons/com.b2international.index/src/com/b2international/index/admin/IndexAdmin.java
```diff
@@ -133,7 +138,26 @@ public interface IndexAdmin {
...
+	/**
+	 * Issue a refresh on the specified indices
+	 * @return {@link RefreshResponse}
+	 * @param indices
+	 */
+	RefreshResponse refresh(String...indices) throws IOException;
 
+	/**
+	 * Issue a remote reindex of sourceIndex to destinationIndex using the parameters specified in remoteInfo. A selective
+	 * query can be pass inside the RemoteInfo object in a serialized form.
```
cmark:
```suggestion
	 * query can be passed inside the RemoteInfo object in a serialized form.
```

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/bundle/Bundle.java
```diff
@@ -55,7 +55,7 @@ public String getResourceType() {
 		return RESOURCE_TYPE;
 	}
 
-	public BundleCreateRequestBuilder toCreateRequest() {
+	public BundleCreateRequestBuilder toCreateRequestBuilder() {
```
cmark: I think we have deliberately named these without the Builder suffix even if the returned object is a builder. This creates a bit more fluent and readable API than including the Builder part. I'm open to suggestions.

cmark (approved):
LGTM!

commons/com.b2international.index/src/com/b2international/index/es/client/EsClusterStatus.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2019 B2i Healthcare Pte Ltd, http://b2i.sg
+ * Copyright 2023 B2i Healthcare Pte Ltd, http://b2i.sg
```
apeteri, line 2: The start date should be preserved in copyright notices.

commons/com.b2international.index/src/com/b2international/index/es/query/QueryBuilderDeserializer.java
```diff
@@ -0,0 +1,61 @@
...
+
+import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.DeserializationContext;
+import com.fasterxml.jackson.databind.deser.std.StdDeserializer;
+
+/**
+ * @since 8.9
+ */
+public class QueryBuilderDeserializer extends StdDeserializer<org.elasticsearch.index.query.QueryBuilder> {
```
apeteri, line 39: This class is entirely black magic, pretty cool :sparkles:

Is there any official documentation on low-level parsing ES queries? I assume the ES runtime has a populated `NamedXContentRegistry`, do we need to hook that instance instead of an empty one?

commons/com.b2international.index/src/com/b2international/index/es/reindex/ReindexResult.java
```diff
@@ -0,0 +1,246 @@
...
+		return remoteAddress;
+	}
+
+	public boolean isRefresh() {
+		return refresh;
+	}
+
+	@Override
+	public String toString() {
+		return "ReindexResult ["
```
apeteri: `MoreObjects#toStringHelper` has an option where null values are skipped (example copied from their documentation): 

```java
// Returns "ClassName{x=1}"
MoreObjects.toStringHelper(this)
  .omitNullValues()
  .add("x", 1)
  .add("y", null)
  .toString();
```

commons/com.b2international.commons/src/com/b2international/commons/exceptions/ApiException.java
nagyo: It's a tricky one. As far as I can recall (it was a long time ago I made this change), there are two potential places:

One is here: 
https://github.com/b2ihealthcare/snow-owl/blob/2e4ba8f70d044f751ccfe765ed3cdc7cdaa1888c/core/com.b2international.snowowl.core.rest/src/com/b2international/snowowl/core/rest/codesystem/VersionRestService.java#L152

The ApiError.withMessage(String message) method is not concerned with the developer message being set or not.

The other one is here:
https://github.com/b2ihealthcare/snow-owl/blob/2e4ba8f70d044f751ccfe765ed3cdc7cdaa1888c/commons/com.b2international.commons/src/com/b2international/commons/exceptions/ApiErrorException.java#L27-L32

Which is used by the VersionRestService and also by the new HTTP client factory in the other repo (which I cannot reference here).

tests/com.b2international.snowowl.test.dependencies/.classpath
```diff
@@ -1,23 +1,23 @@
...
-	<classpathentry exported="true" kind="lib" path="lib/docker-java-api-3.2.13.jar"/>
-	<classpathentry exported="true" kind="lib" path="lib/docker-java-transport-3.2.13.jar"/>
-	<classpathentry exported="true" kind="lib" path="lib/docker-java-transport-zerodep-3.2.13.jar"/>
-	<classpathentry exported="true" kind="lib" path="lib/duct-tape-1.0.8.jar"/>
-	<classpathentry exported="true" kind="lib" path="lib/elasticsearch-1.17.3.jar"/>
-	<classpathentry exported="true" kind="lib" path="lib/jna-5.8.0.jar"/>
-	<classpathentry exported="true" kind="lib" path="lib/system-rules-1.18.0-sources.jar"/>
 	<classpathentry exported="true" kind="lib" path="lib/testcontainers-1.17.3.jar"/>
+	<classpathentry exported="true" kind="lib" path="lib/system-rules-1.18.0-sources.jar"/>
+	<classpathentry exported="true" kind="lib" path="lib/system-rules-1.18.0.jar" sourcepath="lib/system-rules-1.18.0-sources.jar"/>
```
apeteri: The first entry should not be added as a library, as now it is properly set as a source library.

tests/com.b2international.snowowl.test.dependencies/META-INF/MANIFEST.MF
```diff
@@ -6,19 +6,20 @@ Bundle-Version: 8.8.2.qualifier
...
  lib/commons-compress-1.21.jar,
+ lib/contiperf-2.3.2.jar,
  lib/docker-java-api-3.2.13.jar,
  lib/docker-java-transport-3.2.13.jar,
  lib/docker-java-transport-zerodep-3.2.13.jar,
  lib/duct-tape-1.0.8.jar,
  lib/elasticsearch-1.17.3.jar,
  lib/jna-5.8.0.jar,
+ lib/system-rules-1.18.0.jar,
  lib/system-rules-1.18.0-sources.jar,
```
apeteri: Same as above.

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/bundle/Bundle.java
nagyo: Hmm, I was hesitant with this given if I see `toCreateRequest` I personally would assume I have a request ready to go. 
However, it is a builder where I can change basically anything before building it and sending it away.

I'm fine with both ways though, let's settle on one of them and I'm happy to change the rest. Previously it was not consistent for CodeSystem, Bundle, ValueSet and ConceptMap.

commons/com.b2international.index/src/com/b2international/index/es/client/EsClusterStatus.java
apeteri: Three more files where the date was upgraded in place:

- CodeSystemUpgradeRestService
- AbstractSnomedExtensionApiTest
- ApiTestConstants

apeteri (approved):
👍, I have only minor observations listed below.

commons/com.b2international.commons/src/com/b2international/commons/exceptions/ApiException.java
nagyo: Added safety measure to the ApiErrorException ctor instead

commons/com.b2international.index/src/com/b2international/index/es/client/EsClusterStatus.java
nagyo: License dates and since tags will be covered in a separate task before the release.

commons/com.b2international.index/src/com/b2international/index/es/query/QueryBuilderDeserializer.java
nagyo: It's indeed a magic box and that's why there is no good documentation 😏 And yes we need the empty instance for a reason, I think this is to push the dependencies to the lowest possible, otherwise, more digging would be required in deep ES code.

PR #1111, opened 2023-02-14
Adapt batch size of reindex requests

nagyo:
Automatically retry reindex with a smaller batch size in case the documents are larger than the default 100 Mb on-heap buffer


apeteri (approved):
👍

cmark (approved):
🥓

PR #1116, opened 2023-02-21
Force version creation deletes a branch document even when responds with HTTP 400 Bad Request

cmark:
This PR fixes a very rare issue when a branch document got deleted due to an incorrect force recreation attempt of an earlier/older version.

### DIFF
--- a/commons/com.b2international.index/src/com/b2international/index/revision/BaseRevisionBranching.java
+++ b/commons/com.b2international.index/src/com/b2international/index/revision/BaseRevisionBranching.java
@@ -213,33 +213,34 @@ public abstract class BaseRevisionBranching {
 	 *            - the name of the new child branch, may not be <code>null</code>
 	 * @param metadata
 	 *            - optional metadata map
+	 * @param force - to override an existing non-deleted branch instead of reporting an error 
 	 * @return
 	 * @throws AlreadyExistsException
 	 *             - if the child branch already exists
 	 */
-	public final String createBranch(final String parent, final String name, final Metadata metadata) {
+	public final String createBranch(final String parent, final String name, final Metadata metadata, boolean force) {
 		RevisionBranch parentBranch = getBranch(parent);
 		if (parentBranch.isDeleted()) {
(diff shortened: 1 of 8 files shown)


commons/com.b2international.index/src/com/b2international/index/revision/BaseRevisionBranching.java
```diff
@@ -213,33 +213,34 @@ protected final String toAbsolutePath(final String parentPath, final String name
...
-	private String create(final RevisionBranch parent, final String name, final Metadata metadata) {
+	private String create(final RevisionBranch parent, final String name, final Metadata metadata, boolean force) {
 		// prevents problematic branch creation from multiple threads, but allows them 
 		// to respond back successfully if branch did not exist before creation and it does now
 		final String parentPath = parent.getPath();
 		return locked(parentPath, () -> {
 			// check again and return if exists, otherwise open the child branch
 			final RevisionBranch existingBranch = get(toAbsolutePath(parentPath, name));
-			if (existingBranch != null && !existingBranch.isDeleted()) {
+			if (!force && existingBranch != null && !existingBranch.isDeleted()) {
```
apeteri, line 243: If `force` is set to `true`, and two concurrent calls arrive at the public `createBranch` method above, they will:

- both fail the test in line 228 and land here
- both will request a lock for the parent branch, so they are let in the block starting with line 241 one by one
- each will fail the test in line 243 on its own and go on to line 246, re-creating the branch twice.

cmark: I'm not seeing any problem with that. This force flag (and this method as well) is considered internal API. If two requests or method calls have been made to forcefully recreate the branch then two branches will be created to serve both requests. Eventually the system will remain in a consistent state and only one branch will be available. The only "problem" this could lead to is fetching data with the wrong branch visibility information (due to updated base/head timestamps), but that is very unlikely to happen because commits require a bit more time to be processed and by that time the branch will be overridden with the second branch instance.

nagyo (approved):
LGTM!

apeteri (approved):
👍

PR #1118, opened 2023-02-24
SO-5642: API key expiration

cmark:
This PR adds a new `POST /token` endpoint with the option to configure the `expiration` time of the generated token as well. The old `POST /login` endpoint becomes deprecated.

### DIFF
--- a/core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/identity/DefaultJWTGenerator.java
+++ b/core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/identity/DefaultJWTGenerator.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2019-2022 B2i Healthcare Pte Ltd, http://b2i.sg
+ * Copyright 2019-2023 B2i Healthcare Pte Ltd, http://b2i.sg
  * 
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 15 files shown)


nagyo (approved):
LGTM!

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/identity/DefaultJWTGenerator.java
```diff
@@ -36,22 +40,30 @@ final class DefaultJWTGenerator implements JWTGenerator {
...
+	public String generate(String email, Map<String, Object> claims, String expiration) {
+		
 		Builder builder = JWT.create()
 				.withIssuer(issuer)
 				.withSubject(email)
 				.withIssuedAt(new Date())
 				.withClaim(emailClaimProperty, email);
 		
+		if (!Strings.isNullOrEmpty(expiration)) {
+			TimeValue expirationTimeValue = TimeValue.parseTimeValue(expiration, "expiration");
```
apeteri: A check for non-negative and "practical" expiration intervals could be added here in the future 🔮

apeteri (approved):
LGTM!

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/identity/DefaultJWTGenerator.java
cmark: Added a test case and the check to the generator.

PR #1122, opened 2023-03-16
SO-5684: eliminate superfluous branch fetch request

cmark:
This PR changes the core branch access request flow to eliminate a superfluous, unnecessary branch fetch request to gain a small speed improvement when performing requests.


commons/com.b2international.index/src/com/b2international/index/revision/DefaultRevisionIndex.java
```diff
@@ -301,7 +300,7 @@ public void removeHook(Hook hook) {
 	 * @return
 	 */
 	List<Hooks.Hook> getHooks() {
-		return ImmutableList.copyOf(hooks);
+		return List.copyOf(hooks);
```
apeteri, line 303: This conversion is fine, but there is a "fun" difference between `ImmutableSet#copyOf` and `Set#copyOf` &ndash; the latter uses hashing and iteration order is not consistent.

commons/com.b2international.index/src/com/b2international/index/revision/RevisionBranchRef.java
```diff
@@ -18,120 +18,133 @@
...
 	}
-	
+
 	@Override
 	public String toString() {
-		return MoreObjects.toStringHelper(this)
-				.add("branchId", branchId)
-				.add("segments", segments)
-				.toString();
+		return MoreObjects.toStringHelper(this).add("branchId", branchId).add("segments", segments).toString();
```
apeteri: We could list all branches as well as the "deleted" flag here.

apeteri (approved):
👍

commons/com.b2international.index/src/com/b2international/index/revision/RevisionBranchRef.java
cmark: Done

nagyo (approved):
LGTM!

PR #1123, opened 2023-03-17
Introduce utility class to manage files with the NIO2 API

nagyo:


cmark (approved):
🥓

commons/com.b2international.commons/src/com/b2international/commons/io/PathUtils.java
```diff
@@ -0,0 +1,270 @@
...
+	 * Delete all files within the specified directory except the exclusions specified. If there are none, the directory is deleted as well.
+	 * The exclusions must be declared as full paths.
+	 */
+	public static void deleteDirectory(final Path directory, Collection<String> exclusions) throws IOException {
+
+		checkArgument(Files.isDirectory(directory), "The specified path is not a directory");
+
+		final List<Path> filteredPaths = Files.walk(directory)
+				.sorted(Comparator.reverseOrder())
+				.filter(path -> exclusions.stream().noneMatch(p -> p.equals(path.toString())))
```
apeteri: If `exclusions` held a `Collection` of `Path` instances, the check would honor file system specific conditions, eg. case sensitivity, according to `Path#equals`'s javadoc. Which may or may not be a good thing. 😄

commons/com.b2international.commons/src/com/b2international/commons/io/PathUtils.java
```diff
@@ -0,0 +1,270 @@
...
+				.collect(toList());
+
+		for (final Path path : filteredPaths) {
+
+			// do not try to delete the containing directory when there are excluded files present inside
+			if (path.equals(directory) && Files.list(directory).findFirst().isPresent()) {
```
apeteri: I think this should be checked for any directory, not just the initial one. Or do we expect the sub-directory itself to show up in `exclusions` in these cases?

apeteri (approved):
🍃 with a small note.

commons/com.b2international.commons/src/com/b2international/commons/io/PathUtils.java
nagyo: That's a good idea, will change it accordingly!

commons/com.b2international.commons/src/com/b2international/commons/io/PathUtils.java
nagyo: Good point, fixing it.

PR #1124, opened 2023-03-18
SNRAY-1244: authorization performance improvements

cmark:
This PR adds a simple authorization shortcut to each request where the request only cares about a single resource and it does not need to fetch all visible resources. This yields 50% performance gain on not cacheable requests (`100ms` -> `50ms`) while a massive 80% on cacheable requests (`50ms` -> `10ms`).


apeteri (approved):

nagyo (approved):
LGTM!

PR #1125, opened 2023-03-20
Introduce utility class to manage files with the NIO2 API (8.x)

nagyo:
Pair of #1123


cmark (approved):
🥓

commons/com.b2international.commons/src/com/b2international/commons/io/PathUtils.java
```diff
@@ -0,0 +1,271 @@
...
+
+		final Set<Path> matches = newHashSet();
+
+		final PathMatcher pathMatcher = fileSystem.getPathMatcher("glob:**/" + pathMatcherExpression);
+
+		for (final Path directory : directoriesToIterate) {
+			if (recursive) {
+				Files.walk(directory).filter(path -> pathMatcher.matches(path)).forEach(matches::add);
+			} else {
+				Files.list(directory).filter(path -> pathMatcher.matches(path)).forEach(matches::add);
```
apeteri, line 246: Since both `Files.walk(directory)` and `Files.list(directory)` return `Stream<Path>`, the following common portion of each line could be moved after the `if` block:

```java
final Stream<Path> paths;

if (recursive) { 
  paths = Files.walk(directory);
} else { 
  paths = Files.list(directory);
}

paths.filter(pathMatcher::matches).forEach(matches::add);
```

apeteri (approved):
👍

PR #1130, opened 2023-03-28
Support `Accept-Language: "*"` headers

cmark:
This PR adds support for the wildcard-only `Accept-Language` header which is being set by most HTTP clients when not explicitly specified when making the HTTP request.
We do not intend to support mixed `Accept-Language` header values where both wildcard and non-wildcard language tags are present (eg. `de,*`). Using such a header value will still result in an HTTP 400 Bad Request response.

### DIFF
--- a/snomed/com.b2international.snowowl.snomed.core.rest.tests/src/com/b2international/snowowl/snomed/core/rest/components/SnomedConceptSearchApiTest.java
+++ b/snomed/com.b2international.snowowl.snomed.core.rest.tests/src/com/b2international/snowowl/snomed/core/rest/components/SnomedConceptSearchApiTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2021-2022 B2i Healthcare Pte Ltd, http://b2i.sg
+ * Copyright 2021-2023 B2i Healthcare Pte Ltd, http://b2i.sg
  * 
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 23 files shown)


snomed/com.b2international.snowowl.snomed.core.rest.tests/src/com/b2international/snowowl/snomed/core/rest/components/SnomedConceptSearchApiTest.java
```diff
@@ -393,6 +394,18 @@ public void multipleNestedModuleExpand() {
 			.body("items[0].descriptions.items.module.id", allOf(not(emptyIterable()), everyItem(equalTo(Concepts.MODULE_SCT_CORE))))
 			.body("items[0].relationships.items.module.id", allOf(not(emptyIterable()), everyItem(equalTo(Concepts.MODULE_SCT_CORE))));
 	}
+	
+	@Test
+	public void wildcardAcceptLanguageHeader() throws Exception {
```
apeteri, line 399: The wildcard header was not added to the test request, it seems 👀

apeteri (approved):
Needs a small adjustment but otherwise LGTM!

snomed/com.b2international.snowowl.snomed.core.rest.tests/src/com/b2international/snowowl/snomed/core/rest/components/SnomedConceptSearchApiTest.java
cmark: Right, fixing.

cmark: Done

nagyo (approved):
LGTM!

PR #1135, opened 2023-04-20
Port to 8.x: SO-4964 Id registration in nested create requests

AAAlinaaa:
8.x port of PR https://github.com/b2ihealthcare/snow-owl/pull/945.


cmark (approved):
🥓

PR #1137, opened 2023-04-21
SO-5439: Fix dependency resolution during SNOMED CT import

nagyo:


snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/request/SnomedOWLExpressionConverter.java
```diff
@@ -56,7 +61,9 @@
 public final class SnomedOWLExpressionConverter {
 
 	private static final Logger LOG = LoggerFactory.getLogger(SnomedOWLExpressionConverter.class);
-
+	
+	private static final Pattern DIGIT_PATTERN = Pattern.compile("\\d+");
```
apeteri: Can you include the leading `:` in this regexp? The numeric part can be captured in a group.

apeteri (approved):
👍 with a small request.

cmark (commented):
Late 🥓

PR #1150, opened 2023-05-03
Resolve issue when deleting directories with multiple sub-directories (7.x)

nagyo:
If no exclusion filters are provided for PathUtils.deleteDirectory() then every file and sub-directory must be deleted


apeteri (approved):
🍃

cmark (approved):
🥓

PR #1154, opened 2023-05-05
Short ipv6 format parsing from doc values

cmark:
This PR fixes an issue where commit documents could not be parsed when the `mergeSource` field was requested in the request. ES/Lucene stores the short form of the IP field value even if we use the full long form in the `_source`. Added a temporary hack to parse into InetAddress which adds zeros to all `::` short values and then convert that to `RevisionBranchPoint` instances.

Related to https://snowowl.atlassian.net/browse/SO-5774


apeteri (approved):
🍏

nagyo (approved):
🍏

PR #1160, opened 2023-05-11
SO-4901: identical component conflict

cmark:
This PR improves the automatic conflict resolution to allow identical objects to be added to both sides of the merge. This earlier resulted in a conflict now it passes through.


apeteri (approved):
👍

nagyo (approved):
🍏

PR #1161, opened 2023-05-11
SO-3933: published vs unpublished snapshot export

cmark:
This PR fixes an RF2 Snapshot export issue when the RF2 export was missing some of the published content when the configuration was not to include pre-release content but the reference branch contained some.

### DIFF
--- a/snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/request/rf2/SnomedRf2ExportRequest.java
+++ b/snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/request/rf2/SnomedRf2ExportRequest.java
@@ -88,6 +88,7 @@ import com.b2international.snowowl.snomed.datastore.request.SnomedRefSetMemberSe
 import com.b2international.snowowl.snomed.datastore.request.SnomedRequests;
 import com.b2international.snowowl.snomed.datastore.request.rf2.exporter.*;
 import com.fasterxml.jackson.annotation.JsonProperty;
+import com.google.common.base.Preconditions;
 import com.google.common.base.Strings;
 import com.google.common.collect.*;
 import com.google.common.collect.ImmutableList.Builder;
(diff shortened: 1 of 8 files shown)


snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/request/rf2/SnomedRf2ExportRequest.java
```diff
@@ -484,7 +486,15 @@ private List<String> computeBranchesToExport(final String referenceBranch, final
 				}
 				break;
 			case SNAPSHOT:
-				branchesToExport.add(referenceBranch);
+				// in case we are exporting published content only use the version branch instead of the reference branch
+				if (!this.includePreReleaseContent) {
+					if (versionsToExport.isEmpty()) {
+						throw new BadRequestException("Snapshot export without unpublished components requires at least one version.");
+					}
+					branchesToExport.add(versionsToExport.first().getBranchPath());
```
apeteri: `versionsToExport` is a `TreeSet` with an ascending effective time sort order. I think we need the element with the greatest value here, not the smallest.

To confirm, add at least one more version to the test case that makes a different change on the component.

apeteri (changes requested):

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/request/rf2/SnomedRf2ExportRequest.java
cmark: Yes, you are right I thought that is a TreeSet but the versions are in reverse order because of the ordering we use when fetching the versions. The Ordering at the top of the class, unfortunately, orders them by chronological effective time order. To reflect this, I fixed the issue, renamed the variable, and updated the test case with an additional version and component.

apeteri (approved):
LGTM!

nagyo (approved):
🍏

PR #1166, opened 2023-05-31
SO-5811: fix FHIR publisher field selection

cmark:
This PR resolves a bug when trying to load resource documents with the field selection of the settings field, which is a Map type field. Previously the system was unable to detect that this field cannot be loaded from doc_values, and unfortunately Elasticsearch does not throw an error when trying to load unindexed/unsupported/dynamic fields from doc_values.

### DIFF
--- a/commons/com.b2international.index/src/com/b2international/index/mapping/DocumentMapping.java
+++ b/commons/com.b2international.index/src/com/b2international/index/mapping/DocumentMapping.java
@@ -309,6 +309,11 @@ public final class DocumentMapping {
 		final Class<?> fieldType = getFieldType(field);
 		return Set.class.isAssignableFrom(fieldType) || PrimitiveSet.class.isAssignableFrom(fieldType);
 	}
+	
+	public boolean isMap(String field) {
+		final Class<?> fieldType = getFieldType(field);
+		return Map.class.isAssignableFrom(fieldType);
+	}
 
 	private static boolean isCollection(Class<?> fieldType) {
 		return Iterable.class.isAssignableFrom(fieldType) || PrimitiveCollection.class.isAssignableFrom(fieldType) || fieldType.isArray();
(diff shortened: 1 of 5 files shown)


commons/com.b2international.index/src/com/b2international/index/mapping/DocumentMapping.java
```diff
@@ -309,6 +309,11 @@ public boolean isSet(String field) {
 		final Class<?> fieldType = getFieldType(field);
 		return Set.class.isAssignableFrom(fieldType) || PrimitiveSet.class.isAssignableFrom(fieldType);
 	}
+	
+	public boolean isMap(String field) {
```
apeteri, line 313: Does `isObject` need to include maps? It is only used in `handleChangedInSourceAndTarget` which calls `handleObjectConflict`, and that sounds like it would/should work for this type as well: https://github.com/b2ihealthcare/snow-owl/blob/8.x/commons/com.b2international.index/src/com/b2international/index/revision/RevisionConflictProcessor.java#L129-L152

apeteri (changes requested):

commons/com.b2international.index/src/com/b2international/index/mapping/DocumentMapping.java
cmark: Yes, good catch.

apeteri (approved):

commons/com.b2international.index/src/com/b2international/index/mapping/DocumentMapping.java
apeteri: This will be resolved in a separate PR.

cmark: Let's keep the scope of this PR to the publisher field and will fix the other isObject method in a different PR.

nagyo (approved):
🍏

PR #1169, opened 2023-06-02
fix(api): ensure the same term filtering behavior as in the SNOMED CT...

cmark:
...Concept Search API

Always apply locale filtering.


apeteri (approved):
LGTM!

nagyo (approved):
🍏

PR #1177, opened 2023-06-23
fix(validation): Return no issues if in scope refsets can't be found

AAAlinaaa:


apeteri (approved):
LGTM!

Please apply the same diff to the other MRCM validation rule script as well so it doesn't spend too much time running when it becomes clear that no MRCM rule reference sets apply to the module.

apeteri:
Requesting second review.

nagyo (approved):
🍏

cmark (approved):
🥓

PR #1180, opened 2023-06-27
MRCM range validation rule fix

AAAlinaaa:


snomed/com.b2international.snowowl.validation.snomed/src/main/resources/scripts/rule_mrcm_constraint.groovy
```diff
@@ -95,20 +96,20 @@ for (String typeId : allowedRanges.keySet()) {
...
+		} catch (SyntaxException e) {
+			ctx.log().warn("Could not parse clause ${rangeClause}");
 			ctx.log().warn(e.getMessage());
 			continue;
 		}
 		
+		String clause = rangeClause;
+		
+		if (clause.contains("|")) {
+			clause = rangeClause.substring(0, rangeClause.indexOf('|')).strip();
```
apeteri, line 110: The parsed representation above could be put to good use to remove all human-readable labels from the expression. I'm afraid the string cutting logic will break the moment someone adds a range clause with a refinement or filter, eg. `<< 404684003 |Clinical finding| {{ C definitionStatus = primitive }}`.

apeteri (changes requested):

apeteri (approved):
I'm going to approve as if we want to keep index-based querying (especially in the case of relationship snippets on OWL axiom members) for performance reasons, range expressions are not allowed to get more complex than the constructs currently supported.

nagyo (approved):
I'm also not entirely happy with manually parsing ECL expressions where there would be a better programmatic approach, but I'll accept the fix, otherwise, every validation run would fail if a malformed expression is present.

PR #1181, opened 2023-06-28
fix(fhir): add a separate `X-Owner` header to configure ownership...

cmark:
...of a FHIR resource when creating it via FHIR API

Rename `X-Author-Profile-Name` to `X-Owner-Profile-Name`.


nagyo (approved):
🍏

apeteri (approved):
👍

PR #1182, opened 2023-06-29
Collect nested requests for permission checking recursively in TransactionalRequest and SnomedBulkRequest

apeteri:
These usually appear with one packed into the other; if `TransactionRequest` stops at `SnomedBulkRequest`, there might not be enough information collected for permission checks.


nagyo (approved):
LGTM!

cmark (approved):
🥓

PR #1184, opened 2023-06-29
fix(index): support loading unindexed fields from documents

cmark:
This resolves problems around exporting query type reference set to RF2.


nagyo (approved):
LGTM!

apeteri (approved):
👍

apeteri:
For nested docs `@Doc(index = false)` is also allowed. Do we want to handle these as well?

cmark:
> For nested docs `@Doc(index = false)` is also allowed. Do we want to handle these as well?

No need for it now, I think. We can introduce it in another PR if needed.

PR #1185, opened 2023-07-05
feat(datastore.snomed): Improve suggested changes in SnomedQueryOptimizer

apeteri:


snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/core/request/SnomedHierarchyStats.java
```diff
@@ -0,0 +1,360 @@
...
+import com.b2international.snowowl.snomed.datastore.index.entry.SnomedRelationshipIndexEntry;
+import com.b2international.snowowl.snomed.datastore.request.SnomedRequests;
+import com.google.common.collect.HashMultiset;
+import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.Multiset;
+
+/**
+ * @since 8.11.0
+ */
+public record SnomedHierarchyStats(
```
cmark, line 48: :woah:

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/core/request/SnomedQueryOptimizer.java
```diff
@@ -15,169 +15,789 @@
...
+import com.google.common.primitives.Floats;
 import com.google.common.primitives.Ints;
-import com.google.common.util.concurrent.UncheckedExecutionException;
 
 /**
  * @since 7.7
  */
 public final class SnomedQueryOptimizer implements QueryOptimizer {
 
+	static final int PAGE_SIZE = 10_000;
```
cmark: It would be ideal to replace this with the new constant.

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/core/request/SnomedHierarchyStats.java
```diff
@@ -0,0 +1,357 @@
...
+
+		/*
+		 * Remove "SNOMED CT Concept" and "Clinical finding" as it is unlikely that they
+		 * will turn out to be good replacements (and computing descendant counts for
+		 * them is costly)
+		 */
+		positiveDescendantsAndSelf.elementSet().remove(Concepts.ROOT_CONCEPT);
+		positiveChildren.elementSet().remove(Concepts.ROOT_CONCEPT);
+		
+		positiveDescendantsAndSelf.elementSet().remove(CLINICAL_FINDING);
```
cmark, line 183: Would it make sense to treat all top-level concepts like this?

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/core/request/SnomedHierarchyStats.java
```diff
@@ -0,0 +1,357 @@
...
+		return noFalsePositives.stream()
+			.map(ancestorId -> {
+				final boolean member = memberIds.contains(ancestorId);
+				final String operator = member ? "<<" : "<";
+				return new QueryExpression(IDs.base62UUID(), String.format("%s %s", operator, ancestorId), false);
+			})
+			.collect(Collectors.toList());
+	}
+
+	public LongKeyFloatMap initAncestorScores(
```
cmark: It would be great to add minimal javadoc and explanation to these public methods.

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/core/request/SnomedRelationshipStats.java
```diff
@@ -0,0 +1,165 @@
...
+import com.b2international.snowowl.snomed.common.SnomedConstants.Concepts;
+import com.b2international.snowowl.snomed.core.domain.SnomedRelationship;
+import com.b2international.snowowl.snomed.core.domain.SnomedRelationships;
+import com.b2international.snowowl.snomed.datastore.index.entry.SnomedRelationshipIndexEntry;
+import com.b2international.snowowl.snomed.datastore.request.SnomedRequests;
+import com.google.common.collect.HashBasedTable;
+import com.google.common.collect.Table;
+
+/**
+ * @since 8.11.0
```
cmark: Please update the since tags to 8.12.0.

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/core/request/SnomedQueryOptimizer.java
```diff
@@ -15,169 +15,847 @@
...
+
+	public enum OptimizerStrategy {
```
cmark, line 80: Add a since tag here.

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/core/request/SnomedQueryOptimizer.java
```diff
@@ -15,169 +15,847 @@
...
+		DEFAULT,
+		
+		SCORE_BOOST_1 {
+			// Boost scores with the fraction of the precision of direct children, for selected candidates
```
cmark: Turn these comments to enum literal javadoc, please.

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/core/request/SnomedQueryOptimizer.java
```diff
@@ -15,169 +15,847 @@
...
+		LOSSY {
+			// Allow some false positives in lossy optimization mode
+			@Override 
+			public float getFalsePositiveThreshold(float value) { 
+				return value; 
+			}
+		};
+		
+		public float getFalsePositiveThreshold(final float value) {
+			// Default behavior is to disallow false positives
```
cmark: Same applies to these, but as method javadoc.

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/core/request/SnomedQueryOptimizer.java
```diff
@@ -15,169 +15,847 @@
...
+	}
+
+	@FunctionalInterface
+	public interface EclEvaluator {
+
+		EclEvaluator DEFAULT = (context, eclExpression, pageSize) -> SnomedRequests.prepareSearchConcept()
+			.setLimit(pageSize)
+			.filterByEcl(eclExpression)
+			.setFields(SnomedConceptDocument.Fields.ID)
+			.sortBy(SnomedConceptDocument.Fields.ID)
```
cmark: Just a note: the `id` field is the default sort method for all revision types.

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/core/request/SnomedQueryOptimizer.java
```diff
@@ -15,169 +15,847 @@
...
+
+			if (mustNotQueryExpressions.isEmpty()) {
+				mustNotQueryExpressions = null;
+			}
+
+			return CodeSystemRequests.prepareSearchConcepts()
+				.filterByCodeSystemUri(resourceUri)
+				.filterByInclusions(queryExpressions)
+				.filterByExclusions(mustNotQueryExpressions)
+				.setLimit(PAGE_SIZE)
```
cmark: It looks like only the ID is needed here, please add the `id` field selection to this request.

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/core/request/SnomedQueryOptimizer.java
```diff
@@ -15,169 +15,847 @@
...
+	private final int idealClauseCount         = 1;
+	private final int maxIteration             = 1000;
+	private final Duration maxRuntime          = Duration.ofSeconds(580L);
+
+	private OptimizerStrategy optimizerStrategy = OptimizerStrategy.DEFAULT;
+	private int zoom                            = 1;
+	private int maxClauseCount                  = 2000;
+
+	// Run-time fields
+	private Set<String> conceptSet;
```
cmark, line 303: This is a server-side singleton instance in the current implementation (see `SnomedPlugin`). Running multiple optimizations at the same time will mess up the internals. Please refactor. Feel free to create a new instance per request.

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/core/request/SnomedQueryOptimizer.java
```diff
@@ -15,169 +15,847 @@
...
-			if (nonPinnedMembersForParent.size() <= 1) {
-				continue;
+		});
+
+		newInclusions.clear();
+	}
+
+	private void filterRefinementsForInclusion(final SnomedRelationshipStats relationshipStats) {
+		// a) Less than 95% precision
+		relationshipStats.filterByPrecision(0.95f);
```
cmark, line 654: Would it make sense to move any of these magic numbers to constants? Or have them configurable?

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/core/request/SnomedQueryOptimizer.java
```diff
@@ -15,169 +15,847 @@
...
+
+		// See if concepts still to be processed can be converted to "* : c1 = c2" expressions
+		final SnomedRelationshipStats inclusionRelationshipStats = SnomedRelationshipStats.create(
+			context, 
+			conceptsToInclude, 
+			relationshipSearchBySource, 
+			relationshipSearchByTypeAndDestination);
+
+		filterRefinementsForInclusion(inclusionRelationshipStats);
+		final List<QueryExpression> refinementInclusions = inclusionRelationshipStats.optimizeRefinements(context);
```
cmark: This will always try to optimize using refinements first, right? Isn't that costly? I mean if there are 10k conceptIds then we always start with a relationship fetch according to this method call. Would it be possible to move this to a configuration value and check how long it takes to optimize a 1k and 10k refset with refinements enabled and disabled? Thanks!

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/core/request/SnomedQueryOptimizer.java
```diff
@@ -15,169 +15,847 @@
...
+
+		// Pinned inclusions are checked for redundancy but otherwise should appear unmodified in optimized output 
+		final List<QueryExpression> pinnedInclusions = optimizePinned(context, inclusions);
+		log.info("Found {} unique pinned inclusion(s)", pinnedInclusions.size());
+		applyInclusions(context, pinnedInclusions);
+
+		// See if concepts still to be processed can be converted to "* : c1 = c2" expressions
+		final SnomedRelationshipStats inclusionRelationshipStats = SnomedRelationshipStats.create(
+			context, 
+			conceptsToInclude,
```
cmark: This line could be `conceptSet` here to be consistent with the other calls and I think by doing that the comment on line 365 is not needed anymore.

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/core/request/SnomedQueryOptimizer.java
```diff
@@ -15,169 +15,847 @@
...
+				// "applyInclusions" is effectively inlined here because we need access to the number of concepts included
+				final Set<String> includedIds = evaluateEclToSet(context, bestFitInclusion.getQuery());
+				final Set<String> expected = Sets.intersection(includedIds, conceptSet);
+				conceptsToInclude.removeAll(expected);
+
+				final Set<String> unexpected = Sets.difference(includedIds, conceptSet);
+				conceptsToExclude.addAll(unexpected);
+
+				// Tune parameters after a successful selection
+				if (((float) conceptSet.size()) / (includedIds.size() + 1) > 1000.0f) {
```
cmark: What is 1000 here? 😕

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/core/request/SnomedQueryOptimizer.java
```diff
@@ -15,169 +15,847 @@
...
+					final float newFitThreshold = fitThreshold * 0.9f;
+					if (newFitThreshold > minimumFitThreshold) {
+						fitThreshold = newFitThreshold;
+						zoom = idealClauseCount;
+						log.info("Fit threshold changed to {}, zoom is {}", fitThreshold, zoom);
+					}
+				}
+
+				// Elevate strategy after using the same one for 100 iterations (but don't promote to LOSSY from non-lossy strategies)
+				if (iteration > (optimizerStrategy.ordinal() + 1) * 100 && optimizerStrategy.ordinal() < OptimizerStrategy.values().length - 2) {
```
cmark: Ohmy, enum ordinal magic, would it be possible to refactor this to something readable? 

Something like this for example:
```
if (optimizerStrategy.isElevateNeeded(iterations)) {
  optimizerStrategy = optimizerStrategy.elevate();
}
```
And document elevation strategy in the enum type.
Thanks!

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/core/request/SnomedQueryOptimizer.java
```diff
@@ -15,169 +15,847 @@
...
+					log.info("Optimizer strategy changed to {} after {} iterations", optimizerStrategy, iteration);
+				}
+
+			} else {
+				// No acceptable ancestor in this round
+
+				if (zoom <= maxClauseCount && zoom <= conceptSet.size() && zoom != Math.round(zoom * 1.1f)) {
+					// Try adjusting zoom first
+					zoom = Math.round(zoom * 1.1f);
+					log.info("Zoom changed to {}", zoom);
```
cmark: Awesome log messages, but it would be great to turn them to trace messages before we merge this PR to 8.x.

cmark (changes requested):
Looks great, but there are things to do.

Could you please update the document describing the summarized version of the algorithm?
Also, could you please measure the memory usage of the new algorithm for a few refsets, including larger ones with more than 10k single concept clauses? Thanks!

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/core/request/SnomedQueryOptimizer.java
apeteri: We can expose any of the parameters as `Option` keys, and pick up defaults from eg. `SomedCoreConfiguration`, but I'm not sure who would want to adjust them. An "advanced optimize" dialog would look like a studio mixing board.

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/core/request/SnomedHierarchyStats.java
apeteri: These were the costliest to even consider given their large descendant count. `373873005 |Pharmaceutical / biologic product|` might be worth adding as another exception, but I'd leave it open for other top-level concepts.

If we want to handle this dynamically and eg. exclude all candidates above a certain amount of descendants (100 000, for example), a speed-up is available in PR https://github.com/b2ihealthcare/snow-owl/pull/1186.

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/core/request/SnomedQueryOptimizer.java
apeteri: IIRC this distinction was made in the reference algorithm as well for some reason. Ancestors are allowed to cover an included concept multiple times.

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/core/request/SnomedQueryOptimizer.java
apeteri: If every clause would cover as many concepts from the evaluated concept set as the one that we have just found, we'd have about 1000 clauses. This is an indication that the iteration progresses too slowly.

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/core/request/SnomedQueryOptimizer.java
apeteri: Runtimes are in milliseconds:

Scenario*      | 1k disabled | 1k enabled | 10k disabled | 10k enabled
-------------- | ------------ | ------------- | -------------- | -------------
Attempt 1     | 7820           | 8097            | 59672            | 63078
Attempt 2     | 7684           | 8029            | 61261            | 61407
Attempt 3     | 7700           | 7961            | 59968            | 62501
**Average**  | 7734.67      | 8029            | 60300.33       | 62328.67

<sup>*Actual member counts were 932 (VS 999029411000230108) and 9734 (VS 999021651000230109)</sup>

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/core/request/SnomedQueryOptimizer.java
apeteri: I left the initial and final messages on "info" level and turned the rest to "trace".

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/core/request/SnomedQueryOptimizer.java
apeteri: I added comments in https://github.com/b2ihealthcare/snow-owl/commit/ce9898a94701d38b5633bef0a0b2e907a573a4bf to clarify.

apeteri:
With the 1k value set, heap usage is around 300 MB:

image

I could not generate a chart for the 10k version, because with instrumentation the optimizing process takes a lot longer to run. But it seems that one goes up to 1.7 GB of heap usage.

cmark (approved):
🥓

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/core/request/SnomedQueryOptimizer.java
cmark: It's okay to move them to named constants. I'm unsure whether having them configurable externally would be great or not.

cmark:
> With the 1k value set, heap usage is around 300 MB:
> 
> image
> 
> I could not generate a chart for the 10k version, because with instrumentation the optimizing process takes a lot longer to run. But it seems that one goes up to 1.7 GB of heap usage.

Hmm, that's a lot 😄 Would it be possible to identify areas where memory consumption could be lowered a bit? No need to refactor anything now, just mark variables, statements with a comment where optimization could take place.

AAAlinaaa (approved):

apeteri:
I suspect most of that is going in the ECL evaluation cache, as the process invokes the evaluator with different expressions in quick succession. I will need to fire up a profiler to dig into allocation details.

cmark:
> I suspect most of that is going in the ECL evaluation cache, as the process invokes the evaluator with different expressions in quick succession. I will need to fire up a profiler to dig into allocation details.

Did you measure this with a dedicated Elasticsearch instance or in embedded mode?

apeteri:
ES was running separately, I'm thinking of `SnomedConceptRequestCache` here... maybe?

cmark:
> `SnomedConceptRequestCache`

That's only for data expansion queries, ECL caches are outside of Snow Owl.

PR #1187, opened 2023-07-10
feat(snomed.datastore): Adjust page size dynamically for streaming queries

apeteri:
The limit should never exceed the value of `result_window` configured for Elasticsearch.

### DIFF
--- a/commons/com.b2international.index.tests/src/com/b2international/index/StreamTest.java
+++ b/commons/com.b2international.index.tests/src/com/b2international/index/StreamTest.java
@@ -105,10 +105,9 @@ public class StreamTest extends BaseIndexTest {
 		
 		Stopwatch w = Stopwatch.createStarted();
 		
-		final int pageSize = Integer.parseInt((String) getIndexSettings().getOrDefault(IndexClientFactory.RESULT_WINDOW_KEY, "" + IndexClientFactory.DEFAULT_RESULT_WINDOW));
 		final Query<Data> query = Query.select(Data.class)
 			.where(Expressions.matchAll())
-			.limit(pageSize)
+			.limit(1000)
 			.build();
 		
 		stream(query).forEachOrdered(page -> hits.addAll(page.getHits()));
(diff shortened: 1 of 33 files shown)


nagyo (approved):
LGTM!

commons/com.b2international.index.tests/src/com/b2international/index/StreamTest.java
```diff
@@ -104,9 +104,11 @@ public void searchAndReturnAllHitsWithStream() throws Exception {
 		List<Data> hits = newArrayListWithExpectedSize(NUM_DOCS);
 		
 		Stopwatch w = Stopwatch.createStarted();
+		
+		final int pageSize = Integer.parseInt((String) getIndexSettings().getOrDefault(IndexClientFactory.RESULT_WINDOW_KEY, "" + IndexClientFactory.DEFAULT_RESULT_WINDOW));
 		final Query<Data> query = Query.select(Data.class)
 			.where(Expressions.matchAll())
-			.limit(1000)
+			.limit(pageSize)
```
cmark: I think the purpose here was to actually page through the results with the page size of `1000`, now we are going to fetch it in one go, right?

commons/com.b2international.index/src/com/b2international/index/es/EsDocumentSearcher.java
```diff
@@ -194,7 +194,7 @@ public <T> Hits<T> search(Query<T> query) throws IOException {
 		final boolean isLiveStreaming = !Strings.isNullOrEmpty(query.getSearchAfter());
 		if (isLocalStreaming) {
 			checkArgument(!isLiveStreaming, "Cannot use searchAfter when requesting more items (%s) than the configured result window (%s).", limit, resultWindow);
-		} else if (isLiveStreaming) {
+		} if (isLiveStreaming) {
```
cmark: This block looks a bit weird.

cmark (changes requested):

cmark (approved):
🥓

PR #1189, opened 2023-07-17
fix(index): Make bulk "index by query" operations configurable

apeteri:
The magic value `4` is lifted to an index setting named `indexByQueryConcurrencyLevel`.


nagyo (approved):
LGTM!

cmark (approved):
🥓

PR #1190, opened 2023-07-21
Rule 664 - Fix iteration through unpublished components

AAAlinaaa:


snomed/com.b2international.snowowl.validation.snomed/src/main/resources/scripts/rule664.groovy
```diff
@@ -89,13 +89,11 @@ if (params.isUnpublishedOnly) {
 		)
 		.limit(10_000)
 		.build())
-		.each { Hits<String[]> publishedTermsBatch ->
-			// all returned terms are duplicate of an unpublished term
-			publishedTermsBatch.each { publishedTerm ->
+		.each { publishedTerm ->
```
apeteri: We have discussed this in person, but I still can't see how this would work. Groovy's extension method simply calls `iterator()` on the `Stream` using reflection (it has no dedicated version for handling instances of `Stream`):

```groovy
/**
 * Iterates through an aggregate type or data structure,
 * passing each item to the given closure.  Custom types may utilize this
 * method by simply providing an "iterator()" method.  The items returned
 * from the resulting iterator will be passed to the closure.
 *
 * @param self    the object over which we iterate
 * @param closure the closure applied on each element found
 * @return the self Object
 * @since 1.0
 */
public static <T> T each(T self, Closure closure) {
  each(InvokerHelper.asIterator(self), closure);
  return self;
}
```

There is no `flatMap`-like behavior that would unpack instances of `Hits<String[]>` into individual `String[]` elements.

I have also added `println`s in the closure and the output was as follows:

```
[2023-07-28T08:16:21.059] INFO  Worker-0: validationJob                          validation                                                       Executing rule '664'...
Query starts
Published term is Hits{hits=[[Ljava.lang.String;@78b51bb1], limit=10000, total=1, searchAfter=AoEpNjY0MjI0MDE0}
publishedTerm[0] is [664224014, 271781008]
publishedTerm[1] is null
Query finished
```

...so `contains` will check whether an array is a member of set containing single `String`s (the ID of active concepts).

apeteri (changes requested):

snomed/com.b2international.snowowl.validation.snomed/src/main/resources/scripts/rule664.groovy
```diff
@@ -92,12 +92,13 @@ if (params.isUnpublishedOnly) {
 		.each { Hits<String[]> publishedTermsBatch ->
 			// all returned terms are duplicate of an unpublished term
 			publishedTermsBatch.each { publishedTerm ->
-				if (activeConceptIds.get().contains(publishedTerm[1])) {
-					issues.add(ComponentIdentifier.of(SnomedDescription.TYPE, publishedTerm[0]))
-				}
+				String term = publishedTerm[0];
```
apeteri: Now `publishedTerm[1]` is completely unused, and it is not checked whether this (active) description is sitting on an active concept or not.

snomed/com.b2international.snowowl.validation.snomed/src/main/resources/scripts/rule664.groovy
```diff
@@ -92,12 +92,13 @@ if (params.isUnpublishedOnly) {
...
+				descriptionsByTerm.get(term).forEach({ descriptionId -> 
+					//Report the unpublished term
+					issues.add(ComponentIdentifier.of(SnomedDescription.TYPE,  descriptionId))
```
apeteri: Also "unpublished vs. unpublished" cases get two validation issues, whereas "unpublished vs. published" might only get one. You could also report the description ID that is returned by the query &ndash; except it is no longer part of the selected fields.

apeteri (changes requested):

snomed/com.b2international.snowowl.validation.snomed/src/main/resources/scripts/rule664.groovy
AAAlinaaa: True...

snomed/com.b2international.snowowl.validation.snomed/src/main/resources/scripts/rule664.groovy
AAAlinaaa: Modified this to report both in case of unpublished only runs

apeteri (approved):
LGTM!

snomed/com.b2international.snowowl.validation.snomed/src/main/resources/scripts/rule664.groovy
```diff
@@ -94,10 +94,14 @@ if (params.isUnpublishedOnly) {
 			publishedTermsBatch.each { publishedTerm ->
 				if (activeConceptIds.get().contains(publishedTerm[1])) {
 					issues.add(ComponentIdentifier.of(SnomedDescription.TYPE, publishedTerm[0]))
+					String term = publishedTerm[2];
```
cmark, line 97: Please add a test case for this change that fails without this change. Thanks.

cmark (changes requested):

cmark (approved):
🥓

PR #1197, opened 2023-08-29
SO-5843: Support high-level resource comparisons

apeteri:


snomed/com.b2international.snowowl.snomed.core.rest.tests/src/com/b2international/snowowl/snomed/core/request/SnomedDependencyCompareTest.java
```diff
@@ -0,0 +1,359 @@
...
+		assertThat(compareResult.getToUri().getResourceUri()).isEqualTo(toUri);
+
+		assertThat(compareResult.getItems())
+			.extracting(AnalysisCompareResultItem::id) 
+			.containsExactlyInAnyOrder(hasNewDescription, hasNewRelationship, hasNewAxiom, hasNewLanguageMember);
+
+		assertThat(compareResult.getNewComponents()).isZero();
+		assertThat(compareResult.getChangedComponents()).isEqualTo(4);
+		assertThat(compareResult.getDeletedComponents()).isZero();
+	}
```
nagyo, line 328: Would be nice to cover those scenarios where existing components change, e.g. description term, relationship status, OWL axiom?

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/validation/issue/ValidationIssueDeleteRequest.java
```diff
@@ -66,15 +70,23 @@ public Boolean execute(ServiceProvider context) {
...
 				.collect(Collectors.toSet());
+			
 			query.filter(Expressions.matchAny(ValidationIssue.Fields.RULE_ID, rulesToDelete));
 		}
 		
+		if (!CompareUtils.isEmpty(resultIds)) {
+			query.filter(Expressions.matchAny(ValidationIssue.Fields.RESULT_ID, resultIds));
+		} else {
+			// Use "shared" result ID as the default value if not specified
+			query.filter(Expressions.exactMatch(ValidationIssue.Fields.RESULT_ID, ValidationRequests.SHARED_VALIDATION_RESULT_ID));
```
nagyo, line 81: Isn't there a concern here, that if someone does not specify resourceUris then all shared validation issues will be deleted?
Or in other words, is it somehow enforced to not be able to create a delete request that has an effect over all issues, regardless of the resource identifiers?

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/request/resource/ResourceContentCompareRequestBuilder.java
```diff
@@ -0,0 +1,71 @@
...
+		return getSelf();
+	}
+
+	public ResourceContentCompareRequestBuilder setIncludeChanges(final boolean includeChanges) {
+		this.includeChanges = includeChanges;
+		return getSelf();
+	}
+
+	public ResourceContentCompareRequestBuilder setTermType(final String termType) {
+		this.termType = StringUtils.isEmpty(termType) ? "FSN" : termType;
```
nagyo, line 63: The default value allocation is redundant here, it is already set in the field declaration

nagyo (approved):
LGTM!

nagyo:
Merging PR with an expected follow-up review from @cmark

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/validation/issue/ValidationIssueDeleteRequest.java
apeteri: This just maintains behavioral compatibility with earlier releases &ndash; it was possible to delete all ("shared") validation issues before the introduction of "private" validation runs. If we add access control to the request and use the list of resource URIs for which a permission is required, an empty list would require `edit:*` ie. admin-like capabilities for removing all issues.

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/request/resource/ResourceContentCompareRequestBuilder.java
apeteri: I was thinking about this one when reviewing one of @cmark's pull requests 😄 Setters and field defaults need to cater for at least three scenarios:

1. The setter is not invoked at all when building the request
2. Developer sets value to `null` or an empty string by mistake
3. We receive a parameter object from somewhere (eg. JSON input from a request) and systematically apply its properties to the builder

For the first one, an initialized field with the default value should be fine; the second one however should result in an error, but it seems to be indistinguishable from the third case. But this means you kind of need both pieces to be present.

In some cases we repeat this safe setter-pattern in the setter of the request, occasionally also using an initialized field with the expected default value.

nagyo: I had a hunch there was something like that behind it. Totally reasonable 👍🏻

snomed/com.b2international.snowowl.snomed.core.rest.tests/src/com/b2international/snowowl/snomed/core/request/SnomedDependencyCompareTest.java
apeteri: Added in https://github.com/b2ihealthcare/snow-owl/commit/6418af230cd4e3ae580d0065c0cb68fbeaf14490.

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/request/resource/ResourceContentCompareRequestBuilder.java
cmark: This is use case dependent. A default value should be set for optional parameters. For parameters that always require a value even if they are optional, use the pattern above.

cmark (commented):
Awesome work @apeteri!

PR #1199, opened 2023-09-01
SO-5872 Resource has upgrade

AAAlinaaa:


core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/request/version/VersionCreateRequest.java
```diff
@@ -195,10 +197,35 @@ public Boolean execute(RepositoryContext context) {
...
+						.collect(Collectors.toList());
+				dependentResourceIds.addAll(resourcesInCollection);
+			}
+			
 			// create a version for the resource
 			return new BranchSnapshotContentRequest<>(Branch.MAIN_PATH,
 				new ResourceRepositoryCommitRequestBuilder()
 				.setBody(tx -> {
+					dependentResourceIds.forEach(id -> {
+						ResourceDocument oldDocument = tx.lookup(id, ResourceDocument.class);
```
cmark: Would be better to lookup these resources in bulk.

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/request/version/VersionCreateRequest.java
```diff
@@ -195,10 +197,35 @@ public Boolean execute(RepositoryContext context) {
 				monitor.worked(1);
 //			});
 			
+			//Find resources that depend on this resource
+			List<String> dependentResourceIds = ResourceRequests.prepareSearch()
```
cmark, line 201: If you only need the ID then please use field selection.

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/request/version/VersionCreateRequest.java
```diff
@@ -195,10 +197,35 @@ public Boolean execute(RepositoryContext context) {
...
+				.setLimit(10_000)
+				.filterByDependency(String.format("(uri:%s OR uri:%s/* OR uri:%s\\?*) AND scope:domain) ", resource, resource, resource))
+				.stream(context)
+				.flatMap(Resources::stream)
+				.map(Resource::getId)
+				.collect(Collectors.toList());
+			
+			if (resourceToVersion instanceof TerminologyResourceCollection) {
```
cmark: This is incorrect, when versioning a collection the system will version all child resources. This will be handled in a different issue, please remove this block.

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/request/version/VersionCreateRequest.java
```diff
@@ -195,10 +197,35 @@ public Boolean execute(RepositoryContext context) {
 				monitor.worked(1);
 //			});
 			
+			//Find resources that depend on this resource
+			List<String> dependentResourceIds = ResourceRequests.prepareSearch()
+				.setLimit(10_000)
+				.filterByDependency(String.format("(uri:%s OR uri:%s/* OR uri:%s\\?*) AND scope:domain) ", resource, resource, resource))
```
cmark: Restricting to just the domain scope is an invalid assumption. This code should work regardless of the scope value of the dependency entries.

cmark (changes requested):

PR #1205, opened 2023-09-11
SO-5904: ECL search disable synonym token filter

cmark:
This PR fixes a minor lexical search issue in ECL filtering related to using a search analyzer that includes synonyms of the search string. While this is okay for user interfaces, during ECL evaluation this should be disabled by default when using the `match` lexical search type.

https://snowowl.atlassian.net/browse/SO-5904

### DIFF
--- a/core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/request/search/MatchTermFilter.java
+++ b/core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/request/search/MatchTermFilter.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2022 B2i Healthcare Pte Ltd, http://b2i.sg
+ * Copyright 2022-2023 B2i Healthcare Pte Ltd, http://b2i.sg
  * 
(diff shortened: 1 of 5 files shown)


core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/request/search/MatchTermFilter.java
```diff
@@ -73,6 +75,10 @@ public boolean isCaseSensitive() {
 		return caseSensitive;
 	}
 	
+	public Boolean isSynonyms() {
```
AAAlinaaa, line 78: Could this be another verb, like use or needs?

AAAlinaaa (approved):

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/request/search/MatchTermFilter.java
cmark: Hmm, let's keep it in case we would like to serialize it.

apeteri (approved):
🥇

PR #1206, opened 2023-09-11
SO-5905: ECL wild lexical search type should be case insensitive

cmark:
https://snowowl.atlassian.net/browse/SO-5905


apeteri (approved):
👍

AAAlinaaa (approved):

PR #1210, opened 2023-09-14
SO-5907: prevent mapping updates if the addition is the default object type

nagyo:
https://snowowl.atlassian.net/browse/SO-5907

### DIFF
--- a/commons/com.b2international.index/src/com/b2international/index/es/admin/EsIndexAdmin.java
+++ b/commons/com.b2international.index/src/com/b2international/index/es/admin/EsIndexAdmin.java
@@ -224,17 +224,30 @@ public final class EsIndexAdmin implements IndexAdmin {
 					SortedSet<String> incompatibleChanges = Sets.newTreeSet();
 					final JsonDiff schemaChanges = JsonDiff.diff(currentTypeMapping, newTypeMapping);
 					schemaChanges.forEach(change -> {
+						
 						if (change.isAdd()) {
+							
+							// XXX object type is the default type, so if the current mapping does not contain this node, we shouldn't trigger an update
+							if (change.getFieldPath().endsWith("type") && "object".equals(change.serializeValue())) {
+								return;
+							}
+							
 							compatibleChanges.add(change.getFieldPath());
+							
 						} else if (change.isMove() || change.isReplace()) {
 							incompatibleChanges.add(change.getFieldPath());
 						} else if (change.isRemove()) {
+							
 							// XXX while remove is bad it is hard to detect true incompatibility where we try to support dynamic fields (like Maps)
 							// raise the incompatibility warning only when the root field is being reported, not a nested property under the root property
-							if (!change.getFieldPath().contains("/properties")) {
-								incompatibleChanges.add(change.getFieldPath());
+							if (change.getFieldPath().contains("/properties")) {
+								return;
 							}
+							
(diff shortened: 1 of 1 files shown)


commons/com.b2international.index/src/com/b2international/index/es/admin/EsIndexAdmin.java
```diff
@@ -224,17 +224,30 @@ public void create() {
...
 						} else if (change.isMove() || change.isReplace()) {
 							incompatibleChanges.add(change.getFieldPath());
 						} else if (change.isRemove()) {
+							
 							// XXX while remove is bad it is hard to detect true incompatibility where we try to support dynamic fields (like Maps)
 							// raise the incompatibility warning only when the root field is being reported, not a nested property under the root property
-							if (!change.getFieldPath().contains("/properties")) {
-								incompatibleChanges.add(change.getFieldPath());
+							if (change.getFieldPath().contains("/properties")) {
+								return;
```
cmark, line 244: Not sure about this part, the only this PR meant to change is the handling of the default type `object`. I would revert this back to the original state as changing the mapping of a nested object to something else is currently not possible.

cmark (changes requested):

commons/com.b2international.index/src/com/b2international/index/es/admin/EsIndexAdmin.java
apeteri: The inverted form doesn't rely on a sharp reader to spot the leading `!`:

```java
if (change.getFieldPath().contains("/properties")) {
  return;
}
							
incompatibleChanges.add(change.getFieldPath());
```

commons/com.b2international.index/src/com/b2international/index/es/admin/EsIndexAdmin.java
```diff
@@ -224,17 +224,30 @@ public void create() {
 					SortedSet<String> incompatibleChanges = Sets.newTreeSet();
 					final JsonDiff schemaChanges = JsonDiff.diff(currentTypeMapping, newTypeMapping);
 					schemaChanges.forEach(change -> {
+						
 						if (change.isAdd()) {
+							
+							// XXX object type is the default type, so if the current mapping does not contain this node, we shouldn't trigger an update
+							if (change.getFieldPath().endsWith("type") && "object".equals(change.serializeValue())) {
```
apeteri: If this is always part of a path with more than one segment, the separator should be included to avoid partial word matches like eg. `classType`:

```suggestion
							if (change.getFieldPath().endsWith("/type") && "object".equals(change.serializeValue())) {
```

apeteri (changes requested):

commons/com.b2international.index/src/com/b2international/index/es/admin/EsIndexAdmin.java
nagyo: 👍🏻

commons/com.b2international.index/src/com/b2international/index/es/admin/EsIndexAdmin.java
cmark: True. :blobgo:

cmark (approved):
🥓

apeteri (approved):
🍏

PR #1211, opened 2023-09-14
SO-5877: core API changes necessary for the new upgrade flow

cmark:


core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/console/CodeSystemsCommand.java
```diff
@@ -79,10 +79,10 @@ private String getCodeSystemInfo(CodeSystem codeSystem) {
...
-		final String availableUpgradesInfo = getAvailableUpgradesInfo(codeSystem);
-		if (availableUpgradesInfo != null) {
-			infos.add(availableUpgradesInfo);
-		}
+//		final String availableUpgradesInfo = getAvailableUpgradesInfo(codeSystem);
+//		if (availableUpgradesInfo != null) {
+//			infos.add(availableUpgradesInfo);
+//		}
 
 		if (codeSystem.getUpgradeOf() != null) {
```
apeteri, line 87: Does this and the check below still work?

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/request/version/VersionCreateRequest.java
```diff
@@ -114,24 +116,13 @@ public Boolean execute(RepositoryContext context) {
 		
 		if (!resourcesById.containsKey(resource)) {
 			context.log().warn("Resource cannot be found during versioning: " + resourcesById + ", uri: " + resource);
-			throw new NotFoundException("Resource", resource.getResourceId());
+			throw new BadRequestException("Resource", resource.getResourceId());
```
apeteri: `BadRequestException` will require a formatted string instead of a resource type and ID combination.

apeteri (changes requested):
Two small things need to be addressed, looks good otherwise!

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/request/version/VersionCreateRequest.java
nagyo: Shouldn't we adapt the message here to something more meaningful?

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/request/version/VersionCreateRequest.java
```diff
@@ -114,24 +116,13 @@ public Boolean execute(RepositoryContext context) {
...
-		// TODO resurrect or eliminate tooling dependencies
-		final List<TerminologyResource> resourcesToVersion = List.of(resourcesById.get(resource));
-//		final List<CodeSystem> resourcesToVersion = codeSystem.getDependenciesAndSelf()
-//				.stream()
-//				.map(resourcesById::get)
-//				.collect(Collectors.toList());
+		final Collection<TerminologyResource> resourcesToVersion = resourcesById.values();
 		
 		resourcesToVersion.stream()
 			.filter(cs -> cs.getUpgradeOf() != null)
```
nagyo: Is this check still required? Or shouldn't we do the same check using the new upgrade resource id?

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/request/version/VersionCreateRequest.java
```diff
@@ -236,15 +231,43 @@ private String buildVersionUrl(RepositoryContext context, TerminologyResource re
...
+		TerminologyResource terminologyResource = resourcesToVersion.get(resource);
+		if (terminologyResource instanceof TerminologyResourceCollection resourceCollection) {
+			final var registry = context.service(TerminologyResourceCollectionToolingSupport.Registry.class);
+			final Set<String> childResourceTypes = registry.getAllByToolingId(resourceCollection.getToolingId())
+				.stream()
+				.flatMap(toolingSupport -> toolingSupport.getSupportedChildResourceTypes().stream())
+				.collect(Collectors.toSet());
+			
+			ResourceRequests.prepareSearch()
+				.filterByBundleAncestorId(resourceCollection.getId())
```
nagyo: This method still makes me nervous a bit. Can we have a filterByResourceCollection helper method that internally does the very same thing, but at least the naming is more straightforward?

nagyo (approved):
LGTM, with some minor notes 🍏

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/console/CodeSystemsCommand.java
cmark: Yes, upgradeOf is available as a dependency. This command however is something we should completely remove from the system and introduce it somewhere else, like an external `snow-owl-cli`.

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/request/version/VersionCreateRequest.java
cmark: Yes, you are right. Adapting...

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/request/version/VersionCreateRequest.java
cmark: Not required anymore, removing it.

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/request/version/VersionCreateRequest.java
cmark: I was thinking about this as well yesterday. Adding a new method for clarity.

apeteri (approved):
🍃

PR #1213, opened 2023-09-19
SO-5908: validation issue details field dynamic mapping

cmark:


commons/com.b2international.index/src/com/b2international/index/query/Expressions.java
```diff
@@ -315,6 +298,56 @@ public static QueryStringExpression queryString(String query) {
...
 		return new QueryStringExpression(query, defaultField);
 	}
-	
+
+	public static Expression matchDynamic(String field, Collection<String> dynamicFieldFilters) {
+		ExpressionBuilder bool = bool();
+		for (String dynamicFieldFilter : dynamicFieldFilters) {
+			if (dynamicFieldFilter.contains(DYNAMIC_VALUE_DELIMITER)) {
+				final String propertyName = dynamicFieldFilter.split(DYNAMIC_VALUE_DELIMITER)[0];
+				final String propertyValue = dynamicFieldFilter.substring(propertyName.length() + 1, dynamicFieldFilter.length());
```
apeteri: Minor nitpick: since you already need the position of the separator, an `indexOf + substring + substring` might be slightly faster than a `contains + split[0] + substring` series of operations.

commons/com.b2international.index/src/com/b2international/index/query/Expressions.java
```diff
@@ -315,6 +298,56 @@ public static QueryStringExpression queryString(String query) {
...
+
+				// check special syntax first
+				if (propertyValue.startsWith(IN_CLAUSE)) {
+					// multi-valued -> terms query
+					bool.filter(Expressions.matchAny(fieldToMatch, COMMA_SPLITTER.splitToList(propertyValue.substring(IN_CLAUSE.length()))));
+				} else if (propertyValue.startsWith(RANGE_CLAUSE)) {
+					// range -> range query
+					final List<String> rangeValues = RANGE_SPLITTER.splitToList(propertyValue.substring(RANGE_CLAUSE.length()));
+					if (rangeValues.size() > 2) {
+						throw new BadRequestException("Multiple range expressions (<min>...<max>) are found in property value match. Only a single <min>...<max> expression is allowed.", propertyValue);
```
apeteri: Use the same amount of dots for the range in the error message as there needs to be in a valid expression:

```suggestion
						throw new BadRequestException("Multiple range expressions (<min>..<max>) are found in property value match. Only a single <min>..<max> expression is allowed.", propertyValue);
```

commons/com.b2international.index/src/com/b2international/index/query/Expressions.java
```diff
@@ -315,6 +298,56 @@ public static QueryStringExpression queryString(String query) {
...
+					} else if (rangeValues.size() <= 1) {
+						throw new BadRequestException("At least one range expression (<min>...<max>) is required to match a value range.", propertyValue);
```
apeteri: Same as above.

apeteri (approved):
LGTM!

nagyo (approved):
LGTM!

PR #1214, opened 2023-09-20
Add XML media type support to FHIR REST controllers

apeteri:


apeteri:
Closing in favor of #1227.

PR #1224, opened 2023-09-29
Remove corresponding issues on branch delete

AAAlinaaa:


core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/branch/ValidationCleanupService.java
```diff
@@ -0,0 +1,75 @@
...
+	
+	public void removeStaleIssues(ServiceProvider context, List<String> resourceURIs) {
+		removeStaleIssues(context, resourceURIs, Collections.emptyList());
+	}
+	
+	public void removeStaleIssues(ServiceProvider context, String resourceURI, List<String> previousAnalysisRunIds) {
+		removeStaleIssues(context, List.of(resourceURI), previousAnalysisRunIds);
+	}
+	
+	public void removeStaleIssues(ServiceProvider context, List<String> resourceURIs, List<String> previousAnalysisRunIds) {
```
cmark:
```suggestion
	public void removeStaleIssues(ServiceProvider context, List<String> resourceURIs, List<String> resultIds) {
```

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/branch/ValidationCleanupService.java
```diff
@@ -0,0 +1,75 @@
...
+	
+	public void removeStaleIssues(ServiceProvider context, String resourceURI) {
+		removeStaleIssues(context, List.of(resourceURI), Collections.emptyList());
+	}
+	
+	public void removeStaleIssues(ServiceProvider context, List<String> resourceURIs) {
+		removeStaleIssues(context, resourceURIs, Collections.emptyList());
+	}
+	
+	public void removeStaleIssues(ServiceProvider context, String resourceURI, List<String> previousAnalysisRunIds) {
```
cmark:
```suggestion
	public void removeStaleIssues(ServiceProvider context, String resourceURI, List<String> resultIds) {
```

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/branch/ValidationCleanupService.java
```diff
@@ -0,0 +1,75 @@
...
+		removeStaleIssues(context, List.of(resourceURI), previousAnalysisRunIds);
+	}
+	
+	public void removeStaleIssues(ServiceProvider context, List<String> resourceURIs, List<String> previousAnalysisRunIds) {
+		AsyncRequest<Boolean> request;
+		
+		if (!previousAnalysisRunIds.isEmpty()) {
```
cmark:
```suggestion
		if (!CompareUtils.isEmpty(resultIds)) {
```

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/branch/ValidationCleanupService.java
```diff
@@ -0,0 +1,75 @@
...
+			request = ValidationRequests.issues()
+					.prepareDelete()
+					.setResultIds(previousAnalysisRunIds)
+					.buildAsync();
+		} else {
+			request = ValidationRequests.issues()
+				.prepareDelete()
+				.setCodeSystemURIs(resourceURIs)
```
cmark: What happens if the `resourceURIs` is null or empty?

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/branch/ValidationCleanupService.java
```diff
@@ -0,0 +1,75 @@
...
+				.buildAsync();
+		}
+		
+		final String description = String.format("Remove validation issues on stale/removed branch(es) %s", Joiner.on(", ").join(resourceURIs));
```
cmark:
```suggestion
		final String description = String.format("Remove validation issues on stale/removed branch(es) of '%s'", Joiner.on(", ").join(resourceURIs));
```

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/branch/ValidationCleanupService.java
```diff
@@ -0,0 +1,75 @@
...
+		@Override
+		public boolean contains(ISchedulingRule rule) {
+			return rule == this;
+		}
+	};
+
+	public ValidationCleanupService() {
+	}
+	
+	public void removeStaleIssues(ServiceProvider context, String resourceURI) {
```
cmark: Rename all these methods to `scheduleStaleIssueRemoval` since it is not actually removing anything from the index but schedules a job that removes it.

cmark (changes requested):

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/branch/ValidationCleanupService.java
```diff
@@ -0,0 +1,75 @@
...
+				.setCodeSystemURIs(resourceURIs)
+				.buildAsync();
+		}
+		
+		final String description = String.format("Remove validation issues on stale/removed branch(es) %s", Joiner.on(", ").join(resourceURIs));
+		
+		JobRequests.prepareSchedule()
+			.setRequest(request)
+			.setDescription(description)
+			.setUser(context.service(User.class).getUserId())
```
apeteri: This is the default behavior in `ScheduleJobRequest`, you can leave this property unset if you don't want to override the requester: https://github.com/b2ihealthcare/snow-owl/blob/6b2f0c92b5dbe9563a337ae4869a795cd10f2b01/core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/jobs/ScheduleJobRequest.java#L125

apeteri (approved):
👍 with a suggestion below.

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/branch/ValidationCleanupService.java
AAAlinaaa: Nothing good, we shouldn't delete without resource uri filters.

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/branch/BranchDeleteRequest.java
```diff
@@ -15,25 +15,44 @@
...
 	@Override
 	public Boolean execute(RepositoryContext context) {
 		try {
+			//Remove branch
 			context.service(BaseRevisionBranching.class).delete(getBranchPath());
+			
+			try {
+				//Schedule a job to remove issues corresponding to this branch
+				TerminologyResource resource = context.service(PathTerminologyResourceResolver.class).resolve(context, context.info().id(), getBranchPath());
+				String resourceURI = resource.getResourceURI(getBranchPath()).toString();
```
cmark, line 50: It looks like we have to treat the new tilde path variants (`XZX~branch`) as well here. Could you please add support for that?

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/branch/ValidationCleanupService.java
```diff
@@ -0,0 +1,79 @@
+package com.b2international.snowowl.core.branch;
```
cmark, line 17: Missing license and since tag.

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/branch/ValidationCleanupService.java
cmark: Yeah, but that's a flaw in the API. If the caller should only use one input parameter, then we should split this method into two separate implementations instead of merging them into one with if else blocks.

cmark (changes requested):

PR #1228, opened 2023-10-10
SO-5918: special tilde resource character support

cmark:
This PR adds support for `resourceId~secondaryId` URI values where the secondary ID can be either a special derivative resource ID or a direct branch ID. The system will always prefer fetching a resource with a matching ID first then and only then falls back to direct child branch access.


apeteri (approved):
👍

nagyo (approved):
LGTM! ✝️

PR #1229, opened 2023-10-11
Query optimizer fixes

apeteri:


snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/core/request/SnomedHierarchyStats.java
```diff
@@ -484,4 +508,37 @@ public LongKeyFloatMap computeAncestorScores(
...
+			
+			final String query = expression.getQuery();
+			if (query.startsWith("<< ")) {
+				conceptId = query.substring(3);
+				includeSelf = true;
+			} else if (query.startsWith("< ")) {
+				conceptId = query.substring(2);
+				includeSelf = false;
+			} else {
+				throw new IllegalArgumentException("Unsupported query '" + query + "'.");
```
cmark, line 528: What happens when this fails? Will this kill the optimization request or just prevent a certain logic from running further and the optimizer will try a different strategy?

cmark (changes requested):

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/core/request/SnomedHierarchyStats.java
apeteri: I'd consider this a legitimate bug since we feed back the clauses we just built based on hierarchical information and statistics. There shouldn't be any other type of ECL expressions in this set.

cmark: As it turned out this is being used by the algorithm itself and the input ECL clauses are always well-crafted, this refers to a programmer error and not to a user error. Resolving.

cmark (approved):

PR #1239, opened 2023-11-14
SO-5919: Return access token as part of User object on login

AAAlinaaa:


cmark (approved):
🥓

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/client/TransportClient.java
```diff
@@ -226,17 +226,17 @@ public User connect(final String username, final String password) throws Snowowl
...
 				.buildAsync()
 				.execute(bus)
 				.getSync(3L, TimeUnit.SECONDS);
 			
 			// if successfully logged in replace the event bus with an authorized one
-			env.services().registerService(IEventBus.class, new AuthorizedEventBus(bus, ImmutableMap.of("Authorization", token.getToken())));
+			env.services().registerService(IEventBus.class, new AuthorizedEventBus(bus, ImmutableMap.of("Authorization", user.getAccessToken().getToken())));
 			
-			return IdentityProvider.authJWT(token.getToken());
+			return IdentityProvider.authJWT(user.getAccessToken().getToken());
```
apeteri: I thought this was the primary point of the PR:

```suggestion
			return user;
```

`IdentityProvider.authJWT` uses timestamp verification and we don't want to rely on the precision of the client's timekeeping.

apeteri (changes requested):

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/client/TransportClient.java
cmark: It should be yes, @AAAlinaaa pls fix.

cmark (approved):
Yes, it looks good now.

apeteri (approved):
🍃

PR #1241, opened 2023-11-14
SO-3916: set definition status to primitive regardless of user choice

adamfilep:


cmark:
Please find a corresponding test case and verify that inactivating a concept indeed sets the status to primitve. If there is no such test case, please write one.

apeteri (approved):
👍

cmark (approved):
🥓

PR #1242, opened 2023-11-14
fix(ComponentURI): Fix is unspecified checks

AAAlinaaa:


apeteri (approved):
👍 as discussed &ndash; the check was using the resource ID only in previous versions.

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/uri/ComponentURI.java
```diff
@@ -96,7 +96,7 @@ public String identifier() {
 
 	@JsonIgnore
 	public final boolean isUnspecified() {
-		return CodeSystem.uri(TerminologyRegistry.UNSPECIFIED).equals(resourceUri());
+		return TerminologyRegistry.UNSPECIFIED.equals(resourceUri().getResourceId());
```
cmark, line 99: Hmm, this means that the `UNSPECIFIED` value becomes a special resource ID value, we might need to add a validation rule to the resource create API.

cmark (changes requested):

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/uri/ComponentURI.java
apeteri: Historically (in 7.x) we checked it this way, then it became wrapped into `CodeSystem.uri(...)` during the rewrite.

cmark: Yes, I know, the validation logic should be available since the introduction of this special value. I'll approve the PR and merge it. @apeteri please create a spin-off ticket that describes the issue of having this special ID as resource ID, thanks!

cmark (approved):
🥓

PR #1243, opened 2023-11-17
SO-5363: Do not update Moved from association members during concept...

AAAlinaaa:
... reactivation, do not include Moved from replacements for inactive references in quick search widgets, preserve type check in RF2 import


snomed/com.b2international.snowowl.snomed.common/src/com/b2international/snowowl/snomed/common/SnomedConstants.java
```diff
@@ -257,7 +257,6 @@ private Concepts() { }
 		public static final Set<String> HISTORICAL_ASSOCIATION_REFSETS = ImmutableSet.of(
 				REFSET_HISTORICAL_ASSOCIATION, 
 				REFSET_ALTERNATIVE_ASSOCIATION,
-				REFSET_MOVED_FROM_ASSOCIATION,
```
cmark, line 260: I would like to see this set removed and new explicit sets introduced at the various places where we actually need historical association refsets. Refactor please.

cmark (changes requested):

apeteri (approved):
LGTM!

cmark (approved):
🥓

PR #1246, opened 2023-11-22
feat: make validation jobs cancellable

adamfilep:


core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/validation/ValidateRequest.java
```diff
@@ -102,13 +107,17 @@ private ValidationResult doValidate(BranchContext context, Writer index) throws
 		final List<Promise<Object>> validationPromises = Lists.newArrayList();
 		// evaluate selected rules
 		for (ValidationRule rule : rules) {
+			RemoteJob remoteJob = context.service(RemoteJob.class);
```
cmark: This might result in a missing service exception due to the fact that Request implementations are not always wrapped and run in jobs. Please remove this line.

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/validation/ValidateRequest.java
```diff
@@ -102,13 +107,17 @@ private ValidationResult doValidate(BranchContext context, Writer index) throws
...
 			checkArgument(rule.getCheckType() != null, "CheckType is missing for rule " + rule.getId());
 			final ValidationRuleEvaluator evaluator = ValidationRuleEvaluator.Registry.get(rule.getType());
 			if (evaluator != null) {
 				validationPromises.add(pool.submit(rule.getCheckType(), () -> {
 					Stopwatch w = Stopwatch.createStarted();
 					
 					try {
+						if (monitor.isCanceled()) {
+							throw new OperationCanceledException(remoteJob.getName() + " has been canceled");
```
cmark: The message in this case is irrelevant, if there is a no-arg ctor please use that, if a message has to be provided then just indicate `Canceled`.

cmark: The bigger problem here is that the exception thrown by this statement is caught by line 126 and the validation rule execution will fail instead of being reported as canceled.

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/validation/ValidateRequest.java
```diff
@@ -102,13 +107,17 @@ private ValidationResult doValidate(BranchContext context, Writer index) throws
 		final List<Promise<Object>> validationPromises = Lists.newArrayList();
 		// evaluate selected rules
 		for (ValidationRule rule : rules) {
+			RemoteJob remoteJob = context.service(RemoteJob.class);
 			checkArgument(rule.getCheckType() != null, "CheckType is missing for rule " + rule.getId());
 			final ValidationRuleEvaluator evaluator = ValidationRuleEvaluator.Registry.get(rule.getType());
 			if (evaluator != null) {
 				validationPromises.add(pool.submit(rule.getCheckType(), () -> {
```
cmark, line 116: It would be great to reduce pressure on the validation pool by not submitting more and more rules to execution once we know that the operation was canceled by the user. Please add a check before this line.

cmark (changes requested):
@AAAlinaaa pls help @adamfilep resolve these fixes so that we can merge the changes as soon as possible. Thanks!

cmark (approved):
🥓

apeteri (approved):
👍

cmark (approved):
👍

AAAlinaaa:
Also needs the following commit: https://github.com/b2ihealthcare/snow-owl/commit/08fdef0611ad7d063c90a470b147bb6a6ee2c69f

PR #1249, opened 2023-12-05
SO-4122: index schema migration support

cmark:
This PR adds a very basic, index schema migrator functionality to aid existing indices to migrate to a newer schema version.


commons/com.b2international.index/src/com/b2international/index/es/EsDocumentWriter.java
```diff
@@ -280,15 +289,15 @@ private void dumpOps() throws IOException {
...
 		}
 		System.err.println("Bulk updates: ");
 		for (BulkUpdate<?> update : bulkUpdateOperations) {
-			System.err.format("\t%s -> %s, %s, %s\n", admin.mappings().getMapping(update.getType()).typeAsString(), update.getFilter(), update.getScript(), update.getParams());
+			System.err.format("\t%s -> %s, %s, %s\n", admin.getIndexMapping().getMapping(update.getType()).typeAsString(), update.getFilter(), update.getScript(), update.getParams());
 		}
 		System.err.println("Bulk deletes: ");
 		for (BulkDelete<?> delete : bulkDeleteOperations) {
-			System.err.format("\t%s -> %s\n", admin.mappings().getMapping(delete.getType()).typeAsString(), delete.getFilter());
+			System.err.format("\t%s -> %s\n", admin.getIndexMapping().getMapping(delete.getType()).typeAsString(), delete.getFilter());
```
apeteri: These last three reporters could also use the `indexMapping` field instead of going through `admin`.

commons/com.b2international.index/src/com/b2international/index/es/admin/EsIndexAdmin.java
```diff
@@ -185,9 +197,13 @@ public void create() {
 		Set<DocumentMapping> mappingsToRefresh = Sets.newHashSet();
 		
 		// create number of indexes based on number of types
-		for (DocumentMapping mapping : mappings.getMappings()) {
-			final String index = getTypeIndex(mapping);
+		for (DocumentMapping mapping : this.indexMapping.getMappings().getMappings()) {
```
apeteri: These two method calls with the same name look very intense 😵

commons/com.b2international.index/src/com/b2international/index/es/admin/EsIndexAdmin.java
```diff
@@ -217,93 +235,218 @@ public void create() {
...
+						compatibleChanges.add(change.getFieldPath());
+					} else if (change.isRemove()) {
+						
+						// XXX while remove is bad it is hard to detect true incompatibility where we try to support dynamic fields (like Maps)
+						// throw the incompatibility error only when a root field is being reported, not a nested property under the root property
+						if (change.getFieldPath().contains("/properties")) {
+							return;
+						}
+						
+						compatibleChanges.add(change.getFieldPath());
```
apeteri: None of the changes are incompatible now (I can't find any reference to `incompatibleChanges.add(...)`). Shouldn't this be one of those?

commons/com.b2international.index/src/com/b2international/index/es/admin/EsIndexAdmin.java
```diff
@@ -217,93 +235,218 @@ public void create() {
...
+						// XXX object type is the default type, so if the current mapping does not contain this node, we shouldn't trigger an update
+						if (change.getFieldPath().endsWith("/type") && "object".equals(change.serializeValue())) {
+							return;
+						}
+						
+						compatibleChanges.add(change.getFieldPath());
 						
-						if (change.isAdd()) {
+					} else if (change.isMove() || change.isReplace()) {
+						compatibleChanges.add(change.getFieldPath());
```
apeteri: This used to be an incompatible change as well.

commons/com.b2international.index/src/com/b2international/index/es/admin/EsIndexAdmin.java
```diff
@@ -217,93 +235,218 @@ public void create() {
...
+						
+						switch (schema.strategy()) {
+						case NO_REINDEX:
+							// just apply the schema changes by modifying the mapping and we are good to go
+							putIndexMapping(index, typeMapping);
+							break;
+						case REINDEX_INPLACE:
+							// apply the schema changes first, there should be only compatible changes here
+							putIndexMapping(index, typeMapping);
+							// then reindex all documents in place to pick up mapping changes, mostly for
```
apeteri: This comment got cut off at the end. 🙂

commons/com.b2international.index/src/com/b2international/index/es/admin/EsIndexAdmin.java
```diff
@@ -217,93 +235,218 @@ public void create() {
...
 		}
+		
 		log.info("'{}' indexes are ready.", name);
 	}
 
-	private boolean hasFieldAliasChange(SortedSet<String> compatibleChanges) {
-		return compatibleChanges.stream()
-				.map(FIELD_ALIAS_CHANGE_PROPERTY_PATTERN::matcher)
-				.anyMatch(Matcher::matches);
+	private static Iterable<Hits<JsonNode>> readAllRaw(EsDocumentSearcher previousIndexSearcher, DocumentMapping mapping, int batchSize) {
```
apeteri: Keeping the `Stream` abstraction would be cheaper, although the difference is very small.

These batches are only processed in a single `for-each` statement above that could be replaced with a call to `.forEachOrdered(hits ->{ ... })` (unless using a lambda function would complicate access to variables).

commons/com.b2international.index/src/com/b2international/index/es/admin/EsIndexAdmin.java
```diff
@@ -748,13 +893,17 @@ public ReindexResult reindex(String sourceIndex, String destinationIndex, Remote
...
-			.remoteAddress(String.format("%s://%s:%s", remoteInfo.getScheme(), remoteInfo.getHost(), remoteInfo.getPort()))
+			.remoteAddress(remoteAddress)
 			.refresh(refresh)
 			.retries(retries.get() > 1 ? Long.valueOf(retries.get()) : null) // do not track successful first attempts
 			.build();
 		
 	}
 
+	private String getRemoteAddress(RemoteInfo remoteInfo) {
+		return Optional.ofNullable(remoteInfo).map(info -> String.format("%s://%s:%s", info.getScheme(), info.getHost(), info.getPort())).orElse(null);
```
apeteri: If this String is purely informational, the `orElse` portion could return `"(local)"` or something that indicates that the reindex happened on the same ES instance.

apeteri (changes requested):

commons/com.b2international.index/src/com/b2international/index/es/admin/EsIndexAdmin.java
nagyo: +1 What about renaming the second method to `getDocumentMappings()`? After all, it returns a collection of DocumentMapping objects.

commons/com.b2international.index/src/com/b2international/index/es/admin/EsIndexAdmin.java
nagyo: Agree

nagyo (changes requested):
Overall looks good to me, some minor suggestions in the comments!

commons/com.b2international.index/src/com/b2international/index/es/admin/EsIndexAdmin.java
cmark: 🐛

apeteri (approved):
👍

nagyo (approved):
LGTM!

PR #1255, opened 2024-02-14
Disable tycho consumer poms

nagyo:
See additional details here: https://github.com/eclipse-tycho/tycho/issues/664

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -67,7 +67,6 @@
 				<configuration>
 					<strictBinIncludes>false</strictBinIncludes>
 					<strictVersions>false</strictVersions>
-					<skipPomGeneration>true</skipPomGeneration>
 				</configuration>
 			</plugin>
 			<plugin>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -67,6 +67,7 @@
 				<configuration>
 					<strictBinIncludes>false</strictBinIncludes>
 					<strictVersions>false</strictVersions>
+					<skipPomGeneration>true</skipPomGeneration>
```
apeteri: I'm not sure if this one is needed if the `pluginManagement` section below defines it as well. Can you check what the effective POM looks like without this change?

apeteri (approved):
LGTM with a question below.

pom.xml
nagyo: You were right, the second one is enough 👍🏻

cmark (approved):
🥓

PR #1265, opened 2024-03-12
SO-6056: consolidate versioning timestamps

cmark:
Fixes issues when fetching resource state using the `version.createdAt` timestamp value but the commit was made with a later timestamp value due to getting it from a different provider.

Fixes https://snowowl.atlassian.net/browse/SO-6056


apeteri (approved):
👍

nagyo (approved):
LGTM!

PR #1270, opened 2024-03-27
SO-4992: ETag and Cache-Control header support

cmark:
This PR adds support for `ETag` and `Cache-Control` headers so that HTTP clients, proxies, and intermediaries can cache the responses earlier than caching only in an HTTP browser. 

~Still somewhat blocked by https://github.com/spring-projects/spring-security/issues/12865. We need to figure out a way of preventing Spring from adding the default cache-control headers when we set our own.~

Resolved the above in the meantime by explicitly setting the headers in the response before Spring can write its defaults incorrectly.

Background: https://snowowl.atlassian.net/browse/SO-4992

### DIFF
--- a/core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/rate/RateLimitConfig.java
+++ b/core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/rate/RateLimitConfig.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2023 B2i Healthcare Pte Ltd, http://b2i.sg
+ * Copyright 2023-2024 B2i Healthcare, https://b2ihealthcare.com
  * 
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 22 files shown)


apeteri (approved):
👍

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/rate/RateLimitConfig.java
```diff
@@ -0,0 +1,47 @@
...
+
+import jakarta.validation.constraints.Min;
+
+/**
+ * @since 8.10
+ */
+public class RateLimitConfig {
+
+	@Min(0)
+	private long overdraft = 0L;
```
nagyo: I would love to see a simple one-sentence explanation about the purpose of these values. For future devs I think it would be extremely useful.

nagyo (approved):
RSLGTM!

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/rate/RateLimitConfig.java
cmark: Done

apeteri (approved):
👍

nagyo:
👌🏻

PR #1274, opened 2024-04-03
Allow ObjectNode fields on ES documents

apeteri:
The property `index` on the field-level annotation `Field` controls whether it becomes a source-only or a dynamically mapped object in Elasticsearch's mapping. In the latter case top-level properties of the JSON object become searchable.


cmark (approved):
🥓

AAAlinaaa (approved):

PR #1275, opened 2024-04-03
SO-5901: support description languages in generic concept model

cmark:
This PR adds support for generic descriptions instead of a List of Strings on the generic Concept model.
The primary use for this is going to be FHIR endpoints where additional (even on the fly generated) designations need to be sent back in the response along with the concept's display.
In case of SNOMED CT Snow Owl now follows the latest spec version which requires designations to be returned for each languageCode + language reference set id combination.

### DIFF
--- a/core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/domain/Concept.java
+++ b/core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/domain/Concept.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2020-2023 B2i Healthcare, https://b2ihealthcare.com
+ * Copyright 2020-2024 B2i Healthcare, https://b2ihealthcare.com
  * 
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -15,9 +15,11 @@
  */
 package com.b2international.snowowl.core.domain;
 
+import java.util.Collection;
 import java.util.List;
 import java.util.SortedSet;
 
+import com.b2international.commons.collections.Collections3;
 import com.b2international.snowowl.core.ResourceURI;
 import com.b2international.snowowl.core.uri.ComponentURI;
 import com.fasterxml.jackson.annotation.JsonCreator;
@@ -41,7 +43,7 @@ public final class Concept extends BaseComponent {
 	private Boolean active;
 	private String term;
 	private String iconId;
-	private SortedSet<String> alternativeTerms;
+	private SortedSet<Description> alternativeTerms;
 	
 	private List<String> parentIds;
 	private List<String> ancestorIds;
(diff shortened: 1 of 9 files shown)


apeteri (approved):
👍

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/domain/Concept.java
```diff
@@ -104,8 +106,19 @@ public void setIconId(String iconId) {
...
 	}
 	
-	public void setAlternativeTerms(SortedSet<String> alternativeTerms) {
-		this.alternativeTerms = alternativeTerms;
+	/**
+	 * @param alternativeTerms
+	 * @deprecated deserialization only: if descriptions property is set via the constructor during deserialization then this ignores the incoming value, use {@link #setDescriptions(Collection)} instead
+	 */
+	@JsonProperty
+	public void setAlternativeTerms(Collection<Description> alternativeTerms) {
```
AAAlinaaa, line 114: Are the descriptions all alternative descriptions or do they include the description of the term?

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/core/request/SnomedConceptSearchRequestEvaluator.java
```diff
@@ -133,5 +114,43 @@ public Concepts evaluate(ResourceURI uri, ServiceProvider context, Options searc
 		);
 	}
 	
+	private Concept toConcept(ResourceURI codeSystem, SnomedConcept snomedConcept, String pt, boolean requestedExpand) {
+		final Concept concept = toConcept(codeSystem, snomedConcept, snomedConcept.getIconId(), pt, snomedConcept.getScore());
+		
+		SortedSet<Description> descriptions = generateGenericDescriptions(snomedConcept.getPreferredDescriptions());
```
AAAlinaaa, line 120: Why do we use only preferred descriptions here?

AAAlinaaa (approved):
LGTM

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/core/request/SnomedConceptSearchRequestEvaluator.java
cmark: Simplification, acceptable descriptions would require a smart join which is not that easy to do and most of the time only the preferred descriptions are needed.

AAAlinaaa: It just feels a little against the spirit of alternative terms to me to not use getDescriptions() here. 🙈

cmark: I know we can improve this later but let's keep it this way for now.

PR #1277, opened 2024-04-08
feat(snomed.reasoner): Reinstate equivalent concept merging extension

apeteri:


snomed/com.b2international.snowowl.snomed.reasoner/plugin.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<?eclipse version="3.4"?>
+<plugin>
```
cmark: Would it be possible to use the newer plugin service registration mechanism? ClassPathScanner + service/toolingId lookup.

cmark (changes requested):

snomed/com.b2international.snowowl.snomed.reasoner/plugin.xml
apeteri: I've replaced the extension point with `ClassPathScanner` and a registry in commit cfef3f5.

cmark (approved):
🥓

nagyo (approved):
LGTM!

PR #1281, opened 2024-04-26
SO-6087 Avoid creating new revision in absence of changes

AAAlinaaa:


commons/com.b2international.index/src/com/b2international/index/revision/StagingArea.java
```diff
@@ -733,7 +733,10 @@ public StagingArea stageNew(String key, Object newDocument, boolean commit) {
 		if (stagedObjects.containsKey(objectId)) {
 			StagedObject currentStagedObject = stagedObjects.get(objectId);
 			if (!currentStagedObject.isCommit() && currentStagedObject.getObject() instanceof Revision && newDocument instanceof Revision) {
-				stagedObjects.put(objectId, changed(newDocument, new RevisionDiff((Revision) currentStagedObject.getObject(), (Revision) newDocument), commit));
+				RevisionDiff diff = new RevisionDiff((Revision) currentStagedObject.getObject(), (Revision) newDocument);
+				if (diff.hasChanges()) {
+					stagedObjects.put(objectId, changed(newDocument, diff, commit));
```
cmark, line 751: I think we can create a very minimal test case for this in the index layer module.
1. Use the RevisionData fixture class
2. Start a commit by creating a staging area
3. Add an object to the staging area with commit=false
4. Then add an object to the staging area with commit=true with no changes (assert after this that the staging area did not change)
5. Then add an object to the staging area with commit=true with changes (assert after this that the staging area did change)

cmark (changes requested):

apeteri (approved):
LGTM!

commons/com.b2international.index.tests/src/com/b2international/index/revision/SingleDocumentRevisionIndexTest.java
```diff
@@ -54,6 +54,21 @@ public void updateRevision() throws Exception {
...
+	public void stageNewRevisionMultipleTimes() throws Exception {
+		final RevisionData data = new RevisionData(STORAGE_KEY1, "fieldA", "fieldB");
+		StagingArea staging = index().prepareCommit(MAIN);
+		staging.stageNew(data, false);
+		
+		staging.stageNew(data, true);
+		assertEquals(1, staging.getNumberOfStagedObjects());
+		assertEquals(data, staging.getNewObject(RevisionData.class, STORAGE_KEY1));
+		
+		staging.stageNew(data, false);
```
cmark: The implemented test case is a bit different than the one I described, please take a look again.

cmark (changes requested):

commons/com.b2international.index.tests/src/com/b2international/index/revision/SingleDocumentRevisionIndexTest.java
```diff
@@ -54,6 +54,22 @@ public void updateRevision() throws Exception {
...
+		
+		staging.stageNew(data, true);
+		assertEquals(1, staging.getNumberOfStagedObjects());
+		assertEquals(data, staging.getNewObject(RevisionData.class, STORAGE_KEY1));
+		
+		final RevisionData updatedData = new RevisionData(STORAGE_KEY1, "fieldC", "fieldB");
+		staging.stageNew(updatedData, true);
+		assertEquals(1, staging.getNumberOfStagedObjects());
+		assertEquals(updatedData, staging.getNewObject(RevisionData.class, STORAGE_KEY1));
```
cmark: Are you sure this is the right outcome here? According to the implementation, this should give us a changed object rather than a new object.

cmark (changes requested):

cmark (approved):
🥓

apeteri (commented):
💯

PR #1294, opened 2024-05-14
feat(snowowl.core): Small change required for downstream project

apeteri:
- Make test dependencies of `com.b2international.snowowl.validation.snomed` optional so that they do not need to be included in launch configurations
- Make the RF2 importer for relationships less strict, allowing matching `sourceId` and `destinationId` values for non-IS A relationships
- Add Java system property that allows extending the list of Java packages to inspect during classpath scanning (the property name is `so.component.scan` and it expects a comma-separated list of package names as its value)


nagyo (approved):
LGTM!

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/SnowOwl.java
```diff
@@ -73,6 +73,9 @@ public final class SnowOwl {
 	private static final String DEFAULT_DATA_PATH = "resources"; //$NON-NLS-1$
 	private static final String LOG_CONFIG_FILE = "serviceability.xml"; //$NON-NLS-1$
 	
+	// System property for additional packages to scan
```
cmark, line 76: I would indicate in the property name that this is for component scanning. How about `-Dso.component.scan` or something similar?

cmark (changes requested):

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/SnowOwl.java
apeteri: Updated in 039b1c7bf1271e14ed0897757538e25fb4a31792.

cmark: Please update the PR description section as well.

cmark (approved):

PR #1302, opened 2024-06-26
feat(classification): resurrect cd member inference

cmark:
Older systems with member based concrete domain support (via the concreteDomainSupport configuration setting) should be able to keep their inferred concrete data values in cd members (as opposed to moving them to the new relationship value model which has limited support for data types).

NOTE: Test cases will arrive in a separate PR.

https://snowowl.atlassian.net/browse/SO-6164

### DIFF
--- a/snomed/com.b2international.snowowl.snomed.reasoner/src/com/b2international/snowowl/snomed/reasoner/normalform/NormalFormGenerator.java
+++ b/snomed/com.b2international.snowowl.snomed.reasoner/src/com/b2international/snowowl/snomed/reasoner/normalform/NormalFormGenerator.java
@@ -55,12 +55,7 @@ import com.b2international.snowowl.snomed.reasoner.diff.concretedomain.ConcreteD
 import com.b2international.snowowl.snomed.reasoner.diff.relationship.StatementFragmentOrdering;
 import com.google.common.base.Predicates;
 import com.google.common.base.Stopwatch;
-import com.google.common.collect.FluentIterable;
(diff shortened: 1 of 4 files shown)


snomed/com.b2international.snowowl.snomed.reasoner/src/com/b2international/snowowl/snomed/reasoner/normalform/NormalFormGenerator.java
```diff
@@ -531,11 +533,25 @@ private Iterable<StatementFragment> relationshipsFromUnionGroup(final NormalForm
...
+	
+	private Iterable<ConcreteDomainFragment> membersFromGroup(final NormalFormGroup group) {
+		return FluentIterable
+				.from(group.getUnionGroups())
+				.transformAndConcat(unionGroup -> membersFromUnionGroup(unionGroup, 
+					group.getGroupNumber(), 
+					unionGroup.getUnionGroupNumber()));
+	}
+
+	private Iterable<ConcreteDomainFragment> membersFromUnionGroup(final NormalFormUnionGroup unionGroup, final int groupNumber, final int unionGroupNumber) {
```
apeteri: Since CD members have no field to store a union group number in, we can either leave this off and/or check whether somehow a refset member ended up in a non-zero union group during the collection process.

apeteri (approved):
👍

nagyo (approved):
LGTM!

PR #1310, opened 2024-08-06
SO-5868: Retrieve validation results via REST

adamfilep:


apeteri (approved):
🥇

core/com.b2international.snowowl.core.rest/src/com/b2international/snowowl/core/rest/SnowOwlApiConfig.java
```diff
@@ -181,6 +181,16 @@ public SnowOwlOpenApiWebMvcResource openApiWebMvcResource(
 	@Bean
 	@Scope(scopeName = WebApplicationContext.SCOPE_REQUEST, proxyMode = ScopedProxyMode.TARGET_CLASS)
 	public ObjectMapper objectMapper(@Autowired HttpServletRequest request) {
+		return objectMapperFromRequest(request);
```
cmark, line 185: I think for consistency, this should be a provider as well.

cmark (changes requested):

core/com.b2international.snowowl.core.rest/src/com/b2international/snowowl/core/rest/SnowOwlApiConfig.java
adamfilep: As I see it, we are going to need the two different methods:
This method with ObjectMapper return type is necessary for mappingJackson2HttpMessageConverter()
And the other one, with Provider<ObjectMapper> is necessary for the RepositoryValidationRestService class

apeteri: IIRC we tried three approaches:

* Keep the initialization code in this method and refer to it in the provider lambda
    * &rarr; the `Provider` produced `ObjectMapper` delegates (proxies) that expected a request scope to be present, triggering the exception in the validation REST controller (where it was originally reported)
* Move the initialization code to the lambda and create an instance here using the `Provider`
    * &rarr; the returned instance is an actual `ObjectMapper` but creating it requires a request scope which is not available at startup, triggering the exception right away
* Move the initialization code to a non-&commat;`Bean`-annotated method and refer to it from both annotated methods
    * &rarr; the current state of the PR that worked without issues

cmark (approved):
🥓

PR #1311, opened 2024-08-06
feat(snomed.datastore): Allow clients to selectively update concept members

apeteri:
This is intended for users of the Java API who are not able to present a complete list of members (that includes both members to be updated as well as the ones which need to stay as they are) in `SnomedConceptUpdateRequest`'s reference set members collection.


nagyo (approved):
LGTM!

AAAlinaaa (approved):

cmark (commented):
🥓

PR #1313, opened 2024-08-13
fix(api): upgrade transitive dependencies to fix some security vulnerabilities

cmark:


apeteri (approved):
👍, although I'm wondering why `package-lock.json` ended up with 21k lines of additions and deletions. Is `antd` and `qs` used in everything?

cmark:
> 👍, although I'm wondering why `package-lock.json` ended up with 21k lines of additions and deletions. Is `antd` and `qs` used in everything?

No, those two were related to a small number of changes in the `package-lock.json`. I ran an `npm audit fix` which is required to properly bump locked down versions of transitive dependencies in the lock file that are not explicitly depended on by our module.

nagyo (approved):
LGTM!

PR #1314, opened 2024-08-14
SO-6223: Replace log4j with reload4j

adamfilep:


apeteri (approved):
👍

cmark (approved):
🥓

cmark:
While this adds reload4j as an additional jar, if there is no use for log4j 1.x jars anywhere then it would be great to completely remove it from the target platform during a Tycho build. 
@apeteri @adamfilep could you guys take a look at this as well as part of this PR? Or let me know if this cannot be added.

apeteri:
`xtext.common.types` has a non-negotiable `Import-Package` on log4j 1.x:

image

cmark:
> `xtext.common.types` has a non-negotiable `Import-Package` on log4j 1.x:
> 
> image

But that is an Import-Package, not a Require-Bundle right? So in theory the reload4j bundle could server the necessary packages to it. Am I missing something?

apeteri:
The only thing you might be missing is that reload4j is referenced in the "dependencies" feature by OSGi bundle ID, which is still `org.apache.log4j` for maximum compatibility. Excerpt from `~/.m2/repository/ch/qos/reload4j/reload4j/1.2.25/reload4j-1.2.25.jar!/META-INF/MANIFEST.MF`:

```properties
Manifest-Version: 1.0
Implementation-Title: reload4j
Bundle-Description: Reload4j revives EOLed log4j 1.x
Automatic-Module-Name: ch.qos.reload4j
Bundle-License: http://www.apache.org/licenses/LICENSE-2.0.txt
Bundle-SymbolicName: org.apache.log4j
...
```

cmark:
Ahhh, I see, that was unclear to me, I thought the name of the file is identical with the Bundle-SymbolicName, but this is way better. Now everyone who'd like to depend on log4j can do it, and security scans can also be happy. Yay! Thanks!

apeteri:
It is highly likely that Xtext pulled in the _very same library_ with matching content, but for P2 (or other, unknown) reasons the file keeps using `org.apache.log4j_<version>.jar` as its name. This is the most aggravating part 😄

cmark:
Yeah, that's what we need to change, either by renaming it or doing some other magic.

PR #1320, opened 2024-09-13
SO-6267: Update handling of validation thread pool size

apeteri:
- Change the property name from `numberOfValidationThreads` to `workerPoolSize`
- Allow setting either property for backwards compatibility (setting both results in undefined behavior)
- Use `null` as the default value
- Increase allowed maximum pool size to 99
- Compute a reasonable default value on the server side if the value remains unset in Snow Owl's configuration file

### DIFF
--- a/core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/internal/validation/ValidationPlugin.java
+++ b/core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/internal/validation/ValidationPlugin.java
@@ -53,6 +53,7 @@ import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
 import com.google.common.collect.Sets;
+import com.google.common.primitives.Ints;
 
 /**
  * @since 6.0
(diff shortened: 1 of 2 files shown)


core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/internal/validation/ValidationPlugin.java
```diff
@@ -99,12 +100,17 @@ public void preRun(SnowOwlConfiguration configuration, Environment env) throws E
 			// initialize validation thread pool
 			final ValidationConfiguration validationConfig = configuration.getModuleConfig(ValidationConfiguration.class);
 
-			int numberOfValidationThreads = validationConfig.getNumberOfValidationThreads();
+			Integer workerPoolSize = validationConfig.getWorkerPoolSize();
+			if (workerPoolSize == null) {
+				// Restrict thread pool size to the [4..12] range (both inclusive)
```
cmark, line 105: ❤️

cmark (approved):
🥓

nagyo (approved):
LGTM!

PR #1326, opened 2024-09-25
Term display in concept string

AAAlinaaa:


cmark (approved):
🥓

cmark:
@AAAlinaaa please move the logic of detecting id and term equality to the caller side of your code. This might be an unwanted side-effect for some other callers.

PR #1330, opened 2024-09-30
Filter MRCM validation rule results by module

AAAlinaaa:


nagyo (approved):
LGTM!

cmark (approved):
🥓

apeteri (approved):
👍 

I think `rule_mrcm_constraint` should get the same treatment (filter potential violators by module ID).

The two script's stated counterparts, `rule_stated_mrcm_constraint_type` and `rule_stated_mrcm_constraint` are currently completely commented out with a notice that they need to be looked at once we migrate to the new MRCM representation. But I also see that this validation script is already looking at OWL axioms, so they might no longer need to be restored; in that case they should be removed instead.

AAAlinaaa:
@apeteri True, the member filter should be added to the other mrcm rule as well. Regarding the stated forms of the rules; the name might be misleading, it's not reserved for validating stated relationships; it uses the stated form when evaluating ECL expressions for domain/range constraints. They might be worth keeping.

PR #1332, opened 2024-10-07
SO-6081: support multiple FHIR versions in the REST API

cmark:
This PR adds support for multiple FHIR mime types, supporting R4, R4B, and R5.
This PR also adds a new operation `CapabilityStatement$versions` to list the available versions in Snow Owl.


nagyo (approved):
LGTM so far!

fhir/com.b2international.snowowl.fhir.rest/src/com/b2international/snowowl/fhir/rest/FhirMediaType.java
```diff
@@ -0,0 +1,354 @@
...
+		default: 
+			throw new NotImplementedException("No serializer implementation found for format: " + fhirFormat);
+		}
+	}
+	
+	private void writeResourceJson(ByteArrayOutputStream baos, Resource resource, boolean pretty) throws IOException {
+		switch (fhirVersion) {
+		case _4_0:
+			org.hl7.fhir.r4.model.Resource r4 = VersionConvertorFactory_40_50.convertResource(resource);
+			new org.hl7.fhir.r4.formats.JsonParser().setOutputStyle(pretty ? org.hl7.fhir.r4.formats.IParser.OutputStyle.PRETTY : org.hl7.fhir.r4.formats.IParser.OutputStyle.NORMAL).compose(baos, r4);
```
apeteri, line 239: Do `compose` calls close the output stream? (To be fair I'm not sure which one is needed, leaving it open until the end of the request or closing it when it is certain that no more output will be added.)

apeteri (approved):
👍

fhir/com.b2international.snowowl.fhir.rest/src/com/b2international/snowowl/fhir/rest/FhirMediaType.java
cmark: ByteArrayOutputStreams does not need closing. See javadoc here:
https://docs.oracle.com/javase/8/docs/api/java/io/ByteArrayOutputStream.html

According to the format implementations the out stream is not closed after the compose method call completes. It's the responsibility of the caller, which is good I think.

apeteri: Aah, all right, I didn't see / remember that this does not go into the HTTP response directly (an `OutputStream` could be obtained for that as well) but it will get turned into an actual `byte[]` and that will be sent back in the form of a `ResponseEntity` with headers, status code, etc.

apeteri (approved):
🍃

PR #1335, opened 2024-10-14
SO-6307: Do not report module dependency member conflicts

AAAlinaaa:


nagyo (approved):
LGTM!

cmark (approved):
🥓

apeteri:
Is it OK if the module dependency conflict check also falls under these pre-conditions:

```java
// detect if we are merging content between two CodeSystems, if not skip donation check
// get the two CodeSystems
String extensionBranch = staging.getMergeFromBranchPath();
String donationBranch = staging.getBranchPath();
		
CodeSystem extensionCodeSystem = context.service(RepositoryCodeSystemProvider.class).get(extensionBranch);
CodeSystem donationCodeSystem = context.service(RepositoryCodeSystemProvider.class).get(donationBranch);
		
// donation Code System should be marked as extension CodeSystem to be able to detect donation changes, otherwise skip donation check and report all conflicts
// extensionOf is a required property for Code Systems that would like to participate in content donation
if (extensionCodeSystem.getExtensionOf() == null || !extensionCodeSystem.getExtensionOf().getCodeSystem().equals(donationCodeSystem.getShortName())) {
  return conflicts;
}
```

IOW the code system corresponding to the target branch must have an `extensionOf` dependency pointing to the code system that is in control of the source branch.

AAAlinaaa:
@apeteri Yes, this should be fine.

PR #1340, opened 2024-11-19
Make association reference set names less verbose for UI display purposes

apeteri:
https://snowowl.atlassian.net/browse/SO-6306

PR #1341, opened 2024-11-19
Make association reference set names less verbose for UI display purposes (9.x)

apeteri:
This is the counterpart of PR #1340.


cmark (approved):
🥓

AAAlinaaa (approved):

PR #1342, opened 2024-11-22
SO-6148  Exclude irrelevant axiom types from validation (9x)

AAAlinaaa:
7.x version here: https://github.com/b2ihealthcare/snow-owl/pull/1343


cmark (approved):
🥓

apeteri (approved):
👍 

There should also be a faulty GCI expression test which results in a validation issue according to the old code, but no longer creates an issue in the new version.

cmark:
@AAAlinaaa please resolve the conflict.

PR #1344, opened 2024-11-22
SO-6308  FULL import remote job fetch issues

AAAlinaaa:
7.x version in https://github.com/b2ihealthcare/snow-owl/pull/1345

PR #1345, opened 2024-11-22
SO-6308  FULL import remote job fetch issues (7x)

AAAlinaaa:
Port of https://github.com/b2ihealthcare/snow-owl/pull/1344

### DIFF
--- a/snomed/com.b2international.snowowl.snomed.core.rest.tests/src/com/b2international/snowowl/snomed/core/rest/SnomedRf2NextReleaseImportTest.java
+++ b/snomed/com.b2international.snowowl.snomed.core.rest.tests/src/com/b2international/snowowl/snomed/core/rest/SnomedRf2NextReleaseImportTest.java
@@ -70,7 +70,7 @@ public class SnomedRf2NextReleaseImportTest extends AbstractSnomedApiTest {
 		assertTrue("Failed to import RF2 archive", job.isSuccessful());
 		ImportResponse response = job.getResultAs(Services.context().service(ObjectMapper.class),  ImportResponse.class);
 		//assert that the previous import job didn't store all visited components
-		assertTrue("Import result of full import includes visited components", response.getVisitedComponents().isEmpty());
+		assertTrue("Import result of FULL RF2 import should NOT include any visited components", response.getVisitedComponents().isEmpty());
 		// assert that the version for importUntil is present in the system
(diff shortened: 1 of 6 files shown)


apeteri (approved):
👍

snomed/com.b2international.snowowl.snomed.core.rest.tests/src/com/b2international/snowowl/snomed/core/rest/SnomedRf2NextReleaseImportTest.java
```diff
@@ -66,6 +68,9 @@ private void doImport(final String importFile, final String importUntil) {
 			.getSync(1, TimeUnit.MINUTES);
 		RemoteJobEntry job = JobRequests.waitForJob(Services.bus(), jobId, 2000 /* 2 seconds */);
 		assertTrue("Failed to import RF2 archive", job.isSuccessful());
+		ImportResponse response = job.getResultAs(Services.context().service(ObjectMapper.class),  ImportResponse.class);
+		//assert that the previous import job didn't store all visited components
+		assertTrue("Import result of full import includes visited components", response.getVisitedComponents().isEmpty());
```
cmark:
```suggestion
		assertTrue("Import result of FULL RF2 import should NOT include any visited components", response.getVisitedComponents().isEmpty());
```

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/request/rf2/SnomedRf2ImportRequest.java
```diff
@@ -260,7 +260,11 @@ ImportResponse doImport(final BranchContext context, final File rf2Archive, fina
 				updateLocales(context, codeSystemUri.getCodeSystem());
 			}
 			
-			return ImportResponse.success(visitedComponents.build(), reporter.getDefects());
+			if (Rf2ReleaseType.FULL == releaseType) {
```
cmark: We could even use some sort of throw-away Set implementation that does not contain the added values so that it becomes empty forever. That we would only need to alter the builder variable above based on releaseType.

cmark (changes requested):

cmark (approved):
🥓

PR #1346, opened 2024-11-25
SO-6339: improve authorized context propagation

cmark:
When working with async streaming contexts the system was unable to propagate authorization information to the sub-request which was resulted in an error. This PR fixes that.

Jira: https://snowowl.atlassian.net/browse/SO-6339

### DIFF
--- a/core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/request/SearchPageableCollectionResourceRequestBuilder.java
+++ b/core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/request/SearchPageableCollectionResourceRequestBuilder.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2021-2022 B2i Healthcare, https://b2ihealthcare.com
+ * Copyright 2021-2024 B2i Healthcare, https://b2ihealthcare.com
  * 
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


apeteri (approved):
RSLGTM!

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/request/SearchPageableCollectionResourceRequestBuilder.java
```diff
@@ -42,25 +43,52 @@ public final Stream<R> stream(C context) {
...
+	 * @return
+	 * @deprecated - ensure you are using the alternative where you supply request headers if you are calling these from server environments. See referenced methods.
+	 * @see #streamAsync(IEventBus, Map, Function)
+	 */
+	public final Stream<R> streamAsync(IEventBus bus, Function<B, AsyncRequest<R>> req) {
+		return streamAsync(bus, Map.of(), req);
+	}
+	
+	public final Stream<R> streamAsync(IEventBus bus, Map<String, String> headers, Function<B, AsyncRequest<R>> req) {
+		return Streams.stream(new SearchResourceRequestIterator<B, R>(getSelf(), (builder) -> req.apply(builder).withHeaders(headers).execute(bus).getSync(3, TimeUnit.MINUTES)));
```
AAAlinaaa, line 66: Can we document these timeouts?

AAAlinaaa (approved):

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/request/SearchPageableCollectionResourceRequestBuilder.java
cmark: Somewhere yes, but that requires a bit more effort and there are a couple of these lying around in the code. Most of them are around 3 minutes. Probably we need to extract them into a static/dynamic config value. Will create a Jira for it.

PR #1350, opened 2024-11-28
fix(snomed): improve performance of component inactivation...

cmark:
...by skipping existing historical association member search for a given inactivated component when there is no historical reference set identifier present in the current semantic network (memberOf fields).

Also, caching the current association reference set IDs in the scope of the request, so these can be reused which improves performance for those components where there are existing assoc. members.

### DIFF
--- a/snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/request/SnomedAssociationReferenceSets.java
+++ b/snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/request/SnomedAssociationReferenceSets.java
@@ -22,9 +22,8 @@ import java.util.stream.Collectors;
 
 import com.b2international.snowowl.core.ServiceProvider;
 import com.b2international.snowowl.core.domain.BranchContext;
-import com.b2international.snowowl.core.domain.IComponent;
-import com.b2international.snowowl.snomed.common.SnomedConstants.Concepts;
-import com.b2international.snowowl.snomed.datastore.index.entry.SnomedConceptDocument;
+import com.b2international.snowowl.snomed.core.domain.refset.SnomedRefSetType;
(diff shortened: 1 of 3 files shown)


nagyo (approved):
LGTM!

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/request/SnomedAssociationReferenceSets.java
```diff
@@ -0,0 +1,59 @@
...
+
+	public SnomedAssociationReferenceSets(BranchContext context) {
+		this.context = checkNotNull(context, "context");
+	}
+	
+	public Set<String> get() {
+		if (associationReferenceSets == null) {
+			associationReferenceSets = SnomedRequests.prepareSearchConcept()
+					.all()
+					.filterByAncestor(Concepts.REFSET_ASSOCIATION_TYPE)
```
apeteri: It might not make that much of a difference, but a status filter and a reference set type filter could also be used here (which requires changing the concept search into a reference set search).

apeteri (approved):
👍

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/request/SnomedAssociationReferenceSets.java
cmark: We can add those, but keep in mind that inactive concepts don't have parents/ancestors so that's probably just good to have for clarity, regarding refset types that's a good point that returns only the true refsets with at least one member associated.

cmark: Done.

PR #1351, opened 2024-12-02
fix(snomed.datastore): Flush pending expand requests in request cache...

apeteri:
...when exporting content to DSV; otherwise the nested expansion for relationship destination PTs is deferred until the export finishes, which is of no use at that point.


cmark (approved):
🥓

nagyo (approved):
LGTM!

PR #1352, opened 2024-12-05
OWL Relationship utility methods

AAAlinaaa:
Meant to aid with statement fragment generation used by classifiers and to ease conversion between its document form


cmark (approved):
Let's keep this here for now.

apeteri (approved):

PR #1353, opened 2024-12-06
Add nested context descriptions to lock documents

apeteri:
This is required to track re-entrant locks that go more than two levels deep.

### DIFF
--- a/core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/locks/DatastoreLockIndexEntry.java
+++ b/core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/locks/DatastoreLockIndexEntry.java
@@ -39,12 +39,13 @@ import com.google.common.collect.ImmutableList;
 /**
  * @since 7.1.0
  */
-@Doc(type = "lock")
+@Doc(type = "lock", revisions = {
+	@SchemaRevision(version = 2L, description = "Introduce contexts and timestamp field", strategy = DocumentMappingMigrationStrategy.NO_REINDEX)
+})
 @JsonDeserialize(builder = DatastoreLockIndexEntry.Builder.class)
-@SchemaRevision(version = 2L, description = "Introduce contexts field", strategy = DocumentMappingMigrationStrategy.NO_REINDEX)
 public final class DatastoreLockIndexEntry implements Serializable {
 	
-	private static final long serialVersionUID = 1L;
+	private static final long serialVersionUID = 2L;
 	
 	public static final class Fields {
 		public static final String ID = "id";
--- a/core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/locks/DatastoreLockIndexEntry.java
+++ b/core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/locks/DatastoreLockIndexEntry.java
@@ -17,9 +17,9 @@ package com.b2international.snowowl.core.locks;
 
 import static com.b2international.index.query.Expressions.exactMatch;
 import static com.b2international.index.query.Expressions.matchAny;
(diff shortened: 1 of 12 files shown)


core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/locks/OperationLock.java
```diff
@@ -33,8 +34,8 @@ public class OperationLock extends AbstractOperationLock {
 	 * @param id the lock identifier
 	 * @param target the lock target (may not be {@code null})
 	 */
-	public OperationLock(final int id, final Lockable target) {
-		super(id, target);
+	public OperationLock(final int id, final Date creationDate, final Lockable target) {
```
nagyo, line 37: Wouldn't it be better to use a LocalDate type here?

nagyo (approved):
RSLGTM! with a note

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/locks/OperationLock.java
cmark: Date includes the time aspect as well, not just a date, so LocalDateTime would be better, but that requires a larger refactor. Keep it as is for now.

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/locks/DatastoreLockIndexEntry.java
```diff
@@ -140,26 +139,47 @@ public Builder branchPath(final String branchPath) {
...
 	private final String repositoryId;
 	private final String branchPath;
-	
-	private DatastoreLockIndexEntry(final String id, final String userId, final String description, final String parentDescription, final String repositoryId, final String branchPath) {
+	private final Long timestamp;
+	private final List<String> contexts;
+
+	// Only left in for backwards compatibility reasons
+	@Deprecated
+	private final String description = null;
```
cmark, line 177: We currently don't throw away stuck lock entries during restart as we don't know which machine initiated it (yet), therefore there might be a case when there is a lock entry that uses the old format and the system would try to deserialize it but there is not setter in the JSON builder. I'll fix this soon.

cmark (changes requested):

cmark (approved):

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/locks/DatastoreLockIndexEntry.java
cmark: Fixed by [13be6a6](https://github.com/b2ihealthcare/snow-owl/pull/1353/commits/13be6a655c924a0bb41a69a6912a2c651df1cc7d)

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/locks/DatastoreLockIndexEntry.java
```diff
@@ -143,6 +144,20 @@ public Builder timestamp(final Long timestamp) {
...
+		Builder description(final String description) {
+			// lock documents with description value should propagate their values to contexts
+			addContext(description);
+			return this;
+		}
+		
+		@JsonSetter
+		Builder parentDescription(final String parentDescription) {
+			// lock documents with a parentDescription value should propagate their values to contexts
+			addContext(parentDescription);
```
apeteri, line 158: `parentDescription` goes first in the list, so a `contexts.add(0, parentDescription)` equivalent could be used here. Although whether this is required or not will also depend on the order of the two setter calls.

PR #1354, opened 2024-12-10
MRCM: Ensure refset range constraints are properly evaluated

AAAlinaaa:


cmark (approved):
🥓

snomed/com.b2international.snowowl.validation.snomed/src/main/resources/scripts/rule_mrcm_constraint.groovy
```diff
@@ -190,7 +190,7 @@ for (String typeId : allowedRanges.keySet()) {
 			String ancestorId = clause.replaceAll("<", "").strip();
 			ancestorIds.add(ancestorId);
 		} else if (clause.startsWith("^")) {
-			String refsetId = clause.replaceAll("^", "").strip();
+			String refsetId = clause.replace("^", "").strip();
```
nagyo, line 193: Looks OK, but maybe using a `Charmatcher.is('^').trimLeadingFrom(clause)` would be a more robust solution for all of these string manipulations. According to the javadoc of the Pattern class, there are some special cases where `<` and `!` could fall into the special character category, which could also be interpreted as a Java regex expression.

nagyo (approved):
LGTM! with a note

snomed/com.b2international.snowowl.validation.snomed/src/main/resources/scripts/rule_mrcm_constraint.groovy
AAAlinaaa: Groovy's default is method overwrites the CharMatcher's is and the result would be a boolean not a CharMatcher.

nagyo: There is `anyOf()` as well

PR #1358, opened 2025-01-30
SO-6369: Fix handling of enums in revision conflict processor (7.x)

AAAlinaaa:


apeteri:
Can we get a test case that shows what happens to `Enum` values?

apeteri:
The build failed at the Codecov upload so we know that expectations in existing test cases aren't challenged.

apeteri (approved):
👍 

The test case fails with "Couldn't parse json" before the change in `DocumentMapping` is applied, but succeeds afterwards. `isObject` is only called in connection with conflict processing, so it does not affect e.g. mapping for existing indexes.

commons/com.b2international.index.tests/src/com/b2international/index/revision/RevisionBranchMergeConflictTest.java
```diff
@@ -100,6 +103,21 @@ public void rebaseContainerDeletedOnTargetAndComponentAddedOnSource() throws Exc
...
+		indexRevision(MAIN, data);
+		String branchA = createBranch(MAIN, "a");
+		
+		EnumPropertyData updateOnMain = new EnumPropertyData(STORAGE_KEY1, Option.OPTION_A);
+		indexChange(MAIN, data, updateOnMain);
+		
+		EnumPropertyData updateOnBranch = new EnumPropertyData(STORAGE_KEY1, Option.OPTION_A);
+		indexChange(branchA, data, updateOnBranch);
+		
+		branching().prepareMerge(MAIN, branchA).merge();
```
nagyo, line 119: It would be nice to assert the state of the enum after the merge. If I read the code (and not the method name) it's not obvious at first what we are trying to verify here.

nagyo (approved):
LGTM with a note

commons/com.b2international.index.tests/src/com/b2international/index/revision/RevisionBranchMergeConflictTest.java
apeteri: Unfortunately `assertDoesNotThrow` appears in JUnit 5 only: https://junit.org/junit5/docs/current/api/org.junit.jupiter.api/org/junit/jupiter/api/Assertions.html#assertDoesNotThrow(org.junit.jupiter.api.function.Executable)

The test method breaks with an exception before the fix, and it needs to complete in its entirety to pass.

nagyo: I meant something like:
```
assertEquals(OPTION_A, getRevision(MAIN, data).getOption())
```
so we can make sure that the value does not change as a result of the merge.

cmark:
Late LGTM!

PR #1364, opened 2025-02-12
SO-6377: allow multiple same type identity providers

cmark:


core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/identity/IdentityPlugin.java
```diff
@@ -115,31 +120,32 @@ public void init(SnowOwlConfiguration configuration, Environment env) throws Exc
...
+			Optional<IdentityProviderFactory> factory = factories.stream().filter(f -> f.getConfigType() == providerConfig.getClass()).findFirst();
+			if (factory.isPresent()) {
+				try {
+					providers.add(factory.get().create(providerConfig));
+				} catch (Exception e) {
+					throw new SnowOwl.InitializationException(String.format("Couldn't initialize '%s' identity provider", factory), e);
 				}
-			});
+			} else {
+				throw new SnowOwl.InitializationException("Couldn't find identity provider factory for '%s'.", providerConfig.getClass());
```
apeteri: It should be a rare occasion that the configuration class can be deserialized correctly, but the corresponding factory can not be found via plugin lookup (in most cases these will be co-located in the same OSGi bundle). This is more like a development-time issue rather than something that would happen in production.

I think it's OK to use the same exception type either way.

apeteri (approved):

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/identity/IdentityPlugin.java
cmark: Yeah, I was thinking of changing this to IllegalStateException to indicate a programmatic related error rather than a configuration or other error.

cmark: Changed it.

AAAlinaaa (approved):

nagyo (commented):
🍏

PR #1365, opened 2025-02-14
SO-6267: Update handling of validation thread pool size (7.x)

nagyo:
- Change the property name from numberOfValidationThreads to workerPoolSize
- Allow setting either property for backwards compatibility (setting both results in undefined behavior)
- Use null as the default value
- Increase allowed maximum pool size to 99
- Compute a reasonable default value on the server side if the value remains unset in Snow Owl's configuration file

Backport of #1320


AAAlinaaa (approved):

apeteri (approved):
LGTM!

apeteri:
Codecov and FOSSA are reporting confusing errors for this branch which we will resolve at a later time. Merging.

PR #1366, opened 2025-02-14
SO-6372: RF2 import result dialogue bug

adamfilep:
Client-side PR: https://github.com/b2ihealthcare/snow-owl-ext/pull/2903

### DIFF
--- a/core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/request/io/ImportResponse.java
+++ b/core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/request/io/ImportResponse.java
@@ -40,15 +40,18 @@ public final class ImportResponse implements Serializable {
 	private final String error;
 	private final Set<ComponentURI> visitedComponents;
 	private final List<ImportDefect> defects;
+	private final int changeCount;
 	
 	@JsonCreator
 	public ImportResponse(
 			@JsonProperty("error") final String error, 
 			@JsonProperty("visitedComponents") final Set<ComponentURI> visitedComponents, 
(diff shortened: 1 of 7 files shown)


core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/request/io/ImportResponse.java
```diff
@@ -84,21 +87,29 @@ public String getError() {
...
+		return new ImportResponse(error, Set.of(), List.of(), 0);
 	}
 
 	public static ImportResponse success(Set<ComponentURI> visitedComponents) {
 		return success(visitedComponents, List.of());
 	}
 	
 	public static ImportResponse success(Set<ComponentURI> visitedComponents, List<ImportDefect> defects) {
-		return new ImportResponse(null, visitedComponents, defects);
+		return new ImportResponse(null, visitedComponents, defects, 0);
```
apeteri: This `0` can be a bit misleading on the UI side; the conditions in the PR do not prevent us from displaying e.g. `0 component(s) were imported, 99 items are displayed` when this factory method is called with a non-empty `visitedComponents` Set.

AAAlinaaa (approved):

apeteri:
A serialization test case still needs to be amended as `"changeCount": 1` is now also returned:

```
Failures: 
  ImportResponseSerializationTest.testSerialization:56 
  expected:<...e","type":"INFO"}],"[]success":true}> 
  but was: <...e","type":"INFO"}],"[changeCount":1,"]success":true}>
```

apeteri (approved):
LGTM!

PR #1371, opened 2025-03-07
refactor(core): Move evaluation of ECL filter conjunctions and disjunctions to superclass

apeteri:
Previously this was only supported in `SnomedEclEvaluationRequest` but all subclasses should be able to make use of this functionality.


cmark (approved):
❤️

AAAlinaaa (approved):

PR #1374, opened 2025-03-24
SO-6396: fix token refresh for tokens issued by other servers

cmark:
Jira: https://snowowl.atlassian.net/browse/SO-6396


apeteri (approved):
LGTM with a note: we will also have to remember to set `issuer` to use different values in configurations that have multiple JWT configs. Should this be enforced at initialization time?

nagyo (approved):
LGTM! and +1 for András's comment

cmark:
> LGTM with a note: we will also have to remember to set `issuer` to use different values in configurations that have multiple JWT configs. Should this be enforced at initialization time?

I think the best would be to rely on the actual deployment ID, like the assigned hostname for example when configuring the service. Something for the devops team 😄

PR #1387, opened 2025-05-06
SO-6405: Resolve simple type DSV export issues

apeteri:
This PR changes the structure of property occurrence counters in `SnomedSimpleTypeRefSetDSVExporter` to ensure that the correct amount of empty space is added for properties and groups which do not appear in currently exported concept.

Previously these were keyed by group, then by type; the commit swaps this around so that the type ID becomes the first key.


cmark (approved):
🥓

AAAlinaaa (approved):

PR #1390, opened 2025-06-06
SO-6421: concrete domain refset datatype config validation

cmark:
Jira: https://snowowl.atlassian.net/browse/SO-6421


snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/index/entry/SnomedRefSetMemberIndexEntry.java
```diff
@@ -937,6 +937,10 @@ public SnomedRefSetMemberIndexEntry build() {
 					break;
 				default: throw new UnsupportedOperationException("Unsupported concrete domain data type: " + dataType);
 				}
+			} else {
+				if (value != null) {
```
apeteri, line 941: We could check and throw an exception if `(value == null)` in the `(dataType != null)` case above as well. This would mean that either both fields are set or neither of the two: 

```java
if ((value == null) != (dataType == null)) {
  throw new IllegalStateException("...");
}

// The commented part is always true because we have passed the test above
if (dataType != null /* && value != null */) { 
  switch (dataType) {
    ...
  }
}
```

...unless we allow `null` to be set when a `dataType` has been defined for some reason.

apeteri (approved):
👍 with a bonus verification suggestion.

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/index/entry/SnomedRefSetMemberIndexEntry.java
cmark: String members might not have an actual value, not sure what is our preference here. If every member should have a non-null value then yes, we can add this.

AAAlinaaa (approved):

AAAlinaaa (approved):

apeteri (approved):
🍃

PR #1396, opened 2025-07-14
SO-5954 Donated concept upgrade bug

AAAlinaaa:


commons/com.b2international.index/src/com/b2international/index/revision/Conflict.java
```diff
@@ -15,13 +15,17 @@
...
+
 /**
  * @since 7.0
  */
 public abstract class Conflict {
 
 	private final ObjectId objectId;
 	private final String message;
+	
+	private ObjectId containerId;
```
apeteri: The new field is not included in `hashCode` - `equals` computations, is that allowed?

apeteri (approved):
👍

commons/com.b2international.index/src/com/b2international/index/revision/Conflict.java
cmark: It is okay, this data is auxiliary to the conflict and just carries the data for later use. We can add a comment about this to both branches.

cmark (approved):
🥓

PR #1409, opened 2025-09-11
SO-6441: add admin CLI command to synchronize identifiers between the local store and the existing content

nagyo:
- this is considered to be a point-solution for now


apeteri (approved):
RSLGTM!

AAAlinaaa (commented):
LGTM

PR #1421, opened 2025-10-14
SO-6015: Amend security header configuration

nagyo:


cmark (approved):
🥓

nagyo:
Based on more research we have decided to go with the default '0' value instead for the X-XSS-Protection header.

PR #1424, opened 2025-10-17
SO-6491: Resolve issue with tracking document scores in EsQueryBuilder (9.x)

apeteri:
Jira: https://snowowl.atlassian.net/browse/SO-6491


cmark (changes requested):
Can we add a low-level unit test to verify this change? 🙇

apeteri:
Added in commit https://github.com/b2ihealthcare/snow-owl/pull/1424/commits/aeb8a6f127eec716d4e9ae866b95d1bb4dd47dbd. It fails without the fix in `EsQueryBuilder` and passes when it is applied.

cmark (approved):
🥓

AAAlinaaa (approved):

PR #1427, opened 2025-11-18
fix(dependencies): include null check for first dependencies list too

AAAlinaaa:


apeteri (approved):
If there is no reason to "override" (or rather, extend) a null or empty dependencies list, then this should be OK.

cmark (approved):
🥓

PR #1431, opened 2025-11-27
INFRA-504: Fix memory leak issues with build pipeline

eMarci:
This PR fixes a memory leak warning with the build pipeline. Two vars in the `checkout` stage were missing the `def` keyword, making them global, and vulnerable to race conditions.


nagyo (approved):
LGTM!

PR #1008, opened 2022-04-20
Support Elasticsearch 8 clusters

cmark:
This PR adds the ability to connect to Elasticsearch 8 clusters via the API compatibility mode using the now deprecated high-level Java REST client.
The PR also adds the ability to test against Elasticsearch 8 with https://www.testcontainers.org/modules/elasticsearch/

PR #1442, opened 2026-01-15
SO-6540: backport Elasticsearch 8 compatibility to 7.x

cmark:
This PR backports the ability to connect to an Elasticsearch 8 cluster in older 7.x versions of Snow Owl.


commons/com.b2international.index.tests.tools/src/com/b2international/index/IndexResource.java
```diff
@@ -33,18 +44,27 @@
...
 import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.Maps;
 
 /**
  * @since 7.1
  */
 public final class IndexResource extends ExternalResource {
 
+	/**
+	 * Java system property to configure the use of a testcontainer Elasticsearch Docker container and optionally configure the actual image as well. By default it uses the 8.1.3 image.
```
apeteri: This comment should be changed to become always up-to-date 🌲 by using e.g. `{@value #DEFAULT_ES_DOCKER_VERSION}` and changing `DEFAULT_ES_DOCKER_IMAGE` to use this newly introduced constant that holds the version part of the image.

commons/com.b2international.index.tests.tools/src/com/b2international/index/IndexResource.java
```diff
@@ -59,13 +79,47 @@ private IndexResource(Collection<Class<?>> types, Consumer<ObjectMapper> objectM
...
+					testElasticsearchContainer = DEFAULT_ES_DOCKER_IMAGE;
+				}
+				container = new ElasticsearchContainer(testElasticsearchContainer);
+				// XXX elasticsearch-default-memory-vm.options is a classpath resource in the testcontainers:elasticsearch jar since 7.17.4
+				// loading it from the classpath won't work because testcontainers is not ready to handle bundleresource URLs specific to Eclipse OSGi 
+				// remove the entry and replace it with ours
+				container.getCopyToFileContainerPathMap().keySet().removeIf(file -> file.getFilesystemPath().startsWith("bundleresource://") && file.getFilesystemPath().contains("elasticsearch-default-memory-vm.options"));
+				container.withCopyFileToContainer(MountableFile.forHostPath(toAbsolutePathBundleEntry(IndexResource.class, "elasticsearch-default-memory-vm.options")), "/usr/share/elasticsearch/config/jvm.options.d/elasticsearch-default-memory-vm.options");
+				
+				container.withEnv("rest.action.multi.allow_explicit_index", "false");
```
apeteri: Style nitpick: if no method chaining is done, `copyFileToContainer` and `addEnv` should be used here instead. `copyFileToContainer` is also used below in an added code block for copying the synonyms file.

apeteri (approved):
👍 with two suggestions.

nagyo (approved):
RSLGTM

AAAlinaaa (approved):

nagyo:
Not sure if it should be part of this PR or added separately, but we will need the ability to start test environments with an attached Elasticsearch container, see the changes for SnowOwlAppRule in #1008

cmark:
> Not sure if it should be part of this PR or added separately, but we will need the ability to start test environments with an attached Elasticsearch container, see the changes for SnowOwlAppRule in #1008

This is already part of the code and with the last two commits I enabled it to run for the discussed index-http and snomed-e2e tests in all cases. These test suites now only run in dockerized envs and not with the embedded mode.

commons/com.b2international.index.tests.tools/src/com/b2international/index/IndexResource.java
cmark: Added in commit https://github.com/b2ihealthcare/snow-owl/pull/1442/commits/155b8ee9d52ee7170133e6ecafdd6b8057423819.

commons/com.b2international.index.tests.tools/src/com/b2international/index/IndexResource.java
cmark: Unfortunately, this is no good. The non-with method try to modify a running container so they all fail to be called before the start method call. Reverting but adding method chaining to make the temporal coupling explicit.

apeteri (approved):
LGTM!

nagyo (changes requested):
The SnowOwlAppRule changes (referenced here: https://github.com/b2ihealthcare/snow-owl/pull/1442#issuecomment-3784032829) are not present on the PR and because of this the snomed-e2e tests are still running in embedded mode during the maven build. 

When looking at the build logs it is impossible to tell that the test suite is running in embedded mode or not. This is because we suppress all logs for each suite. It might be useful to open up the testcontainers logs to be able to inspect the things happening in the background. An example config already exists here: https://github.com/b2ihealthcare/snow-owl-ext/blob/10.x/syndication/com.b2international.snowowl.syndication.integration.tests/src/configuration/logback-maven.xml#L3

cmark:
> The SnowOwlAppRule changes (referenced here: [#1442 (comment)](https://github.com/b2ihealthcare/snow-owl/pull/1442#issuecomment-3784032829)) are not present on the PR and because of this the snomed-e2e tests are still running in embedded mode during the maven build.
> 
> When looking at the build logs it is impossible to tell that the test suite is running in embedded mode or not. This is because we suppress all logs for each suite. It might be useful to open up the testcontainers logs to be able to inspect the things happening in the background. An example config already exists here: https://github.com/b2ihealthcare/snow-owl-ext/blob/10.x/syndication/com.b2international.snowowl.syndication.integration.tests/src/configuration/logback-maven.xml#L3

Updated the execution profiles, locally it worked. Please check again.

nagyo (approved):
LGTM!

commons/com.b2international.index/src/com/b2international/index/es/EsDocumentWriter.java
```diff
@@ -195,7 +194,7 @@ public void afterBulk(long executionId, BulkRequest request, BulkResponse respon
 								.opType(OpType.INDEX)
 								.source(_source, XContentType.JSON);
 						// XXX revisions has their special local ID, but that's not needed when sending them to ES, ES will autogenerate a non-conflicting ID for them
```
apeteri: This comment is no longer applicable after changes on the PR, the `if` block no longer tests for `Revision` instances.

apeteri (approved):
👍 for fresh commits with a small note.

Please also check whether any direct references to `org.elasticsearch.common.UUIDs` appear in the changed files, I think I saw one but can't find it at the moment. We should strive for using `com.b2international.snowowl.core.id.IDs` everywhere for consistency.

commons/com.b2international.index/src/com/b2international/index/es/EsDocumentWriter.java
cmark: Done, and also checked the usage of `es.common.UUIDs`, we don't have any references other than the allowed `core.id.IDs`.

apeteri (approved):

PR #1398, opened 2025-07-18
SO-6448: donated content reappears as added in revision compare results

cmark:
Jira: https://snowowl.atlassian.net/browse/SO-6448

PR #1445, opened 2026-01-30
SO-6533: Multiple revisions visible after merging changes in succession (7.x)

apeteri:
This PR extends the conditions under which a commit detail entry is added to a merge commit document. These are deliberately kept sparse but for proper housekeeping the server needs know about objects that:

- are part of an "added vs. added" conflict that is recognized as content donation, so one side of the contribution should be considered deleted: #1398 
- were changed on both sides in a non-conflicting manner _or_ end up having the same state (this PR)

This helps maintaining a proper created-revised timestamp chain for revisions in consecutive merge operations.


cmark (approved):
🥓 Please carry this over to 10.x.

adamfilep (approved):
LGTM 👌

AAAlinaaa (approved):

PR #1448, opened 2026-02-04
SO-6550: Support "*" for preferred term and FSN resolution

apeteri:
This PR changes the "determine best matching description" step when expanding preferred terms and fully specified names for SNOMED CT concept to allow adding `*` to the list of locales. This symbol is replaced in the same position with the list of known locales stored in terminology metadata, filed under the `locales` key.

Related behavior like replacing a single `*` input in an HTTP request's `Accept-Language` header with a built-in value is not modified.


cmark (approved):
Generally looks good. One question: do we want to include the wildcard at the end of every Accept-Language header default value?

apeteri:
I didn't want to modify this single wildcard behavior out of caution, and there are also a lot of constants on REST controller classes: https://github.com/b2ihealthcare/snow-owl/pull/1130

AAAlinaaa (approved):

PR #1460, opened 2026-03-03
INFRA-555: Restrict GitHub token permissions in Maven workflow

eMarci:
- Add explicit job-level permissions with least-privilege principle
- Remove Nexus deployment credentials and Maven settings setup


eMarci:
Closed because the branch name was not right, replaced by #1461

PR #1471, opened 2026-03-12
feat(index): support refresh policy to be defined in document level

cmark:
Two policies are supported, one for the standard refresh and another for the deferred (after standard phase) refresh.


apeteri (approved):
LGTM!

nagyo (approved):
LGTM!

PR #1472, opened 2026-03-18
SNRAY-2075: JWT/JWKS identity configuration changes

apeteri:
- Rename `emailClaimProperty` to `userIdClaimProperty`, retaining property setters for backwards compatibility purposes
- Add test cases to verify claim extraction from decoded JWT
- Add test cases to verify whether using the previous property name in configuration files still registers the value correctly


cmark (approved):
🥓

AAAlinaaa (approved):

PR #1477, opened 2026-03-25
SO-6563: upgrade to Eclipse Platform 2026-03

cmark:
Jira: https://snowowl.atlassian.net/browse/SO-6563


apeteri (approved):
LGTM with some questions, notes:

- `RevisionBranchMergeTest.rebaseBranchOnParentWithChangedRevision()` is failing intermittently, both on 10.x and on the PR branch, might need a little more investigation
- The request in `CapabilityStatementApiTest.capabilityStatementTest_5_0_0()` never completes, both on 10.x and here as well
- I haven't tried the `snomed-e2e-tests (cis)` launch configuration as it requires external CIS setup. Would it make sense to point towards e.g. a released version of Snow Owl CIS with testcontainers?
- `slf4j.api*2.0.17` is appearing in launch configurations although there is only one version in the target platform. My IDE is taking the version part away diligently on each visit to these launch configs 😄 Do you have a different set of bundles available in your setup?
- Is contiperf completely out of the picture, do you intend to add a replacement for performance tests later?

cmark:
> RevisionBranchMergeTest.rebaseBranchOnParentWithChangedRevision() is failing intermittently, both on 10.x and on the PR branch, might need a little more investigation

Raise a Jira task for this, please, with some stack traces and errors. Thanks!

> The request in CapabilityStatementApiTest.capabilityStatementTest_5_0_0() never completes, both on 10.x and here as well

I'll take a look at that as all tests were green on my computer, on GitHub Actions and on Jenkins as well earlier.

> I haven't tried the snomed-e2e-tests (cis) launch configuration as it requires external CIS setup. Would it make sense to point towards e.g. a released version of Snow Owl CIS with testcontainers?

I think this test requires a bit of overhaul, which we can again do in a dedicated task.

> slf4j.api*2.0.17 is appearing in launch configurations although there is only one version in the target platform. My IDE is taking the version part away diligently on each visit to these launch configs 😄 Do you have a different set of bundles available in your setup?

That is a remnant when the tp contained both 2.0.16 and 17, I will update all launch configs before merge.

> Is contiperf completely out of the picture, do you intend to add a replacement for performance tests later?

Contiperf is unmaintained and we only used it at two places which can be simply replaced with pure JUnit or Java constructs. For better performance tests we either use JMH or some other lib that is maintained.

apeteri (approved):

AAAlinaaa (approved):

PR #1488, opened 2026-04-09
Remove concrete domain member support

AAAlinaaa:
Spinoff of https://snowowl.atlassian.net/browse/SO-6567


snomed/com.b2international.snowowl.snomed.common/src/com/b2international/snowowl/snomed/common/SnomedConstants.java
```diff
@@ -184,11 +184,17 @@ private Concepts() { }
...
+		@Deprecated
 		public static final String REFSET_INTEGER_DATATYPE = "373998411000154109";
+		@Deprecated
 		public static final String REFSET_FLOAT_DATATYPE = "744104701000154109";
+		@Deprecated
 		public static final String REFSET_STRING_DATATYPE = "513945551000154100";
+		@Deprecated
 		public static final String REFSET_CONCRETE_DOMAIN_TYPE = "289191171000154104";
 		public static final String REFSET_DEFINING_TYPE = "384696201000154108";
 		public static final String REFSET_MEASUREMENT_TYPE = "945726341000154109";
```
apeteri, line 202: I can't find any code referring to these two either (I have the usual external projects open as well, not just `snow-owl`).

snomed/com.b2international.snowowl.snomed.common/src/com/b2international/snowowl/snomed/common/SnomedRf2Headers.java
```diff
@@ -135,7 +135,7 @@ public abstract class SnomedRf2Headers {
 		FIELD_CORRELATION_ID,
 		// description type
 		FIELD_DESCRIPTION_FORMAT,
-		// concrete domain
+		// concrete value
 		FIELD_CHARACTERISTIC_TYPE_ID,
 		FIELD_TYPE_ID,
```
apeteri: With the removal of CD members these field names are not showing up on any other reference set member type, so they can be removed entirely.

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/core/ecl/SnomedEclRefinementEvaluator.java
```diff
@@ -363,75 +359,16 @@ private Promise<Collection<Property>> evalRefinement(final BranchContext context
...
-				final Promise<Collection<Property>> members = evalMembers(context, focusConceptIds, typeConceptFilter, (DataTypeComparison) comparison);
-				return Promise.all(statementsWithValue, members).then(results -> {
-					final Collection<Property> s = (Collection<Property>) results.get(0);
-					final Collection<Property> m = (Collection<Property>) results.get(1);
-					return FluentIterable.concat(s, m).toSet();
+				return statementsWithValue.then(results -> {
+					final Collection<Property> s = (Collection<Property>) results;
+					return FluentIterable.from(s).toSet();
 				});
 			}
```
apeteri: As we no longer have to mix two sources, `FluentIterable.from(s).toSet()` can definitely be replaced with e.g. `ImmutableSet.copyOf(s)`, but I'm also thinking that a single evaluation will not produce duplicate instances of `Property` either, so `statementsWithValue` could be used directly here. Please check!

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/index/entry/SnomedRefSetMemberIndexEntry.java
```diff
@@ -1013,17 +1013,29 @@ public SnomedRefSetMemberIndexEntry build() {
...
 	private Integer integerValue;
+	@Deprecated
 	private BigDecimal decimalValue;
 
+	@Deprecated
 	private Integer relationshipGroup;
+	@Deprecated
 	private String typeId;
+	@Deprecated
 	private String characteristicTypeId;
```
apeteri, line 986: We have discussed it outside the PR but I'll tag @cmark on this: what is the policy on removing fields from an index? Existing datasets might have these declared in the mapping but there are no documents that stored any values for them.

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/index/entry/SnomedRefSetMemberIndexEntry.java
apeteri: These `Expression`s are related to CD members only, so can be safely retired. In this case the compiler might also reveal any spots that were missed.

apeteri, line 447: I'm not sure how much of the `Builder` needs to be emptied out, options are:

- Keep as is
- Remove fields related to CD members, keep builder methods with an empty body
- Remove all references to CD members, add a Jackson annotation that tells the de-serialization process to ignore any fields related to members

Deferring to @cmark on this question as well as it is related to the other one.

snomed/com.b2international.snowowl.snomed.reasoner/src/com/b2international/snowowl/snomed/reasoner/SnomedReasonerPlugin.java
```diff
@@ -61,8 +61,7 @@ public Collection<Class<?>> getAdditionalMappings() {
 			EquivalentConceptSetDocument.class,
 			ConceptChangeDocument.class,
 			DescriptionChangeDocument.class,
-			RelationshipChangeDocument.class,
-			ConcreteDomainChangeDocument.class
```
apeteri, line 65: Same here re: index changes, we are removing the mapping for an entire index.

apeteri (changes requested):

snomed/com.b2international.snowowl.snomed.common/src/com/b2international/snowowl/snomed/common/SnomedRf2Headers.java
AAAlinaaa: True

snomed/com.b2international.snowowl.snomed.common/src/com/b2international/snowowl/snomed/common/SnomedConstants.java
AAAlinaaa: I've marked these as deprecated too.

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/core/ecl/SnomedEclRefinementEvaluator.java
AAAlinaaa: If there are duplicate relationships that match the refinement (same source, type, value and group) then there will be duplicate Property instances too, the other return statements in this method return lists too though.

apeteri (approved):
The non-resolved questions in the previous review about index maintenance are still pending, but the rest looks good to me.

snomed/com.b2international.snowowl.snomed.reasoner/src/com/b2international/snowowl/snomed/reasoner/SnomedReasonerPlugin.java
cmark: In a sense this will be backward compatible since the index won't be checked/loaded at all so it is not a huge problem, we can delete these. Alternatively we can keep them and mark as deprecated.

cmark (approved):
🥓

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/index/entry/SnomedRefSetMemberIndexEntry.java
cmark: Yes, I discussed this with Alina as well, and these should be kept and marked as Deprecated. Destructive index changes are not allowed unless we bump the major version, but we don't want to do that yet.

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/index/entry/SnomedRefSetMemberIndexEntry.java
cmark: The functionality should be restored to the point that we are still able to load these as members if they are present in the index, so I vote for keep as is with deprecated annotations everywhere.

cmark: Looks good now.

cmark (approved):

PR #1492, opened 2026-04-13
fix(fhir): Prevent StackOverflowError when logging FHIR requests

apeteri:
Jackson by default does not work with the auto-create mechanism on FHIR model objects, leading to infinite recursion. Resources should also not appear in full in request logs.

Exhausting the call stack leads to JVM agent errors as well as class loading issues in integration tests.


cmark:
Hmm, should we apply the JSON ignore on the Parameter objects in the https://github.com/b2ihealthcare/fhir-core repository instead?

cmark:
> Hmm, should we apply the JSON ignore on the Parameter objects in the https://github.com/b2ihealthcare/fhir-core repository instead?

For the ConceptMaps and others it is worth adding the JsonIgnore annotation here, but for the parameters we might want to include some of the critical parameters in the JSON log. Maybe we can do this via getters in the request classes to expose the ones like url or system.

apeteri:
This would introduce a dependency on Jackson, but it also wouldn't help. The issue occurs with any datatype, not just resources. I'd add `JsonInclude({ prop1, prop2, ... })` here in that case (or add the array of approved properties to the parameter classes) and add the Java-friendly getters that you mention.

cmark:
> This would introduce a dependency on Jackson, but it also wouldn't help. The issue occurs with any datatype, not just resources. I'd add `JsonInclude({ prop1, prop2, ... })` here in that case (or add the array of approved properties to the parameter classes) and add the Java-friendly getters that you mention.

Hmm, for some reason I thought Jackson is imported by fhir-core but if not, then let's keep them here and expose just the usual suspects from the FHIR operation parameters that we want to see in logs, can you add just a few (at least the usual ones) so we don't loose this info when we merge this into main? Thanks!

apeteri:
It is excluded so it would not appear as an embedded dependency, but through bndtools import instructions we can use the version from the configured target platform. 

I have a Jackson-to-internal JSON bridge as a proof-of-concept which I'll be adding as a PR to the other project. But if we decided to use that, it would still be hard to find a good spot for registering the module to the `ObjectMapper` instance used for request logs - core does not know anything about FHIR classes.

cmark:
> It is excluded so it would not appear as an embedded dependency, but through bndtools import instructions we can use the version from the configured target platform.

Yes, I see this now, it is excluded as a dep and not added back as an import either as we don't need it for mainly a type only fhir-core lib.

> I have a Jackson-to-internal JSON bridge as a proof-of-concept which I'll be adding as a PR to the other project. But if we decided to use that, it would still be hard to find a good spot for registering the module to the ObjectMapper instance used for request logs - core does not know anything about FHIR classes.

Do we need that bridge there if we handle everything here? Bit confused now 😅

apeteri:
Requires https://github.com/b2ihealthcare/fhir-core/pull/5.

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/repository/ObjectMapperCustomizer.java
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * @since 10.0.1
+ */
+@FunctionalInterface
+public interface ObjectMapperCustomizer {
+
+	/**
+	 * Can be used to register custom serializers, deserializers, modules, etc. to
+	 * the main {@link ObjectMapper} instance used by Snow Owl for all JSON
```
cmark:
```suggestion
	 * the main {@link ObjectMapper} instance used by Snow Owl for core JSON
```

cmark (commented):
So far 🥓 but we need to merge the `fhir-core` changes first, get it built and verified here.

cmark (approved):
🥓

AAAlinaaa (approved):

cmark (approved):
🥓

AAAlinaaa (approved):

PR #1495, opened 2026-04-16
SO-6567: Move allowed MRCM expressions configurations to resource settings

AAAlinaaa:
Jira: https://snowowl.atlassian.net/browse/SO-6567


snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/config/SnomedCoreConfiguration.java
```diff
@@ -46,9 +45,6 @@ public class SnomedCoreConfiguration {
 	@Max(60)
 	private long classificationCleanUpInterval = DEFAULT_CLASSIFICATION_CLEANUP_INTERVAL;
 	
-	@NotNull
-	private SnomedMrcmConfig mrcmConfiguration = new SnomedMrcmConfig();
```
apeteri, line 50: Just like with the index, there are also many ways to deal with the configuration change:

- if "allow extra properties" is enabled for the YAML parser, the setting could just hang around in `snowowl.yml`
- otherwise it will cause an error at startup and Snow Owl will require an administrator to remove this section first
- a third option is to let it sit here, print a warning at startup and don't consult the values contained in this section

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/request/SnomedMrcmTypeRequest.java
```diff
@@ -161,4 +160,21 @@ protected SnomedReferenceSetMembers createEmptyResult(int limit) {
...
+			.filter(cs -> path.equals(cs.getBranchPath()) || path.startsWith(cs.getBranchPath() + Branch.SEPARATOR))
+			.collect(Collectors.toMap(CodeSystem::getId, Function.identity()));
+		
+		final Optional<CodeSystem> currentCodeSystem = snomedCodeSystems.values()
+			.stream()
+			// Longest matching working branch path (or prefix) wins
+			.max(Comparator.comparing(cs -> cs.getBranchPath().length()));
+
+		return currentCodeSystem.get().getSettings();
+	}
```
apeteri: `context.service(Resource.class)` should already have the code system IIRC.

apeteri (changes requested):

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/config/SnomedCoreConfiguration.java
apeteri: (This does not require fixing, just a note)

cmark: I agree with what Andras said above, I think minimally we should add a warning if any of these properties are still being specified in the config. Also, deprecate the class and not delete it.

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/request/SnomedMrcmTypeRequest.java
cmark: Yes, it is available, a lot of information already. Fetching extra data is always questionable and should always question yourself whether the one you are about to add is needed or not. Please refactor.

cmark (changes requested):

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/config/SnomedCoreConfiguration.java
AAAlinaaa: This is what the ticket said:
image
To me this suggested remove, not deprecate on 10.x, although admittedly didn't yet deprecate these on 9.x either.

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/request/SnomedMrcmTypeRequest.java
AAAlinaaa: @apeteri Ah, great, I was looking at how we solved this issue in other requests and found the above solution. Glad to hear there is a simpler one.

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/config/SnomedCoreConfiguration.java
cmark: Yes, but this ticket got created before the release of 10.0.0 and we did not update it to reflect the current state.

cmark: Talked about this with Alina and we think this has a very minimal impact so we opt for completely removing the config node in this case and not adding any warning or backward compatibility feature to the system to handle this change.

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/request/SnomedMrcmTypeRequest.java
```diff
@@ -91,7 +90,7 @@ public SnomedReferenceSetMembers doExecute(BranchContext context) {
 		
 		final String eclConstraint;
 		
-		Map<String,Object> settings = getResourceSettings(context);
+		Map<String,Object> settings = context.service(Resource.class).getSettings();
 		String defaultObjectAttributeExpresion = "<" + Concepts.CONCEPT_MODEL_OBJECT_ATTRIBUTE;
 		String defaultDataAttributeExpresion = "<" + Concepts.CONCEPT_MODEL_DATA_ATTRIBUTE;
```
cmark: These are constants somewhere, aren't they? If yes, please consolidate them to use the constants.

cmark (changes requested):

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/request/SnomedMrcmTypeRequest.java
AAAlinaaa: Not anymore, since the Mrcm config was removed, what would be a good place for these?

cmark (dismissed):
🥓

snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/request/SnomedMrcmTypeRequest.java
cmark: I did not mention the MRCM config in my comment above, I only said that these constants are declared somewhere, aren't they? It looks like they were not, but you added them so all good.

AAAlinaaa: I mentioned the config because they were defined there until now https://github.com/b2ihealthcare/snow-owl/blob/10.x/snomed/com.b2international.snowowl.snomed.datastore/src/com/b2international/snowowl/snomed/datastore/config/SnomedMrcmConfig.java#L25

apeteri (approved):
🍏

cmark (approved):
🥓

PR #1498, opened 2026-04-22
Bouncycastle, Log4j and Jetty security fixes

cmark:
This PR resolve the following security vulnerabilities:
- CVE-2026-3505
- CVE-2026-2332


releng/target-platform/target-platform.target
```diff
@@ -596,6 +609,37 @@
 				<version>8.5.16</version>
 				<type>jar</type>
 			</dependency>
+			<!-- Objectweb ASM required for Xtext and Elasticsearch (when upgrading make sure the version is supported by Xtext or it does not get duplicated from that repository) -->
```
apeteri, line 612: Can we apply a filter in the root `pom.xml` for this purpose?

apeteri (approved):
👍

releng/target-platform/target-platform.target
cmark: Yeah, but that's only for the build and we need to ensure that we only see a single jar in the tp, otherwise we need to dance with the version selection in the launch configurations. So I would not duplicate stuff there unless we really need to.

apeteri: The checkbox forest in the TP editor's "Content" tab might be taken into account in newer releases of Tycho, in which case it would be applicable both to IDE workspaces and builds. Unfortunately it generates a large list with the included bundles, not a small exclusion list 😞

cmark: Hmm, we can check that later, it would be nice to control this. Also, for some reason dependabot does not pick up security issues from the target file but in theory it should as they implemented the feature already last year.

AAAlinaaa (approved):

PR #1502, opened 2026-04-28
Map native Snow Owl resource IDs to FHIR-compatible identifiers

apeteri:
## Constraints on allowed values
(note that `\` only appears to escape the dash in regexp character class expressions)

- **Branch names**: allowed characters are `[a-zA-Z0-9._~\-]` with 100 characters limit
- **Native resource IDs**: allowed characters are `[A-Za-z0-9\-_]` with no length limit
  - in practice rules on branch names also apply
- Special resource IDs for **versioned resources**: contain single `/` character separating resource and version identifiers
  - first part should adhere to native resource ID rules
  - second part should adhere to branch name rules
  - maximum combined length is 100 + 1 + 100 characters
- **FHIR identifiers**: allowed characters are `[A-Za-z0-9\-.]` with 64 characters limit

## Proposed mapping rules
Evaluated in order in a left-to-right pass (no stacking of multiple rules:

1. Each occurrence of `"/"` is replaced with `"--"`
2. Each occurrence of `"--"` is replaced with `".h"`
3. Each occurrence of `"_"` is replaced with `".u"`

If the resulting ID would be longer than 64 characters, return the first 46 characters of the result, then `".c"`, then 16 hex digits of a [SipHash-2-4](https://en.wikipedia.org/wiki/SipHash) tag of the original input to ensure uniqueness while retaining a recoverable prefix of the original ID. 

If none of the rules match, the input is returned unmodified.

Jira: 
- https://snowowl.atlassian.net/browse/SO-6576

### DIFF
--- a/fhir/com.b2international.snowowl.fhir.core/src/com/b2international/snowowl/fhir/core/request/FhirResourceSearchRequest.java
+++ b/fhir/com.b2international.snowowl.fhir.core/src/com/b2international/snowowl/fhir/core/request/FhirResourceSearchRequest.java
@@ -113,12 +113,51 @@ public abstract class FhirResourceSearchRequest<T extends MetadataResource> exte
 	 */
 	protected abstract String getResourceType();
 
-	private final void addFhirIdFilter(final ExpressionBuilder query) {
-		// Both documents have an "id" and a "url" field; we will use the expression from ResourceDocument purely by choice
-		addIdFilter(query, ids -> Expressions.bool()
-			.should(ResourceDocument.Expressions.ids(ids))
-			.should(ResourceDocument.Expressions.urls(ids))
-			.build());
+	private final void addFhirIdFilter(final RepositoryContext context, final ExpressionBuilder query) {
+		addIdFilter(query, ids -> {
+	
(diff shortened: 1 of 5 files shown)


fhir/com.b2international.snowowl.fhir.core/src/com/b2international/snowowl/fhir/core/request/FhirResourceSearchRequest.java
```diff
@@ -113,12 +113,51 @@ protected final Bundle createEmptyResult(final int limit) {
 	 */
 	protected abstract String getResourceType();
 
-	private final void addFhirIdFilter(final ExpressionBuilder query) {
-		// Both documents have an "id" and a "url" field; we will use the expression from ResourceDocument purely by choice
-		addIdFilter(query, ids -> Expressions.bool()
-			.should(ResourceDocument.Expressions.ids(ids))
-			.should(ResourceDocument.Expressions.urls(ids))
```
apeteri, line 116: ConceptMap `$translate` operations (and potentially other callers) rely on the flexibility of the ID filter, so this clause might need to be preserved.

PR #1509, opened 2026-05-10
Release scanned classpath entries after boot phase

cmark:
This PR aims to reduce the base minimum memory requirement of the server by releasing the ClassPathScanner object used to scan the classpath for plugins and their configuration classes. After the boot phase and once everything is set up, only a handful of services used it during runtime, but that can be simply changed to cache the necessary classes and instantiate manually rather than relying on the cached scan results.


PekaryGergelyR (commented):
LGTM

cmark:
@AAAlinaaa @adamfilep  don't forget to review this as well please. Also, @PekaryGergelyR  please add a review to make it green as well. Thanks for the LGTM comment.

PekaryGergelyR (approved):

AAAlinaaa (approved):

PR #1510, opened 2026-05-13
SO-6591: Change X-Author header to optional in FHIR Delete requests

PekaryGergelyR:
https://snowowl.atlassian.net/browse/SO-6591


PekaryGergelyR:
@cmark , @apeteri 

I have only created tests for the `CodeSystem` endpoint.

- The `ConceptMap` tests are currently commented
- The `ValueSet` tests has this comment `Note: we do not have write access to VS resources and so result sets will always be empty in tests` I am not sure what could I do here.

cmark:
> @cmark , @apeteri
> 
> I have only created tests for the `CodeSystem` endpoint.
> 
> * The `ConceptMap` tests are currently commented
> * The `ValueSet` tests has this comment `Note: we do not have write access to VS resources and so result sets will always be empty in tests` I am not sure what could I do here.

It is okay to have two tests (and for specifying and one for omitting the header) for the CodeSystem endpoint only in this repo.

cmark (approved):
🥓

AAAlinaaa (approved):

apeteri (approved):
LGTM!

PR #1511, opened 2026-05-19
Minor performance optimizations

cmark:
-  The main performance gain is in the 69834fc011eb315fe6b3fb9bc85815b17ebf1cd3 commit, which avoids fetching the resource metadata MAIN branch twice in a row when trying to determine at what time the system should query content (aka extract a point in time for a versioned query). When there is no explicit timestamp or version in the URI we do not need to run the code therefore no need to initialize an index reader with a branch state.
- Other than this, the remaining commits focus on reducing dependencies and/or generating unnecessary objects (like ResourceURI parsing for example).


core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/ResourceURI.java
```diff
@@ -62,6 +62,9 @@ public final class ResourceURI implements Serializable, Comparable<ResourceURI>
 	 */
 	public static final String TILDE = "~";
 	
+	private static final String HEAD_PATH = "/HEAD";
+	private static final String HEAD_TILDE_PATH = "~HEAD";
```
apeteri: Is this part running frequently enough that we can be confident in the regexp parser not treating `/` and `~` as special characters? In Java they might not have any meaning.

apeteri (approved):
LGTM once the build agrees.

core/com.b2international.snowowl.core/src/com/b2international/snowowl/core/ResourceURI.java
cmark: The main problem is in the boundary when we move from the FHIR layer to the internal Java API layer. We have the URI already in parsed form, but now we convert it to the String version form to submit it to the internal API and then another parsing happens. This is basically involves running a Pattern matcher with the built-in parser and then ensuring we don't have the HEAD part in it. There are two additional improvement that I'd like to add:
- one to allow transferring ResourceURI instances to the Request API and only require over the wire comms to use the String form
- the other is to have an internal package-private constructor on the ResourceURI to simply create the value object rather than parsing it (probably moving the parsing logic to a factory method and in the ctor we would just set the values we already know)

Unfortunately something is off with the SNOMED tests, association targets get duplicated for some reason. :/

cmark: Fyi, running just one user load test shows that Pattern instances can be reduced by 20-25% by moving to this code, which reduces the generated objects from around 1Gb to ~750-800Mb. Running this for 10 or 50 VUs will have a significant impact on GC activity.

apeteri: I'm not contesting the change, just wondering whether it caught literal forward slash and tilde characters in the first place and not treated these symbols as some sort of regexp syntax (e.g. an unclosed delimiter for a regular expression).

cmark: Well, not sure, I only checked the part that run the regexp, which cause the Pattern class to be reported in JFR over 1Gb in the measured scenarios. I can double-check that the logic works properly with some extra unit tests.

cmark: Okay, there was one issue in the URI parsing class related to tilde handling, which changed the implementation a bit, please take a look again.

apeteri (approved):
🍏

PekaryGergelyR (approved):
LGTM

PekaryGergelyR (approved):

PR #1516, opened 2026-05-26
Incorrect revision merge logic (7.x)

cmark:
Backport of #1515

### DIFF
--- a/commons/com.b2international.index/src/com/b2international/index/revision/StagingArea.java
+++ b/commons/com.b2international.index/src/com/b2international/index/revision/StagingArea.java
@@ -964,14 +964,17 @@ public final class StagingArea {
 		
 		applyPropertyUpdates(toRef, propertyUpdatesToApply);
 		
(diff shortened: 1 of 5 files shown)


apeteri (approved):

commons/com.b2international.index/src/com/b2international/index/revision/StagingArea.java
apeteri, line 1197: `toOnlyRef` no longer places problem objects into `oldRevisionsById` but they will be still present in `updatedRevisionsById` (e.g. if `fromRef` is the HEAD of a code system working branch and `toOnlyRef` now only covers revisions that were created on a task branch and that task branch only). Doesn't that mean that we will still hit the `stageNew` branch in such cases?

cmark: Maybe we need to change this a bit different then, let us talk through over a call.

AAAlinaaa (approved):

apeteri (approved):

PR #1551, opened 2026-06-04
SO-6602: CodeSystem lookup null property bug

adamfilep:
JIRA: https://snowowl.atlassian.net/browse/SO-6602

Test: https://github.com/b2ihealthcare/snow-owl-ext/pull/3055

### DIFF
--- a/fhir/com.b2international.snowowl.fhir.core/src/com/b2international/snowowl/fhir/core/request/codesystem/FhirCodeSystemLookupRequest.java
+++ b/fhir/com.b2international.snowowl.fhir.core/src/com/b2international/snowowl/fhir/core/request/codesystem/FhirCodeSystemLookupRequest.java
@@ -141,27 +141,21 @@ final class FhirCodeSystemLookupRequest extends FhirCodeSystemOperationRequest<C
 				? Collections.emptyList()
 				: codeSystem.getProperty();
 		
-		// Official properties are always supported
-		final Set<String> lookupProperties = CodeSystemLookupParameters.OFFICIAL_R5_PROPERTY_VALUES;
+		// First, check if any of the properties are lookup request properties
+		final Set<String> nonLookupProperties = Sets.difference(requestedProperties, CodeSystemLookupParameters.OFFICIAL_R5_PROPERTY_VALUES);
 		
-		// Collect property URIs and codes from the code system
+		// Second, check if the remaining unsupported properties are supported by the CodeSystem either via full URL
 		final Set<String> supportedPropertyUris = codeSystemProperties.stream()
 				.map(CodeSystem.PropertyComponent::getUri)
 				.filter(Objects::nonNull)
 				.collect(Collectors.toSet());
+		final Set<String> unsupportedByUri = Sets.difference(nonLookupProperties, supportedPropertyUris);
 		
+		// or via their code only
 		final Set<String> supportedPropertyCodes = codeSystemProperties.stream()
 				.map(CodeSystem.PropertyComponent::getCode)
 				.filter(Objects::nonNull)
 				.collect(Collectors.toSet());
-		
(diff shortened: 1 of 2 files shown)


fhir/com.b2international.snowowl.fhir.core/src/com/b2international/snowowl/fhir/core/request/codesystem/FhirCodeSystemLookupRequest.java
```diff
@@ -140,7 +141,11 @@ private void validateRequestedProperties(CodeSystem codeSystem) {
 		// second check if the remaining unsupported properties supported by the CodeSystem either via full URL
 		final Set<String> supportedProperties = codeSystem.getProperty() == null 
 				? Collections.emptySet() 
-				: codeSystem.getProperty().stream().map(CodeSystem.PropertyComponent::getUri).collect(Collectors.toSet());
+				: codeSystem.getProperty().stream()
+					.map(CodeSystem.PropertyComponent::getUri)
+					// A CodeSystem.PropertyComponent can exist without a URI, in that case getUri() returns null
```
cmark: This will filter out a property if the URI is not set correctly. Please find the problem where the PropertyComponents are built.

cmark: And also, don't forget to create a test case.

cmark (changes requested):

fhir/com.b2international.snowowl.fhir.core/src/com/b2international/snowowl/fhir/core/request/codesystem/FhirCodeSystemLookupRequest.java
adamfilep: As I saw, PropertyComponents were built well, the null was caused by the fact that FHIR properties( system, version, display, name, designation) do not have a URL assigned to them. After our discussion, these properties will be returned in the response as is (not "**http://hl7.org/fhir/concept-properties/version**", just "**version**")

I put a test assertion into the ext project previously but forgot to link it here.

adamfilep: Added an additional test assertion into this repo too

apeteri (approved):

fhir/com.b2international.snowowl.fhir.core/src/com/b2international/snowowl/fhir/core/request/codesystem/FhirCodeSystemLookupRequest.java
```diff
@@ -128,34 +130,51 @@ protected Set<String> configureFieldsToLoad() {
...
-		
-		if (!unsupportedCodes.isEmpty()) {
-			if (unsupportedCodes.size() == 1) {
-				throw new BadRequestException(String.format("Unrecognized property %s. Supported properties are: %s.", unsupportedCodes, supportedProperties), "LookupRequest.property");
+		final List<CodeSystem.PropertyComponent> codeSystemProperties = codeSystem.getProperty() == null
+				? Collections.emptyList()
+				: codeSystem.getProperty();
+		
+		// Official properties are always supported
+		final Set<String> lookupProperties = CodeSystemLookupParameters.OFFICIAL_R5_PROPERTY_VALUES;
```
cmark: Variable naming is not the best here, previously this was not there, it was only used in the diffing, please restore that. Also, move statements closer to each other to ease understanding of the logic.

cmark (changes requested):

cmark (approved):
🥓

PR #1552, opened 2026-06-05
SO-6601: FHIR Resource PUT/POST improvements

PekaryGergelyR:
https://snowowl.atlassian.net/browse/SO-6601

### DIFF
--- a/fhir/com.b2international.snowowl.fhir.core/src/com/b2international/snowowl/fhir/core/request/conceptmap/FhirWriteSupport.java
+++ b/fhir/com.b2international.snowowl.fhir.core/src/com/b2international/snowowl/fhir/core/request/conceptmap/FhirWriteSupport.java
@@ -16,88 +16,54 @@
 package com.b2international.snowowl.fhir.core.request.conceptmap;
 
 import java.time.LocalDate;
+import java.time.ZoneOffset;
 import java.util.Optional;
+
+import org.hl7.fhir.r5.model.Meta;
+import org.hl7.fhir.r5.model.MetadataResource;
 
 import com.b2international.snowowl.core.ResourceURI;
 import com.b2international.snowowl.core.ServiceProvider;
-import com.b2international.snowowl.core.date.EffectiveTimes;
 import com.b2international.snowowl.core.request.ResourceRequests;
 import com.b2international.snowowl.core.version.Version;
-import com.b2international.snowowl.fhir.core.exceptions.BadRequestException;
 
 /**
  * @since 8.7.1
  */
 public interface FhirWriteSupport {
-
-	/**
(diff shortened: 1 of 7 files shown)


fhir/com.b2international.snowowl.fhir.core/src/com/b2international/snowowl/fhir/core/request/conceptmap/FhirWriteSupport.java
```diff
@@ -16,88 +16,54 @@
...
-		VERSION_EXISTS,
-
-		/**
-		 * Newer version present in the system, write operation should raise an error
-		 */
-		NEWER_VERSION_EXISTS
+	default LocalDate getEffectiveDate(MetadataResource resource, LocalDate defaultEffectiveDate) {
+		return Optional.ofNullable(resource.getDate())
+				.or(() -> Optional.ofNullable(resource.getMeta()).map(Meta::getLastUpdated))
+				// Time zone information is no longer available as we mapped it to a java.util.Date instance; treat value as UTC
```
apeteri, line 37: I checked in the meantime - parsing logic in FHIR Core handles datetime values without a time-zone offset or Z suffix in the input as if they were expressed in the time zone returned by `TimeZone.getDefault()` (lines 633 and 678-679 are relevant):

https://github.com/hapifhir/org.hl7.fhir.core/blob/92330771525202811b071b3158773482e1cc1210/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/model/BaseDateTimeType.java#L630-L737

...I'm not sure what this means in practice though 😄 My initial assumption was that it will be parsed as a UTC value and this revelation makes the conversion step unnecessary. ...possibly?

apeteri (approved):

fhir/com.b2international.snowowl.fhir.core/src/com/b2international/snowowl/fhir/core/request/conceptmap/FhirWriteSupport.java
apeteri: Oh, reading it a second time makes me think this is OK as is. We get a `long` timestamp value wrapped in the `java.util.Date` instance which can be expressed/rendered in any timezone. We choose UTC for storage. 👍

PekaryGergelyR: Well I had to add the jvm arg to force the UTC, @cmark said for now lets go with that. Though I am not sure when will come back to haunt us.

adamfilep (approved):

